PR #3, opened 2015-08-06
test a pull request.

tsteven4:
/etc/os-release is becoming somewhat of a standard.


tsteven4:
rejected as this was just a test.

PR #17, opened 2015-12-19
Have the gpx reader forward namespace declarations for http://www.w3.…

tsteven4:
…org/2001/XMLSchema-instance like all other namespaces.

This avoids the problem of the gpx writer potentially using the prefix when it has not been declared.


robertlipe:
Thank you. I defer to your wisdom on this as I've never really been proud
of our namespace handling.

PR #20, opened 2016-01-03
Add support for Globalsat gh625XT

zingo:


zingo:
Missing test, don't know how to do one for serial target. Maybe a "-bin" version can be created, should I look into this?

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+ * ./gpsbabel -i globalsat -f /dev/ttyUSB0 -o gpx,garminextensions -F xxx.gpx
+ * ./gpsbabel -i globalsat -f /dev/ttyUSB0 -o gtrnctr,course=0,sport=Running -F xxx.fit
+ *
+ * ./gpsbabel-debug -i globalsat -f /dev/ttyUSB0 -o gpx,garminextensions -F xxx.gpx
+ */
+
+
+#include <ctype.h>
+#include <limits.h>
+#include <endian.h>
```
robertlipe: <endian.h> isn't a standard header.  We have be_read_[float,double] and friends like be_write16() in defs.h

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
+/*
+    Globalsat Sport serial protocol.
+
+    Copyright (C) 2012, Zingo Andersen
```
robertlipe: And me, please, for legal reasons.  It's easier to defend in a copyright tussle if I'm an owner.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+
+    Dual Licese Public Domain and GPLv2 or later
```
robertlipe: Strongly prefer not diluting our license.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program; if not, write to the Free Software
+    Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111 USA
+
+ */
+
+/*
+ * This is the bridge between the GPSBable and globalsport device
```
robertlipe: "GPSBabel". Also please include references to protocol definition, if available, or mention that this was reverse engineered.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+ * gh625XT
+ *
+ *
+ * ./gpsbabel -i globalsat -f /dev/ttyUSB0 -o gpx,garminextensions -F xxx.gpx
+ * ./gpsbabel -i globalsat -f /dev/ttyUSB0 -o gtrnctr,course=0,sport=Running -F xxx.fit
+ *
+ * ./gpsbabel-debug -i globalsat -f /dev/ttyUSB0 -o gpx,garminextensions -F xxx.gpx
```
robertlipe: Since globalsat is a company/brand, I wonder if we should be more specific.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+	HeaderTypeTrackPoints = 0x55,
+
+	ResponseInsuficientMemory = 0x95,
+	ResponseResendTrackSection = 0x92,
+	ResponseSendTrackFinish = 0x9A
+} globalsat_commands_e;
+
+
+//Pack structures since they should conform to serial data spec
+#pragma pack(1)
```
robertlipe: We've avoided #pragma pack so far, manually packing structures when we have to just because this is such a messy area.  In fact, we usually avoid structures in protocol definitions, but I could be convinced that's a battle that's not worth fighting.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+	uint16_t PwrCadence;
+	uint16_t Power;
+} gh_recpoint;
+
+//#pragma options align=reset
+
+static void
+serial_init (const char *fname)
+{
+	if (serial_handle = gbser_init (fname), NULL == serial_handle) {
```
robertlipe: I'll fix it with astyle, but we generally write init() instead of init ().  Yes, I know that whitespace code convention arguments are a hassle.   I'll fix before submit if necessary.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+	DeviceCommand = serial_recv_byte ();
+	if (global_opts.debug_level > 1) {
+		printf ("DeviceCommand: 0x%02x ", DeviceCommand);
+	}
+	len_h = serial_recv_byte ();
+	calc_crc ^= len_h;
+	len_l = serial_recv_byte ();
+	calc_crc ^= len_l;
+
+	lenght = (len_h << 8) + len_l;
```
robertlipe: I make that typo a lot, too. :-)

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+	}
+	//crc error
+	free (payload);
+	error_out:
+	*out_lenght = 0;
+	return NULL;
+}
+
+/*
+ * Send on byte package
```
robertlipe: "one" ?

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+{
+	uint8_t payload[1];
+	payload[0] = command;
+	globalsat_write_package (payload, 1);
+}
+
+static void
+globalsat_probe_device ()
+{
+	//2 1 0 191 190
```
robertlipe: Explain.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+
+	uint8_t *in_payload;
+	int len;
+	uint8_t DeviceCommand;
+	in_payload = globalsat_read_package (&len, &DeviceCommand);
+	if ((len > 0) && (in_payload != NULL)) {
+		if (global_opts.debug_level > 1) {
+			printf ("Got package!!!\n");
+		}
+		//TODO figure out what device it is
```
robertlipe: Isn't that TODOne above or are there cases beyond the 561?

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+	}
+	track_read ();
+}
+
+static void
+track_read(void)
+{
+	route_head *trk = NULL;
+	Waypoint *wpt = NULL;
+    gpsbabel::DateTime gpsDateTime;
```
robertlipe: Nit: For readability, you can move the decls to their first use which also cleans up the default initializer case.

Some day we can use C++11 which allows 'auto'...

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+	uint8_t *payload;
+	int length;
+	uint8_t DeviceCommand;
+	payload = globalsat_read_package (&length, &DeviceCommand);
+	if ((length > 0) && (payload != NULL)) {
+		if (global_opts.debug_level > 1) {
+			printf ("Got package!!! headers\n");
+		}
+		int number_headers;
+		number_headers = length / sizeof (gh_trainheader);	//29=packed sizeof(gh_trainheader)
```
robertlipe: int number_headers = length ...

If "29" is the correct number and the code goes up in flames if it's not 29, please add a is_fatal() or a static assertion.  (It's OK to do this at startup instead of in every loop iteration.)

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+		}
+
+		int i;
+		for (i = 0; i < number_headers; i++) {
+			int pos = i * sizeof (gh_trainheader);
+			gh_trainheader *header = (gh_trainheader *) & (payload[pos]);
+			if (global_opts.debug_level > 1) {
+				printf ("%02d-%02d-%02d ", header->dateStart.Year,header->dateStart.Month, header->dateStart.Day);
+				printf ("%02d:%02d:%02d ", header->timeStart.Hour,header->timeStart.Minute, header->timeStart.Second);
+				printf ("Total(points:%6d time:%6ds dist:%9dm) LapCnts:%5d ",be32toh (header->TotalPoint),be32toh (header->TotalTime) / 10,	be32toh (header->TotalDistance), be16toh (header->LapCnts));
```
robertlipe: be32toh isn't standard. See above.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+				printf ("Index/StartPt:%d ", be32toh (header->gh_ptrec.Index));
+				printf ("LapIndex/EndPt:%d ", be32toh (header->gh_laprec.LapIndex));
+				printf ("DataType:0x%x\n", header->DataType);
+			}
+		}
+
+		for (i = 0; i < number_headers; i++) {
+			int pos = i * sizeof (gh_trainheader);
+			gh_trainheader *header = (gh_trainheader *) & (payload[pos]);
```
robertlipe: Will we get into trouble type-punning here if the gh_trainheader and payload have different alignment requirements?

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+			free (track_payload);
+			track_payload = NULL;
+
+
+			// Get laps
+			while (total_laps_left > 0) {
+				globalsat_send_simple (CommandGetNextTrackSection);
+				track_payload = globalsat_read_package (&track_length, &trackDeviceCommand);
+				if ((track_length == 0) || (track_payload == NULL)) {
+					//TODO error
```
robertlipe: For these "todo error", especially in the read cases, it's OK to just call is_fatal() - while the comm protocol may be in a whacked out state if you retry after a failure, at least we don't keep marching on for cases that won't work.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+					printf ("Total(points:%6d time:%6ds dist:%9dm) LapCnts:%5d ", be32toh (header->TotalPoint),be32toh (header->TotalTime) / 10, be32toh (header->TotalDistance), be16toh (header->LapCnts));
+					printf ("Index/StartPt:%d ", be32toh (header->gh_ptrec.Index));
+					printf ("LapIndex/EndPt:%d ", be32toh (header->gh_laprec.LapIndex));
+					printf ("DataType:0x%x\n", header->DataType);
+				}
+				/*
+				 * GPS year: 2000+; struct tm year: 1900+
+				 * GPS month: 1-12, struct tm month: 0-11
+				 */
+/*				gpstime.tm_year  = header->dateStart.Year+100;
```
robertlipe: Heh. Looks like you started this before we had the MUCH more sensible QDateTime family.

globalsat_sport.cc
```diff
@@ -0,0 +1,742 @@
...
+							printf (" Speed:%f HR:%d", (double) be32toh (point->Speed) / 100, point->HeartRate);
+							printf (" Time:%d Cadence:%d", be32toh (point->IntervalTime), be16toh (point->Cadence));
+							printf (" PwrCadense:%d Power:%d\n", be16toh (point->PwrCadence),be16toh (point->Power));
+						}
+
+
+						wpt = new Waypoint(); // waypt_new ();
+//						wpt->creation_time = mkgmtime(&gpstime);
+						wpt->SetCreationTime(gpsbabeltime);
+						wpt->longitude = (double) ((int32_t) be32toh (point->Longitude)) / 1000000.0;
```
robertlipe: double casting should never be necessary.

robertlipe:
Having a binary file read/write for a serial protocol is a bit awkward, but it really pays off in support and when those of us not having the hardware need to make changes, but it's admittedly awkward.  Having someone send you the bytes that would be pulled out of their receiver is quite helpful.  We have a few examples in the code; some are done better than others.  Providing different vecs entries and abstracting out the 'get bytes' from file to gbser\* seems to work well.

We had one format that did this by reading everything it could from the unit (it didn't need ack/retries...) and making a file in the current directory and then doing everything through the file ops. This was bad because you probably didn't have write permissions on the installed directory.

zingo:
Thanks for your time and feedback, Ill look into and fix your comments and re-pullrequest. 
You are correct about the timespan this is an old piece of code that has been on my bad-carma-submit list for a long time.

robertlipe:
Thanx!

I also meant to ask for a few paragraphs for xmldoc/formats/ and
xmldoc/formats/options that explain what this format it, how it can be used
(cables, device settings, supported devices, etc.) and what the options
actually do.

docbook is a bit funky, but if you can provide the words, I'll help with
the markup.

YIL that github stupidly sends an email for every comment in a file instead
of waiting for you to submit them like the code review tools I've used
before. Sorry about that. I'll look for a way to make that less annoying.

globalsat_sport.cc
zingo: Some leftover dev comments (it is the bytes set over the USBserial line to the device) Ill just remove them.

globalsat_sport.cc
zingo: Not sure what you mean here.

globalsat_sport.cc
robertlipe: Did you mean a "one byte package" ?

zingo: Yes, but it was 

---
- len_h = serial_recv_byte ();
- calc_crc ^= len_h;
- len_l = serial_recv_byte ();
- calc_crc ^= len_l;
  +
- lenght = (len_h << 8) + len_l;

"I make that typo a lot, too. :-)" 

---

I didn't understand :)

GPSBabelDeveloper: lengTH, not lengHT.

zingo:
I think I have fixed everything now :)

globalsat_sport.cc
zingo: Thanks, I was looking for a logical error :)

globalsat_sport.cc
```diff
@@ -0,0 +1,901 @@
...
+	} else {
+		int bytes;
+		bytes=gbfread(&result, 1, 1, in_file);
+		is_fatal((bytes != 1), MYNAME ": read error");
+	}
+	// Check if byte sould be dumped also into a file
+	if (dumpfile) {
+      gbfwrite(&result, 1, 1, dumpfile);
+    }
+return result;
```
robertlipe: I'll fix indention later.

globalsat_sport.cc
```diff
@@ -0,0 +1,901 @@
...
+	if (global_opts.debug_level > 1) {
+		printf(MYNAME " wr_init()\n");
+	}
+	serial_init(fname);
+}
+
+
+static void
+rd_deinit(void)
+{
```
robertlipe, line 447: I'll fix indention later.

GPSBabelDeveloper, line 447: There are some style nits, but overall, this is progressing nicely. Thank
you.

If you can find and address the leaks found by
https://travis-ci.org/gpsbabel/gpsbabel/jobs/102729158 to get a clean
presubmit, I'd welcome this into the next release. (Which is likely to be
this weekend.)

I've not checked, but I THINK it's the
  static char buf[80]
  sprintf(junk into buf)
  wpt->something = xstrdup(buf);

Either buf should be correctly sized by xasprintf and then freed or we
should just let QString do char\* copy/duplication from the char\* into the
QString and not introduce an additional malloc/copy via xstrdup.

I could be wrong, but that caught my eye during review, so it seems like a
bit of a smoking gun.

Thanx!
RJL

robertlipe, line 447: Parting shot before I call it a day...

Since you seem to be a Linux dev, install valgrind then run ./vtesto
globalsat_sport  that runs 'testo', but under valgrind with leak detection
turned on. It should take much less time than running the full test and it
should let you zero in pretty quickly.  ISTR there are diagnostics showing
where the leaked blocks were allocated, but they don't seem to show up in
our travis log.

Thanx for looking into it.
RJL

zingo:
You where completely right, it was the xstrdup() switch to QString as you suggested, I think it improves the code as the destination struct-field is a QString also. Thanks for helping out, I did a rebase also since that was needed.

Issue #1, opened 2015-07-31
testing review system


GoogleCodeExporter:
```
Purpose of code changes on this branch:


When reviewing my code changes, please focus on:


After the review, I'll merge this branch into:
/trunk

```

Original issue reported on code.google.com by `robertl...@gmail.com` on 24 Sep 2013 at 3:00

Attachments:
- [string_scribble.patch](https://storage.googleapis.com/google-code-attachments/gpsbabel/issue-1/comment-0/string_scribble.patch)

PR #27, opened 2016-01-14
mtk_logger mtk_parse_info: Fix track start for Holux devices

stefan-a-bauer:
Previously, a new track was created when the event "GPS Logger turned on"
was read. This is correct for the MTK_LOGGER only, but not for the
Holux devices M241 & GR245.

Case #1: Firmware versions where the event is never logged. Effect: One large
track is created.
Applies to GR245 v1.x

Case #2: Firmware versions where the event is logged, but not
at the beginning of a track, but _after_ the first track point of a track
(at least most of the times).
Applies to GR245 v2.00, M241 v1.12, v1.13

For all of the currently known Holux firmware versions, the correct
indication of a track start is the "Log period change" /
"Log distance change" event.

This commit adds some more testcases (241 case provided by Erik Krause)
that were suffering from Case #2 before.

The existing 245 test is an example for Case #1 and has been fixed.


robertlipe:
Integrated. Thank you.

PR #28, opened 2016-01-16
Add Geogrid-Viewer binary file reader

ra1fh:
Supports file format version 2.0, 3.0 and 4.0

This is an improved version of the patch I sent to the gpsbabel mailing list about a year ago. It reads all of the ~150 random ovl samples I found on the internet.

I have tested on a number of platforms (Debian 8 i386, OpenBSD-current amd64, OpenBSD 5.8 sparc64, Fedora 23 amd64) to make sure the binary parsing works ok and endianess conversion is correct.


robertlipe:
Thanx, Ralf.   That seems worth taking in the version I hope to cut this
weekend.

Interesting approach. Did you consider using
http://doc.qt.io/qt-4.8/qdatastream.html ? I've long wanted to try it with
GPSBabel, but it's not clear how well it plays with externally defined
formats which are, of course, our whole world.  If you could readRawData
and then just stream to and from with qints and qdoubles, that could be
pretty convenient and less error prone than our historical approach.

Stylistically, you can use the C99/C++ approach of moving your declarations
to thei point of first use instead of at the top of the function, but that
doesn't bother me enough to change it. :-)

foo() {
  MyType bar;
  /// a billion lines of code.
  bar = GetThing();   // What, what type of Thing is Bar?  MyType bar =
GetThing() would be clearer.
}

Some day we'll be able to use C++ 11 and use 'auto bar = GetThing()'.

robertlipe:
Thank you. Merged.

ra1fh:
Thanks for merging.

If I remember correctly I did consider QDataStream when starting with the Geogrid reader, and then discarded it for some reason that I don't remember.

Anyway, for a quick test I just changed ggv_bin.c completely to QDataStream, works just fine. Even on sparc64. The 150 ovl samples still parse correctly on all of the above listed platforms. The stream operators are used for quint16, quint32 and double.

Should I submit a pull request?

The minimum required QT version for this would probably be 4.6, as the floating point precision was introduced in that version. My Fedora box uses qt-4.8, the others either qt-5.3 or qt-5.4. So those versions do work.

robertlipe:
Cool!  I really was just looking to see if there was a reason to avoid
QDataStream, but that's pretty interesting.  Is the code substantially more
maintainable?

We already have Qt 4.6 as a floor. Per
https://www.gpsbabel.org/htmldoc-development/Source.html "Qt versions 4.6
and newer are supported."   Even that is like 6-8 major revisions back, so
I don't see that as unreasonable... unless some version of Centos or
something ships with 4.6.0 and we trigger a bug that was fixed in 4.6.1 or
something.   Frankly, I'm more interested in having a good, readable
example of a binary file format that's endian independent and not a total
mess to read than in supporting every possible combination out there, so
I'd consider just jettisoning the format for Qt <= bug fix, but I hope
that's not even an issue.

So, yes, please. I've been pulled away tonight (saturday night) and am
unlikely to work on release things until Sunday night or Monday, a U.S.
holiday.

ra1fh:
Yes I think code using QDataStream will be more maintainable as it should automatically avoid any kind of word size, endianess and alignment problems (see the sparc bug report I submitted today). In general the code should be better readable as well. Not sure how much that applies to this particular reader, as it has lots of debug code that is useful for reverse engineering the file format.

I've uploaded the QDataStream change and made a pull request.

PR #30, opened 2016-01-17
Use QDataStream for Geogrid-Viewer binary files

ra1fh:
This eliminates the need to do the manual endianess handling
and converting binary data into a double. Works on SPARC64
as well (big endian).


robertlipe:
Thankyou. This does flow better on the writes, especially.

PR #31, opened 2016-01-17
Add GH625XT support

zingo:
Fixed the memory leak using QString, I think turened out a lot cleaner then before.


zingo:
Ignore this, did a rebase in the original pull request instead ( https://github.com/gpsbabel/gpsbabel/pull/20 )

PR #33, opened 2016-02-18
use local encoding for file name command line parameters (-f, -F).

tsteven4:


brauniger_iq.cc
```diff
@@ -56,13 +56,13 @@ inline state_t operator++(state_t& rs, int)
 
 static const int reqd_bytes[num_states] = { 6, 1, 2, 2, 25, 2, 2, 2, 2, 2, 2, 1, 2, 2, 2, 1 };
 
-static void rd_init(const char* fname)
+static void rd_init(const QString& fname)
 {
-  if (serial_handle = gbser_init(fname), NULL == serial_handle) {
-    fatal(MYNAME ": Can't open port '%s'\n", fname);
+  if (serial_handle = gbser_init(qPrintable(fname)), NULL == serial_handle) {
+    fatal(MYNAME ": Can't open port '%s'\n", qPrintable(fname));
```
robertlipe, line 62: As an aside,  I've been slowly moving these to Fatal() << "prose" << fname but it's just complicated enough that I've not automated it.
I should probably come up with a clever way to get the module name by default.

https://github.com/gpsbabel/gpsbabel/blob/master/src/core/logging.h

bushnell.cc
```diff
@@ -168,20 +169,18 @@ rd_deinit(void)
...
   // If user provided an extension in the pathname, whack it.
-  dot = strrchr(ofname, '.');
-  slash = strrchr(ofname, GB_PATHSEP);
-  if (dot > slash) {
-    *dot = 0;
+  ofname = fname;
+  int suffix_len = QFileInfo(fname).suffix().length();
+  if (suffix_len > 0) {
+    /* drop the suffix and the period */
+    ofname.chop(suffix_len + 1);
```
robertlipe, line 183: Is this ofname = QFileInfo(fname).basename() ? Maybe a little more if we support directory names here.
Given where Bushnell is in the GPS industry, I'm not sure I care much about doing this well.

bushnell_trl.cc
```diff
@@ -53,15 +53,17 @@ rd_deinit(void)
...
+  int i;
+  int l = fname.length();
   char obuf[20] = { 0 } ;
   char* p = obuf;
   file_out = gbfopen_le(fname, "w", MYNAME);
   trkpt_count = 0;
+  QString fnameu = fname.toUpper();
   for (i = 0; (i < l) && (i < 20); i++) {
-    char c = toupper(fname[i]);
+    char c = fnameu[i].toLatin1();
```
robertlipe, line 66: Should this be ... fnameu[i].toUpper().toLatin1() ?

kml.cc
```diff
@@ -548,12 +548,12 @@ kml_wr_deinit(void)
...
 
-  if (posnfilenametmp) {
+  if (!posnfilenametmp.isEmpty()) {
 #if __WIN32__
-    MoveFileExA(posnfilenametmp, posnfilename,
+    MoveFileExA(qPrintable(posnfilenametmp), qPrintable(posnfilename),
                 MOVEFILE_REPLACE_EXISTING);
 #endif
-    rename(posnfilenametmp, posnfilename);
+    QFile::rename(posnfilenametmp, posnfilename);
```
robertlipe, line 556: I'd bet we don't actually need this abstraction as QFile::rename() should work everywhere, but this is fine.

robertlipe:
An overdue cleanup. Thanx for taking on this bulky task.

bushnell_trl.cc
tsteven4, line 66: no.  toUpper taken care of at line 64.

PR #40, opened 2016-03-28
Support for iGo Primo POI (.upoi) format

achmafooma:
I have written support for the iGo Primo POI (.upoi) format. This is different from the iGo formats already available (though it has some similarities to iGo 2008 .upoi).

iGo Primo is navigation software for Windows Embedded that is used in a number of OEM and aftermarket automotive head units. I tested this on a Jensen VX7022 unit and it is working well.

Converting FROM .upoi is working reliably without issues in my testing.

Converting TO .upoi results in a UTF-8 file, but the device demands UTF-16 LE (UCS-2 LE BOM). The actual format of the file is correct, but I have to manually convert it (using a text editor) to UTF-16 LE encoding before the device will read it.

I put a note to that effect in the XML description for the format.


tsteven4:
You shouldb be able to get your format to write UTF-16 using 
QTextCodec.  A codec is setup based on your ff_vecs encode as 
global_opts.codec.  We need to be careful to use codecs that are 
available on all platforms, but

> - UTF-16
> - UTF-16BE
> - UTF-16LE
> 
> are all on the list of generally supported codecs.

The cet stuff is on its way out and it pretty much gutted.  But enough 
is left that it will try to set up a QTextCodec for you automatically.

tsteven4:
I had assumed you wrote a new format from scratch and based my comments 
on the codec setup based on that assumption.  Now I noticed you used a 
style file to do this conversion.   I don't believe the encoding support 
for formats based on style files is currently functional.

tsteven4:
You forgot to include your reference file:
XCSV: Cannot open file './reference/igoprimo_poi.upoi'!
this is why the checks failed.

achmafooma:
Thanks -- sorry about that. This should fix it.

tsteven4:
I am exploring repairing the support for encoding in style files.  Is it correct that the upoi files in the igo primo format are encoded with UTF16LE and have a BOM?  Is this always the case?  Can you supply a reference file in the state it exists on the device?

Would you know about the encoding in the igo 2008 format?  Our style file says CP-1252, but the reference file we have is UTF16LE.

tsteven4:
I have restored support for ENCODING in xscv style files.  You may want to change the ENCODING in your style file to UTF-16LE.

achmafooma:
Thanks! It does seem to always be the case that it's UTF-16LE with BOM, at least on my device (sample size of 1 ;-)).

I updated the style to UTF-16LE, and rebuilt with your change and mine. After doing that, I converted a GPX file (from my Garmin) to the iGo Primo POI format and it loaded on my device with no further modification. So I think we're in business.

I also created a new set of reference files, including a .upoi as it exists on the device.

As for iGo 2008, I don't know anything about it...except that it seems to have a lot of similarities with this newer iGo Primo format. My hunch is that it would share the same UTF-16LE encoding, but it is just a hunch.

tsteven4:
It would be nice to test the writer as well as the reader.  Perhaps you can add something like this to you test:
gpsbabel -i igoprimo_poi -f ${REFERENCE}/igoprimo_poi.upoi -o unicsv -F ${TMPDIR}/igoprimo_poi~upoi.ucsv 
gpsbabel -i unicsv -f ${TMPDIR}/igoprimo_poi~upoi.ucsv -o igoprimo_poi -F ${TMPDIR}/igoprimo_poi~upoi~ucsv.upoi
compare ${REFERENCE}/igoprimo_poi.upoi ${TMPDIR}/igoprimo_poi~upoi~ucsv.upoi

To get this to work I had to change the RECORD_DELIMITER in your style file to CRNEWLINE.  I also added a newline at the end of your style file.

achmafooma:
Added a test for the writer and changed to CRNEWLINE. The check is still running but it got past the igoprimo test so I'm pretty sure it'll pass.

I think I found a bug -- on Windows, with CRNEWLINE, the file comes out with mangled line endings (CRCRLF). I was able to fix it by modifying xcsv.cc:

This line:

`xcsv_file.file->open(QFile::WriteOnly | QFile::Text);`

Changed to:

`xcsv_file.file->open(QFile::WriteOnly);`

Looks like we can't open the file in text mode, or else on Windows it tries to do its own line endings and combines them with the one we have defined in the style.

I didn't push the fix since I don't know if it'll have unintended consequences. Let me know if you want me to go ahead and add it to this pull request (or if you want me to open a new one).

Thanks!

tsteven4:
Ugg.  That looks like a very old bug, with lots of potential side effects.  The CRNEWLINE will force unix to output windows line endings, but as you discovered windows doesn't like that.  We have a conflict between the explicit line endings in the style file, and the historical automatic conversion of CR or NEWLINE to CR NL on windows platforms.  We could try to plug the hole by not use text mode if the record delimiter is CRNEWLINE, but I want to solicit some more feedback before we go there.

Maybe you should just go back to a RECORD_DELIMITER of CR and leave the open as it is in master.  On windows platforms you will get your CR NL, on unix platforms NL.  I think our test suite will ignores these differences, so it should pass on either.  It is unclear to me if the consumer of these files requires one type of line ending or will work with either.

tsteven4:
Actually you probably want to use NEWLINE.  Either as another bug, or a feature, we treat CR as a newline anyway.  With the built in style files there are no other cases of CR being used as the record delimiter.

tsteven4:
Although our testsuite function compare ignores whitespace changes, which would normally cover the line ending differences, it doesn't work out in this case because of the encoding.  Something like this might work, depending on the availability of iconv on our test systems.  Note we don't really need to go through unicsv as I suggested before.
gpsbabel -i igoprimo_poi -f ${REFERENCE}/igoprimo_poi.upoi -o igoprimo_poi -F ${TMPDIR}/igoprimo_poi~upoi.upoi
iconv -f UTF-16LE -t UTF-8 ${REFERENCE}/igoprimo_poi.upoi >${TMPDIR}/igoprimo_poi.upoi.utf8
iconv -f UTF-16LE -t UTF-8 ${TMPDIR}/igoprimo_poi~upoi.upoi >${TMPDIR}/igoprimo_poi~upoi.upoi.utf8
compare ${TMPDIR}/igoprimo_poi~upoi.upoi.utf8 ${TMPDIR}/igoprimo_poi~upoi.upoi.utf

I would like to merge this PR, with your write test, with a record delimiter of NEWLINE, and without the open changes to xcsv.cc.  If the write test proves problematic I could make the test conditional on the availability of iconv.

achmafooma:
Okay -- switched it to NEWLINE and updated the test to use iconv. It seems to have worked; the builds are still running but they seem to have made it past the igoprimo_poi checks.

After the checks complete, hopefully you're good to merge the PR. Thanks!

robertlipe:
Thumbs up to both of you for your diligence. This is open source working as
intended. Thanx!

tsteven4:
committed.  thanks.

PR #41, opened 2016-03-31
Fix default values for targetlocation and configlog flags.

miz884:
The previous default value "" was considered as TRUE in the IF condition. The targetlocation and configlog flags took effect even if users didn't specify it. There were no workarounds. It prevented users to reach to the main function skytraq_read_tracks.

Those should be NULL by default so that the IF condition should be FALSE if a user doesn't specify it.


kanru:
Can this patch be merged? Anything else needed to fix issue #51 ?

robertlipe:
Thanx. There had been some waffling on this being the right fix, but I've just merged it.

PR #42, opened 2016-03-31
Use libudev to find serial devices dynamically on Linux

otrichet:


gui/app.pro
```diff
@@ -28,6 +28,13 @@ unix:OBJECTS_DIR = objects
 unix:RCC_DIR = objects
 
 mac:LIBS += -framework IOKit -framework CoreFoundation
+unix {
```
robertlipe, line 31: Let's put this under "linux" instead of "unix". I doubt the below will work on FreeBSD/OpenBSD and we do have a few users there.  Then put the body of your new serial_unix.cc into a new file, serial_linux.cc.

gui/serial_unix.cc
```diff
@@ -21,6 +21,68 @@
...
+
+        bool okMaj, okMin;
+        int major = QString(udev_device_get_property_value(dev, "MAJOR")).toInt(&okMaj);
+        int minor = QString(udev_device_get_property_value(dev, "MINOR")).toInt(&okMin);
+        if(!okMaj || !okMin) {
+             major = -1;
+             minor = -1;
+        }
+
+        // see Documentation/devices.txt in the linux tree
```
robertlipe, line 55: "Linux kernel" tree.

gui/serial_unix.cc
```diff
@@ -21,6 +21,68 @@
...
+    struct udev *udev = udev_new();
+    if(!udev) {
+        qDebug() << "Can't create udev";
+        return QStringList();
+    }
+
+    QSet<QString> devices;
+
+    struct udev_enumerate *enumerate = udev_enumerate_new(udev);
+    udev_enumerate_add_match_subsystem(enumerate, "tty");
```
robertlipe, line 39: Will this get the bluetooth and USB/serial devices? Seems like it should, but it's perhaps an edge case.

GPSBabelDeveloper, line 39: You know, if we could depend on Qt 5.1 this could be

auto blah = QSerialPortInfo::availablePorts();
for (port : blah) {
   box->addItem(port.portName);
}

Then all our OS-specific versions could be this.

http://doc.qt.io/qt-5/qserialport.html#details

Replacing my Windows/UNIX abstraction of *serial.cc with QSerialPort would
be a little more risky just because that code has been battle tested and
it's really a pain to test any reasonable matrix of OSes and devices.
http://doc.qt.io/qt-5/qserialportinfo.html  It's funny that I wanted a QT
serial system years ago and didn't even noticed that one sneaked in...

At https://www.gpsbabel.org/htmldoc-development/Source.html we set the
floor for Qt at 4.6, which is ridiculously old. Is it time we raise the
floor for Qt?
My read is that our travis instance runs Ubuntu Trusty (2014) and has Qt
5.2 https://lists.ubuntu.com/archives/ubuntu-devel/2014-February/038114.html

tsteven4, line 39: on travis we are on Ubuntu precise (12.04.5 LTS) running Qt version 
4.8.1.  I have most of the version information printed to the logs on 
travis every time you do a build.

robertlipe, line 39: Four years old. Ugh.

https://docs.travis-ci.com/user/trusty-ci-environment/ looks like adding
two lines to our build could change that. If that pans out, any objections?

tsteven4:
I think we still have users building against qt4, I think we had recent mail regarding this.  But, if we want to leave them behind then we can go to be beta trusty environment on travis.  There is no container version for trusty, and the builds take several minutes to get things installed.  There is a confused branch containing a .travis.yml to do this at https://github.com/tsteven4/gpsbabel/blob/travis/.travis.yml

we won't need the package removals if we take the *-qt4 versions out of configure.in.

qt5 isn't installed by default.  I am hung up on the gui.  I will try to work through the package approval process as suggested.

> Disallowing packages: qttranslations5-l10n
> If you require these packages, please review the package approval process at: 
> https://github.com/travis-ci/apt-package-whitelist#package-approval-process

otrichet:
With all your comments, I created a new version of this change. It uses qtserialport if it is available, otherwise it keeps using the current code (except for windows where the current code is always used).

You can see the change at https://github.com/otrichet/gpsbabel/commit/0ab7e8efb34376236e71a0c444811022868eee0b

Let me know if that would be better for you.

robertlipe:
I'm on the road, but I think I like it. I'm a bit clammy to (non-trivial)
source in *.h vs. *.cc. We surely have inconsistencies in this, especially
in the GUI.

Why did you keep Windows on the old path?  ("Because I can't build/test on
Windows." is a perfectly reasonable answer.) While I have some sympathy for
Linux users with old software shipped by their vendors, here the tool
poverty of Windows works in our favor - it's quite reasonable to expect
developers to have new-ish Qt, I think, because they probably had to
download it.

Maybe we kill the OS/X and Windows versions totally and have only QtSerial
and what's currently in serial_unix.cc as a fallback. Maybe we #if Windows
|| Q_OS_MACX #error in that file because it's not like /dev/ttyS0 is ever a
viable answer there.  The FreeBSD/NetBSD crowd hasn't fussed about our
(admittedly under-ambitious) dev node enumeration in that file so far and
QtSerial is likely to be only a win for them anyway.

On the third hand, this is a compile-time thing, not a runtime thing.
Anyone running a desktoppy Fedora or Ubuntu system is already running a
modern Qt. Do we have any evidence that anyone is building on the LTS
releases, especially building and running the GUI?

Overall, I like this a lot better than the libudev approach.  Thanx for
working through this and talking it out.

RJL

robertlipe:
OP has gone quiet, so this thread went cold, but I like this version so I'm merging it.

PR #47, opened 2016-07-17
added output unit in description

JakobMiksch:
The output unit is meter per second. ( if I understand correctly)
This should be mentioned in the description


xmldoc/filters/options/track-speed.xml
```diff
@@ -5,5 +5,5 @@ speed information or for trackoints synthesized by the
 <link linkend="filter_interpolate">interpolate</link>
 filter.  The speed at each trackpoint is the average speed from the
 previous trackpoint (distance divided by time).  The first trackpoint 
-in each track is assigned a speed of "unknown."
+in each track is assigned a speed of "unknown." The output unit of the speed is meter per second.
```
robertlipe, line 8: "meters"

robertlipe:
Thanx. I've committed a slight improvement to this.

PR #50, opened 2016-08-08
Speed up the download from MTK Logger by requesting multiple chunks at once

yehorov:
In this modification of the mtk format the option block_size was added.
It allows to specify block size per request to increase download speed.

For example, download of 5022K track log from the Qstarz BT-Q1000XT device with 65536 bytes blocks takes 1 minute 5 second, while download of the same log with current 1024 blocks takes 2 minutes.


mtk_logger.cc
```diff
@@ -227,6 +227,7 @@ static char* OPT_erase;  /* erase ? command option */
 static char* OPT_erase_only;  /* erase_only ? command option */
 static char* OPT_log_enable;  /* enable ? command option */
 static char* csv_file; /* csv ? command option */
+static char* OPT_block_size; /* block_size ? command option */
```
robertlipe: Does the block size have to be powers of 2 or multiples of 1024 or have any other limits or can it be free form?

yehorov: http://www.rigacci.org/wiki/lib/exe/fetch.php/doc/appunti/hardware/gps/mtk_logger_library_user_manual_1.2_tsi.pdf
According to the documentation an MTK data may be requested with any even length.
If the requested length is larger than 0x800, an MTK device returns several output NMEAs with the maximum size 0x800 bytes.

In current implementation of the mtk_logger module the returned chunk that contains only 0xff bytes is used as a marker for the end of the MTK data.
So the reminder of the requested length by 0x800 should be large enough to avoid false recognition for the end of the data.

The sequences of 0xff bytes are most common for the Sector Pattern (the header of the 64K sector) that may contains up to 486 bytes and for the tail of the 64k sector with length that not enough for storing a GPS log record.
But I don't know exactly what is the maximum size of a sequence that can appears.

I think it's reasonable to assume that there's no problem with current request size equals to 1024 and restrict the block_size to be a multiple of 1024.

robertlipe: Should the argument then perhaps be "block size in KB", accepting an
integer from 1-64 to make it impossible for users to pick odd values?
Bounds check it from 1-64 in the code and let the max min values in the
vecs_t entry so it plays nice with the GUI.

yehorov: You are right. It will be better to avoid to input a wrong value.

yehorov:
Could anyone check CI?
I believe that checks failed not because of the latest commit.

tsteven4:
I think travis will be happier if you can update your pr to https://github.com/gpsbabel/gpsbabel/commit/4ab5cc0ecf638c77b7e488a3c3e689b0f37d9aa0

As it stands travis merges up with that pr, but only after the apt packages listed in .travis.yml are installed, and it needs a new .travis.yml to figure out that there is a new package to install.

PR #52, opened 2016-08-28
garmin_fit: Added allpoints option

tovisen:
Added allpoints option to convert all points even if they do not have latitude or longitude (primarily intended for devices without gps, e.g., Garmin Vivosmart HR)


testo.d/garmin_fit.test
```diff
@@ -13,3 +13,6 @@ compare ${REFERENCE}/track/garmin-edge-800-output.gpx ${TMPDIR}/fit-sample-800.g
 
 gpsbabel -i garmin_fit -f ${REFERENCE}/track/garmin-forerunner-10.fit -o gpx -F ${TMPDIR}/fit-sample-10.gpx
 compare ${REFERENCE}/track/garmin-forerunner-10-output.gpx ${TMPDIR}/fit-sample-10.gpx
+
+gpsbabel -i garmin_fit,allpoints -f ${REFERENCE}/track/garmin-vivosmart-hr.fit -o gpx,garminextensions -F ${TMPDIR}/fit-sample-vivosmart-hr.gpx
```
robertlipe, line 17: Please include a paragraph in xmldoc/formats/options/....xml explaining what this option does, why one would wish to use it, etc.  We initially did this because people were using their bikes on rollers inside sometimes and without a GPS fix, we made a big mess of their data.Thanx!

tovisen: Ah, ok, that's why. I guessed that there was some reason for the points to be excluded, so I did my change as an option to not change the behavior in the default case.

I have added a short description of the option in xmldoc/formats/options/garmin_fit-allpoints.xml.

robertlipe:
Merged.  Thanx for your contribution.

tovisen:
Thank you for the review and the acceptance of the pull request.

Issue #53, opened 2016-09-08
Provide support for exporting to GeoJSON


ajmas:
It would be useful to be able to export directly to GeoJSON. Currently I need to convert to GPX and then with another application GPX to GeoJSON. Having native GeoJSON would be useful.

GeoJSON is covered by [RFC-7946](https://tools.ietf.org/html/rfc7946)

robertlipe:
I looked into this some time ago. Now that we have a Qt floor that includes
QJsonDocument (I think) the Points case would be super straight-forward and
a lovely starter project for someone wanting to get their feet wet. Tracks
and Routes would map to MultiLineStrings easily. There would be some
special casing for their weird handling at the antimeridian, but I'd bet we
could ship millions of copies before anyone noticed we got that wrong.

We don't really have concepts of Polygons and such. We could possibly
special case if first and end points were the same. We have a few formats
that have this concept (KML, shapefile) but it's not fundamental to us
given the "GPS" in "GPSBabel".

Anyone wanting to tackle this, please sign up.

GPSBabelDeveloper:
For waypoints, you can sort of see what this would look like with commit 4ce284a It probably doesn't validate, but the output for trivial waypoints (no alt) sure looks pretty GeoJSON-y. The serialization is probably not super memory efficient, but I'm not sure it has to be. Do you (or anyone else) want to run with it? My (scarce) nights and weekends are better devoted to fixing the problems in the 1.5.3 release I made in Jan (sigh) and rolling up the fixes to date over a feature requested by one user.  There may have been a second, but this has been on my distant radar for a while...

GPSBabelDeveloper:
Added in https://github.com/gpsbabel/gpsbabel/commit/ecbf04080c7f5107ffe29dc6c40c80046cb732c8 and a few fast-follow submits to fix issues with older Qt.

HarelM:
What about conversion from GeoJson to other formats? I see that the tests in the above commit are only referring to GeoJson as Output...?
Since this was not supported when I started using GPSBabel I wrote it (in C#).
If this helps feel free to translate the code (conversion from GPX to GeoJson):
https://github.com/IsraelHikingMap/Site/blob/master/IsraelHiking.API/Converters/GpxGeoJsonConverter.cs
It uses LINQ and some C# specific operators (like ?? which is a null check + assignment) which reduces the amount of code a bit but the general idea is there.

robertlipe:
[...]

PR #63, opened 2017-01-23
Added geojson read capablity, 

HarelM:
moved magic strings to constants, fixed windows compilation issues.
Related to:
Provide support for exporting to GeoJSON #53


GPSBabel.pro
```diff
@@ -148,7 +148,7 @@ win32 {
   }
   SOURCES += gbser_win.cc
   JEEPS += jeeps/gpsusbwin.cc
-  LIBS += "C:/Program Files/Windows Kits/8.0/Lib/win8/um/x86/setupapi.lib" "C:/Program Files/Windows Kits/8.0/Lib/win8/um/x86/hid.lib"
+  LIBS += "C:/Program Files (x86)/Windows Kits/8.1/lib/winv6.3/um/x64/setupapi.lib" "C:/Program Files (x86)/Windows Kits/8.1/lib/winv6.3/um/x64/hid.lib"
```
GPSBabelDeveloper, line 151: I hate that we have to hardcode pathnames like this as this will certainly break builds on my system. Now that Delorme is dead, I don't think we need HID any more.

GPSBabelDeveloper (changes requested):
Overall, this looks quite good for someone nervous about C++. Thank you. 

I'll run it through clang-format after commit to make the styles consistent.  There's some non-obvious tweaks I'll make, too. For example, I think your use of loop iterators results in an extra copy. (No, I don't think thost microseconds actually matter.) I'll look at those after commit.

Please add a change to geojson_vecs so that it shows it can read routes and tracks. Review https://www.gpsbabel.org/htmldoc-1.5.4/fmt_geojson.html and tweak xmldoc/formats/geojson.xml (don't fret about angle brackets - the words are more important than the xmldoc markup) also please provide a json file that exercises the reader. in reference/tracks.  I'll then convert that to, say, GPX (KML?) and tweak testo.d/geojson.test so that it'll become part of the automated regression test.

GPSBabel.pro
HarelM: I tried to use ```$(WindowsSDK_LibraryPath_x64)``` but there's a bug in visual studio preventing in from working, I believe it can be solved properly but I don't have the time to play around with it. Also I figured that no one is using a non 64bit windows system nowadays. and 8.1 replaced 8.0 so I figured this is a safer bet...

HarelM:
Updated according to code review comments. Not sure I updated the docs right, feel free to rewrite it.

GPSBabelDeveloper:
Thank you. I'll clean up the style later.

HarelM:
Is it possible to create a release for this after you clean it up, or should I use the locally compiled version?

robertlipe:
Releases are expensive for us.  Use your locally built version.

PR #68, opened 2017-03-07
Fixed altitude in interpolate filter

cj4567:
The interpolate filter currently interpolates only longitude and latitude. Fixed it so it also handles altitude.

### DIFF
--- a/interpolate.cc
+++ b/interpolate.cc
@@ -61,11 +61,13 @@ interpfilt_process(void)
   int count = 0;
   int first = 0;
   double lat1 = 0, lon1 = 0;
+  double altitude1 = unknown_alt;
   unsigned int time1 = 0;
   unsigned int timen;
   double distn;
   double curdist;
   double rt1, rn1, rt2, rn2;
+  double frac;
 
   if (opt_route) {
     route_backup(&count, &backuproute);
@@ -108,12 +110,14 @@ interpfilt_process(void)
             wpt_new->shortname = QString();
             wpt_new->description = QString();
 
+            frac = (double)(timen - time1) / (double)(wpt->creation_time.toTime_t() - time1);
             linepart(lat1, lon1,
                      wpt->latitude, wpt->longitude,
-                     (double)(timen-time1)/
-                     (double)(wpt->creation_time.toTime_t() - time1),
+                     frac,
                      &wpt_new->latitude,
                      &wpt_new->longitude);
+            if (altitude1 != unknown_alt && wpt->altitude != unknown_alt)
+              wpt_new->altitude = altitude1 + frac * (wpt->altitude - altitude1);
             if (opt_route) {
               route_add_wpt(rte_new, wpt_new);
             } else {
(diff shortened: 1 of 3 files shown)


reference/track/tinterptrack.gpx
```diff
@@ -1,53 +1,53 @@
...
-        <time>2002-05-25T17:06:09Z</time>
-      </trkpt>
-      <trkpt lat="36.000000000" lon="-115.000000000">
-        <ele>0.000000</ele>
-        <time>2002-05-25T17:06:10Z</time>
-      </trkpt>
-    </trkseg>
-  </trk>
-</gpx>
+<?xml version="1.0" encoding="UTF-8"?>
```
GPSBabelDeveloper: No carriage returns in the repo, please.

reference/track/interptrack.gpx
```diff
@@ -1,149 +1,149 @@
...
-        <time>2002-05-25T17:06:09Z</time>
-      </trkpt>
-      <trkpt lat="36.000000000" lon="-115.000000000">
-        <ele>0.000000</ele>
-        <time>2002-05-25T17:06:10Z</time>
-      </trkpt>
-    </trkseg>
-  </trk>
-</gpx>
+<?xml version="1.0" encoding="UTF-8"?>
```
GPSBabelDeveloper: No carriage returns in the repo, please.

interpolate.cc
```diff
@@ -108,12 +110,14 @@ interpfilt_process(void)
             wpt_new->shortname = QString();
             wpt_new->description = QString();
 
+            frac = (double)(timen - time1) / (double)(wpt->creation_time.toTime_t() - time1);
```
GPSBabelDeveloper, line 113: This probably breaks down for time < 1970 or > 2038 or something)  This problem isn't new and it's not your fault. Anyplace that a double is used to hold a time is probably slightly wrong these days. You could probably make frac a QDateTime and eliminate the toTime_t() and rely on the overloaded +/- operators to DTRT. If that doesn't fall out within a couple minutes, that's cool; I won't push back more than a bit on that.

GPSBabelDeveloper (changes requested):
Nice! Shows a good understanding of the problem. Thanx. If changing that to a QDateTime doesn't fall out, I'll take it after the CR's are removed.

interpolate.cc
cj4567: frac is just between 0 and 1 here and does not hold the time. You're right though, toTime_t() will break before 1970 and after 2106. Replaced it with toSecsSinceEpoch().

reference/track/interptrack.gpx
cj4567: Sorry, fixed it.

interpolate.cc
cj4567: Never mind, that's only available in newer qt versions...

GPSBabelDeveloper (approved):
That's due diligence. Thanx for looking into it.

PR #71, opened 2017-06-12
Master

robertlipe:
Prepare for more dynamic format vets.

Move the format arrays from static C arrays to container classes so they can be changed at runtime.


robertlipe:
Thank you, git, for uploading 3,811 more changes than I intended to...

PR #72, opened 2017-06-12
Move format vectors into container classes to prep for dynamic formats.

robertlipe:
Change from two arrays for the format vectors to be closer to one.


robertlipe:
Subsumed by TSteven4's Format class work.

PR #73, opened 2017-06-16
Various cleanups to configure

turboencabulator:
Rename configure.in to configure.ac to avoid a warning.
Use portability macros instead of shell code.


GPSBabelDeveloper (commented):
Hi, and welcome. I'll admit to being "out of the loop" of autoconfiscation, while I was less so in the early 90's, but can you help me (us) understand the motivation here? This looks like a nearly mechanical (and that's OK!) replacement of some of the shell constructs to use more obtuse m4 style constructs, but fundamentally, it's still a shell script but harder to read. What's the motivation/benefit to abstracting 'if' into 'AS_IF'. https://giphy.com/gifs/filmeditor-clueless-movie-3o7aTIGlhSo1bL8QUg ? It's not 'if' isn't pretty universal between any of the shells we encounter in the wild.

Oh, we probably don't want HAVE_CXX11. We require CXX11 - we kind of coasted into that and probably didn't capture that requirement correctly in configure since everything we've cared about has supported the parts we care (auto, range for) about for years.

turboencabulator:
So I started updating it to use AS_HELP_STRING to fix the formatting of ./configure --help...  then went all-in with AS_IF and AS_CASE.  A couple reasons:  Those macros purportedly exist to handle portability issues, so we might as well gain those benefits.  Also when the file is a mixture of about 2/3 macros and 1/3 shell, it tends to confuse or break syntax highlighting.  Having the file be nearly all macros usually helps with that, and gives the file a more consistent look.

HAVE_CXX11 is provided by ax_cxx_compile_stdcxx_11.m4.  It looks like autoheader was never run after that script was added, otherwise you'd have noticed it earlier.  Usually the purpose of the HAVE_xyz stuff in config.h is so that you can code up some alternatives if you want to support users that don't have that feature.  But there is no code like that in the project, so you can ignore it like you ignore most of the other HAVE_xyz macros.

BTW, I'm trying out a newer version of that ax_cxx_compile_stdcxx_11.m4 script to see if it produces anything different.  I don't have it working yet.

robertlipe:
Thanx. It seems a bit weird of an abstraction as we've received zero complaints about shells in the wild and my general experience is to avoid m4 as much as possible, but if this is the way the cool kids are doing it, OK.

PR #74, opened 2017-06-23
More autoconf modernizations

turboencabulator:
Plus a few other cleanups.


GPSBabelDeveloper (approved):
Looks good. Thanx.

I really do think that our use of autoconf and maybe even Makefile are approaching EOL. We just don't see the variety of target OSes that we did. A 200Kline 'configure' script is just not funny. Checking for the presence of <string.h> every time in 2017 when you're already requiring C++11 is pretty silly.

Issue #65, opened 2017-01-31
Compatibility with FIT 2.0 format


YousefED:
Gpsbabel breaks when provided a 2+ versioned FIT file. I'm not sure how much has changed from version 1 to version 2, but I can provide a couple of sample files if this helps.

Any plans to support the latest fit formats?

robertlipe:
FIT is a reverse engineered format for us. There was a spec published, but
it contained such onerous restrictions that using it in an open source
program wasn't feasible. I haven't reviewed it in the last few years to see
if that's still the case.

You're the first to mention it that I know of and I don't know of anyone
working on an implementation, so if you're planning to do so, let's assign
this to you.

RJL

GPSBabelDeveloper:
Fix committed in https://github.com/gpsbabel/gpsbabel/pull/75

Issue #54, opened 2016-09-30
FIT to GPX conversion ignores base time and provides times in 1989


ajmas:
When I try converting a Garmin FIT file to GPX I find all the times in the file are in 1989, rather than the date recorded.

Looking at the specification, it would seem that we should probably be reading the time_created value in the file_id message and adding that to the timestamp value, if timestamp is less than 0x10000000.

Command line:

`gpsbabel -i garmin_fit -f "2016-09-09-10-29-53.fit" -o gpx -F 2016-09-09-10-29-53.gpx`

I can provide data files on request (about 1Mb in size for the .fit).

Was testing with gpsbabel 1.5.3_1, installed with MacPorts, on MacOS X 10.12

Looking at the 'D00001309 FIT File Types Description Rev 2.1' in the FitSDKRelease_20.10.01, for specs.

GPSBabelDeveloper:
The license for the FIT SDK isn't compatible with open source projects like ours. (At least it wasn't when I last looked.) We have a fair number of .fit users, so it seems odd that this is totally broken.  If you can provide a detailed textual description of how to make this better, why your fit files are different than the ones we have, perhaps referencing https://github.com/gpsbabel/gpsbabel/blob/master/garmin_fit.cc (hint: search for 'timestamp') perhaps someone can take a swing at improving it.

dark:
This was fixed by PR #75 too.

PR #75, opened 2017-06-24
Support v2 of the Garmin FIT format

dark:
As briefly mentioned in #65, there have been recent updates to the Garmin FIT format. One of those (#54) breaks the correct interpretation of data point timestamps by gpsbabel.

This patchset includes several fixes to support the most recent version of Garmin FIT, while maintaining backwards compatibility. These changes have been tested against FIT files created by a Garmin VIRB Ultra 30.

* Fix a pre-existing bug in the switch logic (+ fix tests)
* Add support for two new fields: enhanced speed and enhanced altitude
* Fix handling of timestamp fields (#54)
* Use constants rather than magic numbers for better code readablity

### DIFF
--- a/garmin_fit.cc
+++ b/garmin_fit.cc
@@ -57,6 +57,7 @@ static struct {
   int endian;
   route_head* track;
   uint32_t last_timestamp;
+  uint32_t global_utc_offset;
   fit_message_def message_def[16];
 } fit_data;
 
@@ -135,6 +136,8 @@ fit_parse_header(void)
     // Unused according to Ingo Arndt
     gbfgetuint16(fin);
   }
+
+  fit_data.global_utc_offset = 0;
 }
 
 static uint8_t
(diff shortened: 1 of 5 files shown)


dark:
Thanks for the review. I can push another commit tonight / tomorrow with the changes you asked for (i.e. using global integer constants for the IDs/fields rather than C-style defines).

robertlipe:
const ints, enums, or even class enums would be lovely, if you don't mind.
(C++ enums actually work for strong typechecking, unlike C enums which
decompose to an int almost immediately. That said, in a smallish project
like this and where such things tend to have single file scope class, I can
certainly see MyEnum::FooEnabled just being busy decoration.) If you DO
mind, I'll gratefully accept this unchanged and change it myself.

I'll readily admit that there is a strong C spelling of things over C++ as
we were originally in C. I'll also readily admit that our style is
inconsistent and it looks like it was written by dozens of programmers that
never spoke to each other because, well, it was. :-)

dark:
I don't mind :)
I'll go for `const int`s, since C++ enums require a few casts and that would probably make the code a bit uglier (everything is an int as of now).

dark:
Whoops, some of the previous review comments were lost due to some refactoring / rebasing I did to straighten up the branch history. Oh well :)

GPSBabelDeveloper:
Awesome. Thank you.

pseyfert:
I'm confused, the title of this pull request says it would support garmin fit format version 2, but the check in garmin_fit.cc is still in place and rejects version 2.0:

```cpp
ver = gbfgetc(fin);
if (ver == EOF || (ver >> 4) > 1)
  fatal(MYNAME ": Unsupported protocol version %d.%d\n",
        ver >> 4, ver & 0xf);
if (global_opts.debug_level >= 1) {
  debug_print(1,"%s: protocol version=%d\n", MYNAME, ver);
}
```

dark:
That's odd. I was assuming the camera I own was creating files with the correct v2 header. Or maybe it's not v2 at all.

pseyfert:
So for checking I created a small fit file here https://github.com/pseyfert/fitfile-jugglers/tree/master/org-fit-files

I made it with the "Professional skiing app" (free trial version) https://apps.garmin.com/en-CA/apps/47e87496-459b-46a0-8b9d-c344cb0d1df9 and for reference used `python-fitparse` to convert it to human readable text. my copy of gpsbabel rejects the file with `fit: Unsupported protocol version 2.0`.

0bscur3:
I also tried to convert fit files from my Wahoo element, which I downloaded from Strava. I also get the same error. Are there any news on this. I am using gpsbabel version 1.5.4.

tsteven4:
@pseyfert, the 2.0 check was fixed in #163 which was merged Apr 14, 2018.  Please either build from source or try a pre-release version at https://github.com/gpsbabel/gpsbabel/releases "Continuous build".  There is a continuous build available for mac and windows.

@0bscur3, 1.5.4 does not have the fix, please build from source or try a pre-release version at https://github.com/gpsbabel/gpsbabel/releases "Continuous build".  There is a continuous build available for mac and windows.

pseyfert:
Thanks, I can confirm building gpsbabel from master works for me.

PR #79, opened 2017-07-01
Parsing safety for pcx files

languitar:
There are also several parsing safety issues in the pcx parser. With this PR I have tried to address the most obvious ones. I am not completely sure whether I have accidentally changed some parsing as I didn't understand the testing infrastructure.

There are still a few cases that this doesn't catch. Test cases are attached.
[pcx-crashes.zip](https://github.com/gpsbabel/gpsbabel/files/1117398/pcx-crashes.zip)

### DIFF
--- a/pcx.cc
+++ b/pcx.cc
@@ -123,15 +123,20 @@ data_read(void)
       time[0] = 0;
       date[0] = 0;
       desc[0] = 0;
-      alt = -9999;
-      sscanf(ibuf, "W  %6c %s %s %s %s %ld",
+      alt = 0;
+      n = sscanf(ibuf, "W  %6c %s %s %s %s %ld",
              name, tbuf, nbuf, date, time, &alt);
-      if (alt == -9999) {
+      if (n < 5) {
+        fatal(MYNAME ":Unable to parse waypoint, not all required columns contained\n");
+      }
+      if (n != 6) {
         alt = unknown_alt;
       }
 
       if (comment_col) {
-        strncpy(desc, &ibuf[comment_col], sizeof(desc)-1);
+        if (snprintf(desc, sizeof(desc), "%s", &ibuf[comment_col]) < 0) {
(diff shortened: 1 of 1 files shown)


robertlipe:
Thanx. I see some other low hanging fruit regarding fixed size buffers, so I'll pull this in and carry on.

To run the test for a single format (usually super fast - awesome for development) just run
./testo FORMAT_NAME
so
./testo pcx 
would run the pcx test. If you need to see what commands are invoked when this is run, look into testo.d/pcx.test which is a shell script fragment that invokes GPSBabel in lots of interesting ways.

./testo runs everything for all formats. On my laptop, it takes under 30 seconds.  The tests need a bourne-like shell, so they run easily on UNIX-y systems like MacOS, Linux, the BSDs, etc. and probably need something like Cygwin for Windows, though I understand that a Linux-y subsystem is available for Windows 10. I have no personal experience with that as I try to be as Windows-free as I possibly can.

The monster test (vtesto) runs under valgrind (tests for use-after-frees, writes past end of buffers, etc.) and runs every combination of input and output. By getting a a "all checks have passed", that's actually done on the Linux system via Travis continuous integration.

robertlipe:
https://github.com/gpsbabel/gpsbabel/commit/b500587ef215d4e24749a5af44ce3eb0ad85d999 should be much better.

I can't say I'm proud of how deep into a rewrite rathole I went into this format given that it seemed to have died out over ten years ago.

GPSBabelDeveloper:
Rats. I've angered the valgrind gods.

I'm 3 CPU hours into identifying the problem. I know if I boot Parallels
and wait for all the required Linux updates to download and install, it'll
be that again. Eight cores and a gigabit connection shouldn't be so
frustrating., I'll pick that back up tomorrow.

Do any of you have Docker experience? Is this the kind of thing that Docker
is good at? Do any of you have recipes for valgrinding GPSBabel on Docker?

I'll fix it one way or another.

PR #82, opened 2017-08-24
Extend KML, so that <TimeSpan> can be used to describe the time span

bubeck:
of a track (which does not have TimeStamps).


tsteven4:
Welcome & Thanks.

Could you please supply a test case?  The script testo runs our basic 
tests.  testo.d/kml.test will have the existing kml tests. The testcase 
will include a reference kml file with TimeSpan elements as well as a 
reference for the expected output in the format of your choice.  I will 
manually verify that the input reference is valid, i.e. does not violate 
the kml schema.

Steve

bubeck:
Thanks for the "Welcome" :)

Test supplied as requested. If there is anything in addition, feel free to let me know.

Tilmann

tsteven4:
Thanks for the test case.  Adding test coverage reduces the chances that we will break the functionality you have added.

I do have some issues with the testcase.  Apparently the SkyDrop vario is is using the namespace http://earth.google.com/kml/2.0.  This version of kml is quite old and documentation is hard to come by at this point.  But it would appear that some of the elements used are not part of kml 2.0.  All the elements are present in kml 2.1/2.2.  Kml 2.0 also does not have a xsd document available, making it hard to validate.  Is it possible to get an update to the SkyDrop vario that would produce kml 2.2?  It seems they do regularly issue updates.  To do this a few things have to change:

1) the namespace needs to change to http://www.opengis.net/kml/2.2

2) the original reference violates the kml 2.2 schema.  the TimeSpan element should be a child of the Placemark element, not the LineString element.  The TimeSpan element should follow the open element and precede the Style element.  This move will require a change in kml.cc in gpsbabel.

3) the original version violates the kml 2.2 schema.  the alititudeMode element should follow the extrude element.  This isn't an issue for gpsbabel.

An updated reference is here:
[skydrop_kml22.kml.txt](https://github.com/gpsbabel/gpsbabel/files/1267809/skydrop_kml22.kml.txt)

Additionally, your modifications to kml.cc introduce a bug.  If the kml file contains two Placemark elements, where the first has a TimeSpan but the second does not then the TimeSpan data from the first Placemark will be used for both.

As a Skydrop user can you approach them and request the kml changes?

I would be happy to discuss these issues further.

bubeck:
Thanks for your detailed and very good answer!

As I am also an active submitter to SkyDrop (also open source), this should be no problem. I will first make a change to SkyDrop to generate proper KML 2.2 according to your suggestions and then do the proper handling in gpsbabel's kml.cc.

I will come back with this to you, but it will probably take some weeks.

Thanks!

robertlipe:
Good stuff, fellows.

KML 2.0 was "what Earth did" - very informal and not really externally
reviewed. 2.1 had more attention outside of the Earth developers and
community, but 2.2 is almost ten years old. It was the first version that
was really formalized and scrutinized because it's the one submitted to the
Open Geospatial Consortium (kind of the ANSII or ISO of this world) to be
formalized.  2.3 was just a few years ago and a pretty minor bump as I
think it mostly formalized the Track extensions from Earth 5.2 back in 09
or 2010. It's not like that's a crazy fast moving target.

If it's the writer at
https://github.com/fhorinek/SkyDrop/blob/63da36f11ceb94bf510d4b8741ddbcc2f2fb51e1/skydrop/src/fc/logger/kml.cpp
it looks pretty straight-forward to change. I don't see a reader and that
writer looks pretty much like ours did in the early days; you just kind of
mocked up what Earth itself wrote. (Before validation, I'd frequently knock
up something that worked in Earth, load it into Earth, save it in Earth as
KML, and then make my writer do what Earth did since it was the golden
standard of KML.) You'd probably be doing a favor to readers of any KML
that program generates to run it through a validator once in while.

It's a bit surprising to find another IGC implementation...Interesting.

tsteven4:
I recommended kml 2.2 instead of 2.3 because
1) public google documentation is still 2.2
2) google earth still produces 2.2
3) my validator is based on Xerces C++ which doesn't support XML schema 1.1 which is used starting with kml 2.3.  Support is available with Xerces2 java, but I haven't migrated to that yet.

bubeck:
Thanks again.

I will now prepare the changes to Skydrop's kml logger according to the above provided example file. The device only writes the flight 
that you are currently doing, so there is no reader. Because the device 
only has 192 k of flash ram, everything must be as compact as possible. The link is correct.

How would a validation be done at best? Do you have any suggestions? 
Just using the schema on a number of example files generated by skydrop?

Thanks again, your help is highly appreciated.

bubeck:
* Bug fixed
* test reference "skydrop.kml" updated to KML 2.2.

(Patch for skydrop is underway to https://github.com/fhorinek/SkyDrop)

bubeck:
SkyDrop's pull request to produce KML 2.2 is at https://github.com/fhorinek/SkyDrop/pull/300
Thanks again for the advice.

kml.cc
```diff
@@ -288,13 +290,21 @@ const char* kml_tags_to_ignore[] = {
...
   if (wpt_tmp) {
     fatal(MYNAME ": wpt_s: invalid kml file\n");
   }
   wpt_tmp = new Waypoint;
   wpt_tmp_queued = 0;
+
+  /* Invalidate timespan elements for a beginning Placemark,
+   * so that each Placemark has its own (or no) TimeSpan. */
+  wpt_timespan_begin.setTime_t(0);
+  wpt_timespan_end.setTime_t(0);
```
tsteven4: Can you use the gpsbabel::DateTime() constructor here instead of setTime_t(0)?  This encapsulates the workings of gpsbabel::DateTime better, and will avoid trouble in the unlikely event that the crutch is removed.
`  // As a crutch, mimic the old behaviour of an uninitialized creation time
  // being 1/1/1970.
`
For example:
gtrnctr.cc:  gtc_least_time = gpsbabel::DateTime();
gtrnctr.cc:  gtc_most_time = gpsbabel::DateTime();

kml.cc
```diff
@@ -392,6 +415,35 @@ void trk_coord(xg_string args, const QXmlStreamAttributes*)
...
+   * </TimeSpan>
+   *
+   * If this is specified, then SetCreationDate
+   */
+  if ( wpt_timespan_begin.isValid() && wpt_timespan_end.isValid() ) {
+
+	  // Count the number of Waypoints
+	  n = 0;
+	  QUEUE_FOR_EACH(&trk_head->waypoint_list, curr_elem, tmp_elem) {
+		  n++;
```
GPSBabelDeveloper: Consider n = trk_head.rte_waypt_ct;
This is admittedly poorly named and not obvious and not the very "C++" way of doing things, but see if that plays out well.

kml.cc
```diff
@@ -392,6 +415,35 @@ void trk_coord(xg_string args, const QXmlStreamAttributes*)
...
+	  }
+
+	  // If there are some Waypoints, then distribute the TimeSpan to all Waypoints
+	  if ( n > 0 ) {
+		  qint64 timespan_ms = wpt_timespan_end.toMSecsSinceEpoch() - wpt_timespan_begin.toMSecsSinceEpoch();
+		  qint64 ms_per_waypoint = timespan_ms / n;
+		  QUEUE_FOR_EACH(&trk_head->waypoint_list, curr_elem, tmp_elem) {
+			  trkpt = (Waypoint*) curr_elem;
+			  trkpt->SetCreationTime(wpt_timespan_begin);
```
GPSBabelDeveloper: Could this be done while we're reading above to not make another pass over the queue?

kml.cc
```diff
@@ -392,6 +415,35 @@ void trk_coord(xg_string args, const QXmlStreamAttributes*)
...
+
+	  // Count the number of Waypoints
+	  n = 0;
+	  QUEUE_FOR_EACH(&trk_head->waypoint_list, curr_elem, tmp_elem) {
+		  n++;
+	  }
+
+	  // If there are some Waypoints, then distribute the TimeSpan to all Waypoints
+	  if ( n > 0 ) {
+		  qint64 timespan_ms = wpt_timespan_end.toMSecsSinceEpoch() - wpt_timespan_begin.toMSecsSinceEpoch();
```
GPSBabelDeveloper: There isn't an operator-, oddly, but the msecsTo method may play nicer with large timespans.
http://doc.qt.io/qt-5/qdatetime.html#msecsTo

bubeck:
Thanks again for the feedback, which I followed to further improve the patch.
Please merge. :) Thanks!

robertlipe:
This is probably about spot-on now. Can you please do a 'git merge' (maybe
it's a 'git pull') to resynchronize with base? When we look at the CL
you're submitting, its showing every file that's changed in the last
several weeks.

Thanx for humoring us.
RJL

bubeck:
Closed in favor of https://github.com/gpsbabel/gpsbabel/pull/115

PR #83, opened 2017-09-03
Fix building with the latest qt5 version.

bzed:


gpx.cc
```diff
@@ -494,7 +494,7 @@ tag_cache_desc(const QXmlStreamAttributes& attr)
 {
   cache_descr_is_html = 0;
   if (attr.hasAttribute("html")) {
-    if (attr.value("html").toString().compare("True") == 0) {
+    if (attr.value("html").toString().compare(QString("True")) == 0) {
```
GPSBabelDeveloper, line 497: QStringLiteral() here (and throughout) would save some alloc/free cycles.

Also, please report which qt5 is "latest" in your commit message.

gpx.cc
```diff
@@ -972,7 +972,7 @@ gpx_end(const QString& el)
      * last date we saw in this log.
      */
   case tt_cache_log_type:
-    if ((cdatastr.compare("Found it") == 0) &&
+    if ((cdatastr.compare(QString("Found it")) == 0) &&
```
GPSBabelDeveloper, line 975: Most of these look like mechanical conversions from C. Would they be more readable with the cdatastr == QStringLiteral() form since QString has overloaded operator == ?

GPSBabelDeveloper (changes requested):

bzed:
I don't know the details, but it seems that qt 5.8 added a new static overload of QStringRef::compare() - which seems to make some trouble. Didn't spend much time to investigate the details, just made gpsdbabel build in debian untsable again.

robertlipe:
Bernd, does this work for you?
https://github.com/gpsbabel/gpsbabel/commit/30a345577a644125502f6e4f8b58ea3077ec8dad

zstadler:
With the new support for QT 5.7, perhaps this issue can be closed?

https://github.com/gpsbabel/gpsbabel/blob/1d439235ab5ff272fbfc859c19fd35356293e651/configure#L5786-L5787

It seems like [QT 5.9 is also working fine](https://ci.appveyor.com/project/gpsbabel/gpsbabel/build/1.0.209), at least on Windows.

PR #86, opened 2017-09-26
Add Unicode file path argument support to Windows build.

viettaml:
Use QCoreApplication::arguments() to properly ingest Unicode arguments and convert them to UTF-8. Remove use of QString::fromLocal8Bit(), as file path arguments are no longer "local 8-bit" on Windows (they're UTF-8) and "local 8-bit" on Mac and Linux means UTF-8.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -241,7 +241,15 @@ main(int argc, char* argv[])
   // For example, this will get the QTextCodec::codecForLocale set
   // correctly.
   QCoreApplication app(argc, argv);
-  // TODO: use QCoreApplication::arguments() to process command line.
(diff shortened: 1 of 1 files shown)


robertlipe (commented):
Thanx. This relatively minor change in Qt 5.8 where C->QStrings encoding that didn't have an explicit coding set with it resulted in a silent change is really biting us in unfortunate places. We just need to get more of these C strings moved to (corretly encoded) QStrings and get (and stay!) out of the conversion business.

tsteven4:
Not sure what change to Qt you are talking about.  Do you have a reference?

PR #89, opened 2017-10-01
Use Latin1 when converting from Garmin device.

goeranu:
When copying waypoints from my Garmin eTrex Legend HCx any non-ASCII characters showed up as the Unicode _replacement character_.  I'm not sure when exactly this started, but when digging into it I found in garmin.cc it was converted to QString using assignments.  In Qt5, such assignments assume the C string is in UTF-8, but the string from the device is in Latin-1.  This patch should fix the problem.

While staring at the code, I realized _simplified_ was called twoce on the description string.  As that operation is idempotent that is harmless, but also pretty pointless.

### DIFF
--- a/garmin.cc
+++ b/garmin.cc
@@ -394,8 +394,8 @@ waypt_read(void)
   for (i = 0; i < n; i++) {
     Waypoint* wpt_tmp = new Waypoint;
 
-    wpt_tmp->shortname = way[i]->ident;
-    wpt_tmp->description = QString(way[i]->cmnt).simplified();
+    wpt_tmp->shortname = QString::fromLatin1(way[i]->ident);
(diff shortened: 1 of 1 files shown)


robertlipe:
Sorry for the delay as I've been quite under it. Nice find. Thanx. There are probably variations of this problem all over the Garmin module (reading track names, routes, etc.) so if you or another Babel-head would like to investigate any of these silent C-strings->QStrings, please do feel encouraged.

The Garmin code is particularly nasty as it actually dates back not only to C strings but fixed-length (space-padded, not zero-terminated) buffers. This code has been through a lot on my watch.

This relatively quiet - and hard to detect - change in Qt5 has been a big pain for us.

Thanx!

PR #95, opened 2017-10-26
Sync ICAO country code table with Wikipedia

fbonzon:
Plus minor fixes, see individual commits.


robertlipe (commented):
Thanx. Let's take the leap that whatever program actully uses these codes updates from Wikipedia, too. We don't want any Gaza/Palestinian squabbles to leak into our maps.

PR #97, opened 2017-11-02
Minor Header fixes

robertlipe:
Make more safe for re-inclusion.
Bring more into style for eventual C++ style guide fix.
Drop a few unused prototypes.


robertlipe:
Should auto-close. Thanx, uuuuh, me!

PR #114, opened 2017-11-14
Qstring-ification of main.cc.

tsteven4:
This request moves the command line parsing from char[] to QStrings.
This eliminates some rounds trips between the two string representations.  When necessary we use CSTR to pass a char[] to a legacy function.
Batch processing has been rewritten mostly used Qt classes.
AFL fuzzing was used on argv to look for crashes and hangs.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -17,9 +17,10 @@
 
  */
 
+#include <QtCore/QStack>
 #include <QtCore/QTextCodec>
+#include <QtCore/QTextStream>
 #include <QtCore/QCoreApplication>
-#include <QtCore/QVector>
 
 #include "defs.h"
 #include "filterdefs.h"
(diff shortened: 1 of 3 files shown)


main.cc
```diff
@@ -28,106 +29,84 @@
...
-    str = lrtrim(str);
-    if ((*str == '\0') || (*str == '#')) {
+  QString line;
+  QString str;
+  QStringList qargs(arg0);
+
+// Use a QTextStream to read the batch file.
+// By default that will use codecForLocale().
+// By default Automatic Unicode detection is enabled.
+  gpsbabel::File* file = new gpsbabel::File(filename);
```
GPSBabelDeveloper: gb::File  file(filename);
file.open(...);
QTextStream stream(file);
then, when either goes out of scope, the destructors are called, making it impossible to have dangling pointers, unclosed files, etc.

main.cc
```diff
@@ -28,106 +29,84 @@
...
-  res->argn = argn;
-  res->argc = argc;
-  res->argv = (char**)argv;
-
-  return res;
-}
-
-static arg_stack_t
-* pop_args(arg_stack_t* stack, int* argn, int* argc, char** argv[])
+class QargStackElement
```
GPSBabelDeveloper: Nice use of a lesser known Qt container to replace our organically grown one.

main.cc
```diff
@@ -28,106 +29,84 @@
...
-      line = xstrdup(str);
+    if (line.isEmpty()) {
+      line = str;
     } else {
-      char* tmp;
-      xasprintf(&tmp, "%s %s", line, str);
-      xfree(line);
-      line = tmp;
+      line.append(' ');
+      line.append(str);
```
GPSBabelDeveloper, line 88: Why are we making line basically a QVector or QList that's push_backed here (with special casing for the first) and then calling our C-string parse to break it back apart? Doesn't the bottom half turn into return return line.split()?

main.cc
```diff
@@ -219,46 +198,40 @@ print_extended_info(void)
...
+  AFL_INIT_ARGV();
+#endif
   int c;
   int argn;
   ff_vecs_t* ivecs = NULL;
   ff_vecs_t* ovecs = NULL;
   filter_vecs_t* fvecs = NULL;
-  char* fname = NULL;
-  char* ofname = NULL;
+  QString fname = NULL;
```
GPSBabelDeveloper: No need to explictly set.  If the argument is "for consistency of those around it", I'd move these down closer to first use.

main.cc
```diff
@@ -219,46 +198,40 @@ print_extended_info(void)
...
   int did_something = 0;
   const char* prog_name = argv[0]; /* argv is modified during processing */
   queue* wpt_head_bak, *rte_head_bak, *trk_head_bak;	/* #ifdef UTF8_SUPPORT */
   signed int wpt_ct_bak, rte_ct_bak, trk_ct_bak;	/* #ifdef UTF8_SUPPORT */
-  arg_stack_t* arg_stack = NULL;
+  QStack<QargStackElement> qargs_stack = QStack<QargStackElement>();
 
-  // Create a QCoreApplication object to handle application initialization. 
+  // Create a QCoreApplication object to handle application initialization.
   // TODO: Someday we may actually use this, but for now we are just trying
```
GPSBabelDeveloper: Is that day now? (Or recently?)

main.cc
```diff
@@ -308,8 +281,8 @@ main(int argc, char* argv[])
 #ifdef DEBUG_MEM
   debug_mem_open();
   debug_mem_output("command line: ");
-  for (argn = 1; argn < argc; argn++) {
-    debug_mem_output("%s ", argv[argn]);
+  for (int i = 1; i < qargs.size(); i++) {
+    debug_mem_output("%s ", qPrintable(qargs.at(i)));
```
GPSBabelDeveloper, line 284: Just talking out loud, DEBUG_MEM probably has a short life...maybe measured in days. It's not like having our own leak detector makes sense theses days. So no change needed here - just a note to not go crazy in those blocks.

main.cc
```diff
@@ -325,77 +298,76 @@ main(int argc, char* argv[])
...
-      opt_version = atoi(&argv[argn][2]);
+    if (qargs.at(argn).size() > 2) {
+      opt_version = qargs.at(argn)[2].digitValue();
     }
 
     switch (c) {
     //case 'c':
-    //  optarg = argv[argn][2] ? argv[argn]+2 : argv[++argn];
-    //  cet_convert_init(optarg, 1);
+    //  optarg = qargs.at(argn).size() > 2 ? QString(qargs.at(argn)).remove(0,2) : qargs.size()>(++argn) ? qargs.at(argn) : QString();
```
GPSBabelDeveloper: Let's just drop this.

main.cc
```diff
@@ -325,77 +298,76 @@ main(int argc, char* argv[])
...
   argn = 1;
-  while (argn < argc) {
-    char* optarg;
+  while (argn < qargs.size()) {
+    QString optarg;
 
-    if (argv[argn][0] != '-') {
+//  we must check the length for afl input fuzzing to work.
+//    if (qargs.at(argn)[0].toLatin1() != '-') {
+    if (qargs.at(argn).size() > 0 && qargs.at(argn)[0].toLatin1() != '-') {
```
GPSBabelDeveloper: Can we simplify these blocks? Now that we can rely on Unicode literals, can all these actually become qargs[argn][0]== QStringLiteral("-")
(deferring array vs. at notation to style)

main.cc
```diff
@@ -325,77 +298,76 @@ main(int argc, char* argv[])
...
-    //  optarg = argv[argn][2] ? argv[argn]+2 : argv[++argn];
-    //  cet_convert_init(optarg, 1);
+    //  optarg = qargs.at(argn).size() > 2 ? QString(qargs.at(argn)).remove(0,2) : qargs.size()>(++argn) ? qargs.at(argn) : QString();
+    //  cet_convert_init(CSTR(optarg), 1);
     //  break;
     case 'i':
-      optarg = argv[argn][2]
-               ? argv[argn]+2 : argv[++argn];
-      ivecs = find_vec(optarg, &ivec_opts);
+      optarg = qargs.at(argn).size() > 2 ? QString(qargs.at(argn)).remove(0,2) : qargs.size()>(++argn) ? qargs.at(argn) : QString();
```
GPSBabelDeveloper, line 346: Can we move all these .size() > 2 chunks higher up so we don't repeat them over and over? These blocks are just so heavy.

main.cc
```diff
@@ -325,77 +298,76 @@ main(int argc, char* argv[])
...
+      ivecs = find_vec(CSTR(optarg), &ivec_opts);
       if (ivecs == NULL) {
-        fatal("Input type '%s' not recognized\n", optarg);
+        fatal("Input type '%s' not recognized\n", qPrintable(optarg));
```
GPSBabelDeveloper, line 349: If you get tired of that (no change needed) you can
Fatal() "Input tpe '" << optarg << "'is not recognized";
It's admittedly more difficult to fine tune the stringstream style operators.

main.cc
```diff
@@ -527,7 +498,7 @@ main(int argc, char* argv[])
 
       break;
     case 'S':
-      switch (argv[argn][2]) {
+      switch (qargs.at(argn).size() > 2 ? qargs.at(argn)[2].toLatin1() : '\0') {
```
GPSBabelDeveloper: It's 2008. :-) Please just remove this entire #if 0 block and the case associated with it.

GPSBabelDeveloper (commented):
Lovely job tackling some admittedly hairy code. I think you made wise inflection points of C Strings and QStrings between subsystems/modules.

main.cc
tsteven4: 1. "The QString::at() function can be faster than operator[](), because it never causes a deep copy to occur."  So it is probably better to use at for read only access as we are doing here, and more than just a choice of style.
2. qargs.at(argn).at(0) is a QChar, so we could do qargs.at(argn).at(0) != QChar('-'), but I think that is less efficient as the QChar construction will involve a QTextCodec.  QChar::toLatin1 in an inline define to return an existing private field.

tsteven4: In 1, should have read "... faster than the \[\] operator ..."

main.cc
tsteven4: The issue is the conditional execution of ++argn.  We don't know if we want to do that until we know what option we are processing, and if it is run together like "-fFilename" or in two pieces like "-f Filename".  Also note that only one of these optarg assignments ever gets executed for a given option.
I could define a macro so we don't risk having a typo in one of them.

main.cc
tsteven4: 1. I just literally translated the existing code.
2. The first one is special just to avoid a leading blank.
3. The format of the batch file apparently can have the arguments spread willy nilly over 1 or more linnes.  This code just packs them all into one line so we can parse it.
4. I believe xcsvparse handles quoting, as would be neccesary for a file name containing a blank.  So it is a bit more than split().

main.cc
```diff
@@ -219,46 +198,40 @@ print_extended_info(void)
...
+#endif
   int c;
   int argn;
   ff_vecs_t* ivecs = NULL;
   ff_vecs_t* ovecs = NULL;
   filter_vecs_t* fvecs = NULL;
-  char* fname = NULL;
-  char* ofname = NULL;
+  QString fname = NULL;
+  QString ofname = NULL;
```
viettaml: These look like awkward constructor semantics for QString. According to the documentation, a default QString object is null, so just "QString ofname;" should suffice.

tsteven4:
I took care of all the suggestions except the three I made comments on: batch file parsing, optarg heaviness, and the QStringLiteral idea (need to show outdated to see this one).  I await any further comments or discussion on the unaddressed suggestions.

PR #115, opened 2017-11-16
Extend KML, so that <TimeSpan> can be used to describe the time span

bubeck:
Updated anything according to the feedback and rebased, so that ChangeLog is not including all the "noise" of the previous pull request. Sorry for that.


tsteven4 (approved):
thanks

PR #116, opened 2017-11-16
build using msvc x64 on AppVeyor.

tsteven4:
Changing the platform variable win32-mcsv* to use the default TEMPLATE (app) creates a Makefile we can run nmake on.  This is an alternative to a TEMPLATE of vcapp which creates a Visual Studio project file to build the app.


GPSBabelDeveloper (approved):
The "if not platform else" line scales badly, but the payoff here is big enough let's not nitpick. Go for it.

PR #120, opened 2017-11-19
ensure track filter merge operation is consistent.

tsteven4:
previously qsort might or might not rearrange waypoints with equal creation times depending on the qsort implementation.  This could cause the results of a merge operation to vary across different operating systems.

### DIFF
--- a/trackfilter.cc
+++ b/trackfilter.cc
@@ -276,7 +276,15 @@ trackfilter_merge_qsort_cb(const void* a, const void* b)
     return 1;
   }
   if (dta == dtb) {
-    return 0;
+    int seqno_a = gb_ptr2int(wa->extra_data);
+    int seqno_b = gb_ptr2int(wb->extra_data);
+    if (seqno_a > seqno_b) {
+      return 1;
+    } else if (seqno_a == seqno_b) {
+      return 0;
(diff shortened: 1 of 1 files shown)


GPSBabelDeveloper (commented):
Replacing qsort with std::sort or QAlgorithm::sort would be a good starter project for a budding Babel-head.  That may be easier after replacing my too-clever linked lists everywhere with sane type-safe containers.

tsteven4:
QtAlgorithms::qsort is not stable, and is obsolete.
QtAlgorithms::qStableSort is stable, and is obsolete.
std::sort is not stable, this is recommended as a replacement for 
QtAlgorithms::qsort
std::stable_sort is stable, this is recommended as a replacement for 
QtAlgorithms::qStableSort

stable => "A *sorting* algorithm is said to be *stable* if two objects 
with equal keys appear in the same order in *sorted* output as they 
appear in the input array to be *sorted*."

What we wanted in this PR was a consistent sort order, a stable sort 
will deliver this.  We want to produce the same output independent of 
the implementation of the underlying sort, e.g. across platforms.

We likely have the same issue in trackfilter with the other qsort, but 
our test cases don't exercise this.

PR #123, opened 2017-11-23
cleanup support for shape files.

tsteven4:
memory leaks, memory corruption, and other bugs fixed.
on write the required .dbf file is now created.
test added.
document updated.


GPSBabelDeveloper (approved):
Nice. That's API from Shapelib is so weird...

PR #124, opened 2017-11-26
Appveyor

tsteven4:
1. testo is modified to accept a compare between either of two files and a third.  This is used in the mtk test to accommodate different rounding modes used in vsnprintf, which is used by gbfprintf, on different systems.  This allows the test to pass on windows builds.  Many more tests will fail with Qt >=5.7 as QString changed rounding modes.
2. the obsolete translation "TRANSLATIONS += gpsbabel.ts" is removed from the GUI.
3. the appveyor build is modified to run testo.
4. the appveyor build is modified to build the GUI and run Inno setup.  The install executable is saved as an artifact.  This is only done for "Environment: name=win32, platform=mingw, qt=5.5\mingw492_32, tools=mingw492_32".  Note that Qt 5.5 is the last release that includes webkit.  In Qt 5.6 - Qt 5.9.2 webengine is not usable with the mingw tool chain.

I will wait for consensus before I act on this PR.  It does make changes that could be useful in making a new release, but I don't want to interfere in that process if these changes aren't agreeable.  If necessary I can split this PR up.


appveyor.yml
```diff
@@ -5,19 +5,94 @@ environment:
...
   - if not %platform%==mingw call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" %platform%
 
 # Flesh out with amd64/msvc cases later.
 build_script:
-  - qmake GPSBabel.pro
-  - if not %platform%==mingw (nmake) else (mingw32-make)
+  - ps: |
+      if (($env:platform -eq "mingw") -and ($env:qt -eq "5.5\mingw492_32"))
+      {
+        Get-Location
```
GPSBabelDeveloper, line 40: Should the bulk of this perhaps be in a tools/something.bat file with the release Qt directory passed in so we can avoid tying that into Appveyor?  ("No" is an OK answer.)

appveyor.yml
```diff
@@ -5,19 +5,94 @@ environment:
...
+        cd build-app-Desktop_Qt_5_5_1_MinGW_32bit-Release
+        qmake ..\gui\app.pro -spec win32-g++
+        mingw32-make qmake_all
+        mingw32-make
+        if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
+        lupdate ..\gui\app.pro
+        lrelease ..\gui\app.pro
+        # windeployqt in 5.5.1 suffers from bug
+        # https://bugreports.qt.io/browser/QTBUG-48946
+        # which trashes the translations.  Work around
```
GPSBabelDeveloper: How gross that was to find! We've been immune from this in release so far becayse I run lupdate and lrelease (and everything else I can...) on my release system (a Mac) which has VMs in it running a shared source tree.

gui/setup.iss
```diff
@@ -39,27 +39,27 @@ Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{
 [Files]
 ; This isn't as wreckless as it seems; these directories are populated on a
 ; controlled way by the batch file.
-Source: qtdir\bin\*.dll;		DestDir: "{app}"; Flags: ignoreversion
-Source: qtdir\plugins\*; 		DestDir: "{app}\plugins"; Flags: ignoreversion recursesubdirs createallsubdirs
+;Source: qtdir\bin\*.dll;		DestDir: "{app}"; Flags: ignoreversion
```
GPSBabelDeveloper: Is the whole block of using qtdir dead and replaced by the copying below? If so, should we nuke this and teh related comments?

gui/setup.iss
```diff
@@ -39,27 +39,27 @@ Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{
...
-Source: qtdir\plugins\*; 		DestDir: "{app}\plugins"; Flags: ignoreversion recursesubdirs createallsubdirs
+;Source: qtdir\bin\*.dll;		DestDir: "{app}"; Flags: ignoreversion
+;Source: qtdir\plugins\*; 		DestDir: "{app}\plugins"; Flags: ignoreversion recursesubdirs createallsubdirs
 
 Source: gmapbase.html; 			DestDir: "{app}"; Flags: ignoreversion
 Source: qt.conf;       			DestDir: "{app}"; Flags: ignoreversion
 
-Source: ..\build-app-Desktop_Qt_5_2_1_Mingw_32bit-Release\release\gpsbabelfe.exe; 	DestDir: "{app}"; Flags: ignoreversion
-Source: ..\..\build-GPSBabel-Desktop_Qt_5_2_1_Mingw_32bit-Release\release\gpsbabel.exe;   	DestDir: "{app}"; Flags: ignoreversion
+Source: ..\build-app-Desktop_Qt_5_5_1_Mingw_32bit-Release\release\*; Excludes: "*.cpp,*.o"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsubdirs
```
GPSBabelDeveloper, line 42: Doesn't this make it more easy to accidentally ship files we didn't mean to ship? Granted, in an automated environment, that's less of a big deal than on our workstations where we may have a stray script file or editor backup.

GPSBabelDeveloper (commented):
I like the idea of building Windows in automation a lot, so I really appreciate you tackling that.

I do wonder if we're leaving the rails with multiple reference files and thus having more than one "true" conversion. Can we perhaps be more critical with floor(), ceil(), round(), and trunc() and the edges of our output and get the resolutions and rounding modes consistent?

appveyor.yml
tsteven4: maybe.  I think the most important reason to do that would be to allow the script to run outside of appveyor.  One can also argue against having lengthy scripts in CI configuration files.  Moving it has the potential to affect appveyor error handling.  I'm not there yet.

I did modify appveyor.yml to set ps variables for the shadow directories, so at least they only have to be changed in one place.

Note that we don't have to use the default Qt creator shadow directory names.  We could use Qt version and compiler independent names in appveyor.yml and setup.iss.in.

gui/setup.iss
tsteven4: agreed.  will nuke.  additionally mksetup.bat is obsoleted by this PR and should be deleted from the HEAD.

gui/setup.iss
tsteven4: I tried a few changes to hopefully mollify your concerns:
1) I explicitly delete the build directories we will be getting stuff out of before filling them with the make. This isn't necessary in the appveyor environment, at least without build caches, but I think it makes the intent more explicit.
2) I have inno setup create a manifest and make it an artifact.  It can be examined to see what got included in the setup executable.

tsteven4:
I share your concerns about alternate references.  Although a one line kludge to mtk to "fix" our current mismatch with csv elevation data is possible I don't want to go there.  There will be many more test failures with Qt >= 5.7.  At least with the way I proposed it is easy to scan for the new testo compare routine to find the rough spots.

To fix this in general, beyond our test cases and pre-Qt and Qt code is not trivial.  The industry has failed to come up with a fix for a few decades for printf et.al.   C11 standard 7.21.6.1 still says 

> The value is rounded to
the appropriate number of digits.

Anyway, I would like to move forward with the dual reference system so we can get windows test online.  We need to get some encoding tests on that platform ASAP.

without defining how to do the rounding.

robertlipe:
I'd like the ability to spin a release outside of CI. The Qt4 references
all over makesetup.bat make me suspect it's not been used in a while but
offhand, I'm not remembering what was.

I'm concerned the dual-rounded reference thing will lead to ongoing
maintenane hassles as we'll then have two official conversions, but for now
it's done. Go for it.

tsteven4:
I think the Qt4 references in makesetup.bat are all in comments.  It looks to me like it was used for the 1.5.4 release.  
I think all of makesetup.bat is replaced by running windeployqt.  Using windeployqt saves us from chasing down all the new Qt files we need to distribute as well as patching the dlls so they don't contain paths from the build system (at the cost of finding the windeployqt bugs). Most of the stuff in appveyor.yml build_script is doing the build itself.  The appveyor build_script, minus the Push-AppveyorArtifact lines and the control flow code could be pasted into a .ps1 script and run outside of CI in a "Qt 5.5 for Desktop (MinGW 4.9.2 32 bit)" cmd window with something like "powershell.exe -noexit -file xxx.ps1".  It isn't necessary to match the shadow directory names unless you want to do the builds in Qt Creator and use the default shadow directory names, but they could be script parameters.

appveyor.yml
GPSBabelDeveloper: That's just historical. The batch files have tended to be "automate what I just did" so the last few releases have been "run Qt Creator and accept the defaults and then feed that to innosetup and then give more steroids to the pre-Qt batch file" It was thus approximately the "if mingw" case below.

gui/setup.iss
```diff
@@ -25,6 +23,7 @@ DefaultDirName={pf}\GPSBabel
 DefaultGroupName=GPSBabel
 OutputDir=release
 OutputBaseFilename=GPSBabel-1.5.4-Setup
+OutputManifestFile=GPSBabel-1.5.4-Manifest.txt
 SetupIconFile=images\babel2.ico
```
GPSBabelDeveloper, line 27: Wow, there's an icon file from the (pre innosetup?)  past. Must find and destroy... and add "qtinstaller" to my list of things to learn more about.

gui/setup.iss
GPSBabelDeveloper: I like it.

tools/make_windows_release.ps1
```diff
@@ -0,0 +1,51 @@
...
+cd ..\gpsbabel
+# make sure we are staring with a clean build directory
+Remove-Item "$($gui_build_dir)" -Recurse -ErrorAction Ignore
+New-Item "$($gui_build_dir)" -type directory -force
+cd "$($gui_build_dir)"
+qmake ..\gui\app.pro -spec win32-g++
+mingw32-make qmake_all
+mingw32-make
+if ($LastExitCode -ne 0) { $host.SetShouldExit($LastExitCode)  }
+lupdate ..\gui\app.pro
```
GPSBabelDeveloper, line 41: I don't think it would make me sad if we didn't run lupdate/lrelease on Windows because the outputs are checked in.  We have so little activity in the GUI and the line number fuzzer is pretty good, so if this makes windeployqt easier, take advantage of that.

appveyor.yml
```diff
@@ -5,19 +5,68 @@ environment:
...
+      Write-Host "Finished build_script"
+
+test_script:
+  - ps: |
+      $env:projdir=(C:\msys64\usr\bin\bash.exe -c "pwd")
+      C:\msys64\usr\bin\bash.exe -lc "echo $env:projdir"
+      # work around issue with standard error output from xmllint causing a build failure by redirecting stderr.
+      # in this case making sure the last command succeeds is not sufficient.
+      # see https://help.appveyor.com/discussions/problems/10014-false-build-fail-status
+      C:\msys64\usr\bin\bash.exe -lc "cd $env:projdir; PNAME=./release/GPSBabel.exe GBTEMP=./gbtemp ./testo 2>&1"
```
GPSBabelDeveloper, line 70: I'm impressed that works as well as it does. :-)

GPSBabelDeveloper (approved):
Very nice. Lots of icky problems chased down here. Thank you!

The idea that the official windows build could hatch from an environment that I don't maintain is pretty sweet.

PR #126, opened 2017-11-27
Header sorting, moving to C++11. Iterator simplifications.

robertlipe:


tsteven4:
It looks like there was merge trouble with main.cc.

PR #129, opened 2017-11-27
Fix KML writer crash bug

viettaml:
When Google Earth uses gpsbabel to import KML in "live" USB mode, kml_wr_position_init() is called before kml_wr_init() resulting in deferencing an uninitialized "writer" pointer. The solution is just to remove the setAutoFormatting(false) call; there's no real need for it as it doesn't measurably impact performance (the extra cost to ingest/parse additional whitespace is trivial). Also remove other calls to remove nice formatting on timestamps (addressing a FIXME comment), and update the golden test data to match.

### DIFF
--- a/kml.cc
+++ b/kml.cc
@@ -91,8 +91,6 @@ static bounds kml_bounds;
 static gpsbabel::DateTime kml_time_min;
 static gpsbabel::DateTime kml_time_max;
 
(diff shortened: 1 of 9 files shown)


GPSBabelDeveloper (approved):
Thanx. That code's had a hard life (beginning in C, "serializing" via printf, little used realtime mode) so it's time to pull those compat props.

PR #130, opened 2017-11-28
add a test of utf8 filename handling.

tsteven4:
4/6 tests fail on windows without PR #87 
initially this test is only deployed on unix on travis, after that PR is merged we can add it on appveyor.


tsteven4:
This PR now also enhances the latin1 test to cover more file apis.

PR #132, opened 2017-12-01
Support Unicode environment variables.

viettaml:
Create ugetenv() OS-abstraction to return a QString environment variable value. Use it in place of getenv() in most places. Update inifile.cc routines to use QString to support Unicode paths for the gpsbabel.ini file, also greatly streamlining the code.


csv_util.cc
```diff
@@ -1215,11 +1215,11 @@ xcsv_parse_val(const char* s, Waypoint* wpt, const field_map_t* fmp,
     wpt->SetCreationTime(sscanftime(s, fmp->printfc, 1));
     break;
   case XT_LOCAL_TIME:
-    if (getenv("GPSBABEL_FREEZE_TIME")) {
+    if (ugetenv("GPSBABEL_FREEZE_TIME").isNull()) {
```
viettaml, line 1218: I flipped the if/else clauses rather than doing a double negative "!...isNull()".

util.cc
```diff
@@ -258,6 +258,19 @@ xfopen(const char* fname, const char* type, const char* errtxt)
   return f;
 }
 
+/*
+ * OS-abstracting wrapper for getting Unicode environment variables.
+ */
+QString ugetenv(const char* env_var) {
+#ifdef __WIN32__
+  // Use QString to convert 8-bit env_var argument to wchar_t* for _wgetenv().
```
GPSBabelDeveloper, line 266: qtenv does us no favors here. Ugh.
http://doc.qt.io/qt-5/qtglobal.html#qgetenv

inifile.cc
```diff
@@ -44,98 +46,55 @@ typedef struct inifile_section_s {
...
-  test = fopen(buff, "rb");
-  if (test) {
-    fclose(test);
-    return buff;
+  if (path.isNull()) {
+    return QString();
   }
-  xfree(buff);
-  return NULL;
+  QString inipath(QDir(path).filePath(GPSBABEL_INIFILE));
```
GPSBabelDeveloper, line 60: A lovely example of how Qt saves us from OS-specific icky stuff.

inifile.cc
```diff
@@ -44,98 +46,55 @@ typedef struct inifile_section_s {
...
   gbfile* res = NULL;
 
-  envstr = getenv("GPSBABELINI");
-  if (envstr != NULL) {
-    FILE* test;
-
-    test = fopen(envstr, "r");
-    if (test != NULL) {
-      fclose(test);
+  envstr = ugetenv("GPSBABELINI");
```
GPSBabelDeveloper, line 71: Sinze the FREEZE_TIME stuff is only used by our test framework to set "current time" to 1-1-1970, and not really an externally documented/useful behaviour, this is really the only place where getenv returning funky characters is "real", right?

util.cc
```diff
@@ -258,6 +258,19 @@ xfopen(const char* fname, const char* type, const char* errtxt)
   return f;
 }
 
+/*
+ * OS-abstracting wrapper for getting Unicode environment variables.
+ */
+QString ugetenv(const char* env_var) {
```
GPSBabelDeveloper, line 264: I'd have named this "xgetenv" for consistency with our other C-era wrappers, but rather than asking to change it, I'll add it to the rainy day list of things to move to gb::something or src/core/something.cc or other better fit.

GPSBabelDeveloper (approved):
Minor reviewing out loud, but a nice cleanup for an esoteric problem. Thnxa nd merged.

Sidebar: I wonder if the gpsbabel.ini stuff is ever used in reality...

inifile.cc
viettaml: The most plausible example of how you can get environment variables with weird characters is if the Windows username is non-ASCII, which seems a common enough scenario in non-English locales. Presumably the Windows user directory is then "C:\Users\%USERNAME%" (unless Windows instituted some workaround that hashes unicode usernames into ASCII, which wouldn't surprise me at all, 8.3-hashed filenames anybody?), and any "INI file path" based off of that, would also have non-ASCII characters.

util.cc
viettaml: I considered this, but felt that "xfopen" was a bit more than a thin wrapper around "fopen", whereas "ufopen" (from PR #87) is a Unicode-supporting thin wrapper, hence the "ugetenv".

util.cc
viettaml: Yup. Supposedly in Qt 5.10 there's finally a Unicode-supporting qgetenv or somesuch, but not until then.

robertlipe:
I'm sure our views on Qt 5.10 are similar. I need to refresh my mental
status of QtWebEngine on Linux, but last time I checked we had some demand
to stay on 5.2 with a forked WebKit/WebEngine. A floor of 5.6 or 5.9 (the
LTS versions) isn't unreasonable but 5.10 isn't in our near future.

I reran the mysteriously failing presubmit and it's all tidied up now.Thanx
for all the fixes.  Nap time!

PR #135, opened 2017-12-04
Code cleanups in gopal.

robertlipe:


tsteven4:
For an argument of QString::compare it may be preferrable to use QLatin1String instead of QStringLiteral.
https://woboq.com/blog/qstringliteral.html

PR #137, opened 2017-12-04
Fix redundant file rename on Windows

viettaml:
Just use cross-platform QFile calls.


tsteven4:
I want to think about this more.  Pre PR we were atomic on windows, but on unix we would fail once the output file existed.  I have never tested real time tracking.

There is also the renaming concern in mtk_logger, not sure what to do about that either.

tormet:
Just a remark if we really wants to have an atomic operation for the move operation. Then we can use rename under Linux/Unix systems.  This should also work under Mac OS X.

BTW I think the unified solution with QFile:: is fine in this context. I wouldn't care about race conditions. 

Nevertheless I would check the return status of Qfile::rename given a warning if it fails.  Or do I overlook something, because I am not familiar with realtime positioning?

robertlipe (approved):
Sorry for the delay on this. Thanx for the fix, Tam.

Agreed that if there's some transient problem, it'll be self-correcting. On the next fix update (typ. ~2 seconds) we'll get a whole new file written to the tmp one and will attempt the rename here. The one consumer of this code that I know of already has (had?) retry logic.

I'll fold it into master since you've accepted it upstream.

PR #139, opened 2017-12-05
Nuke remainder of NEW_STRINGSs.

robertlipe:


tsteven4:
This needs the main revert, then it has a chance at CI.

tsteven4:
I don't think .qmake.stash should be included.

tsteven4:
the appveyor fails are a floating point mismatch issue on mingw 32, i.e. noise.  But the .qmake.stash needs to go away, probably into .git.ignore.

GPSBabelDeveloper:
These changes have long ago been integrated.

PR #142, opened 2017-12-06
Make track-filter use Qt times in more places. Minor code cleanups.

robertlipe:


GPSBabelDeveloper:
Everything here has been absorbed in other CLs and in some places, had these improvements replaced by yet more improvements.

PR #143, opened 2017-12-06
Fix a bad merge base in PR 126. Only main.cc seems to be affected.

robertlipe:


tsteven4:
It looks like this main.cc is equivalent to e4bced74359bffbea7da9c94ceb2858488f1fac4 main.cc which is good by me.  Are you going to add your cleanup tricks to main.cc?

GPSBabelDeveloper:
I've picked through the changes on my branch and there's was very little to
cherry pick.

https://github.com/gpsbabel/gpsbabel/commit/a1751e03fa71616b3aaeba42314baa2f41580f31#diff-d759577e53e2ce061239002d8a592cc5
seems to be where I left the rails.

There's a reason it's recommended I not operate heavy machinery...

PR #144, opened 2017-12-07
Loosen coupling of upgrade queries to OS version by using newer (2014) Qt services.

robertlipe:


robertlipe (commented):
pointless to try to trigger an automerge

robertlipe (commented):
silly comment here

PR #145, opened 2017-12-07
Webengine

tsteven4:
1. get webengine working
2. get js logging working with webengine (mclicker.logTimeX)
3. update to google maps javascript api v3.
4. add conditional code to capture generated js for debugging and jslint.

This seems to be mostly working with webkit and webengine.
**I have seen intermittent segmentation faults with webkit.**, so I don't think this is ready yet.


tsteven4:
gmapbase.html has 3 errors flagged by http://validator.w3.org/check.  I don't think any of these are new. I am not sure what -khtml-user-select is all about, I deleted some related bits in map.cc.

> Error Line 3, Column 50: there is no attribute "style"

<html xmlns="http://www.w3.org/1999/xhtml" style="height:100%">

> Line 167, Column 66: there is no attribute "topmargin"

…dy style="height:100%; -khtml-user-select:none;"  topmargin="0" leftmargin="0">

> Line 167, Column 81: there is no attribute "leftmargin"

…dy style="height:100%; -khtml-user-select:none;"  topmargin="0" leftmargin="0">

gui/map.cc
```diff
@@ -69,28 +76,57 @@ Map::Map(QWidget *parent,
...
 }
 
 //------------------------------------------------------------------------
 Map::~Map()
 {
   if (busyCursor_)
     QApplication::restoreOverrideCursor();
+#ifdef DEBUG_JS_GENERATION
+  if (dbgout_) {
+    delete dbgout_;
```
GPSBabelDeveloper, line 117: OPtional: If we embed those in the Map object instead of making them raw pointers in the Map Object, they'll self-destruct. Even cleaner than smart pointers.

gui/map.cc
```diff
@@ -69,28 +76,57 @@ Map::Map(QWidget *parent,
   manager_ = new QNetworkAccessManager(this);
   connect(this,SIGNAL(loadFinished(bool)),
 	  this,SLOT(loadFinishedX(bool)));
-  this->logTimeX("Start map constuctor");
+  this->logTime("Start map constuctor");
```
GPSBabelDeveloper, line 79: Is /this/ ever needed in this file?

GPSBabelDeveloper (commented):
Spiffy modernization. Thanx! (Submit when you're ready)

robertlipe:
There's little question that this 2009-era code predates trendy HTML5
goodness, so I'm not surprised that a validator fusses.

gui/map.cc
tsteven4: Are you asking about using "this->" syntax, or are you asking about the logTime(X) functions?  The logTime functions track various timings and report to the gui console if you enable debug in the options dialog.  I think this used to be more important with all the encoded polylines, but they are gone with v3 except for driving directions.

robertlipe:
The former. The explicit form is rarely used elsewhere in our code; we let
it remain hidden.  Dropping "this->" probably works the same.

I think. I'm typing from a fone.

PR #146, opened 2017-12-09
modify igc writer for consistent rounding.

tsteven4:
uggh.  writing code to use round(), which has defined behavior for halfway cases, unlike *printf.  But it gets our CI passing again.


GPSBabelDeveloper (approved):
I have a feeling we have a lot of these in our future. It's something we need to look for in new formats and incoming patches, too.

Thanx.

tsteven4:
This original commit in this PR was problematic.  It could end up with n degrees and 60000 milliminutes.

The original test data that failed before the PR is 48.2709750, or 2,896,258.5 milliminutes.  In this case any accumulated fp math errors can cause it to go either way.

PR #151, opened 2017-12-16
generalize make_windows_release for MinGW or MSVC.

tsteven4:
~~Note that the created setup contains the MSVC redistributable installer, but setup does not execute it.  Thus, on a system without the VC redistributables the installed GPSBabelFE will fail.~~

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,10 @@ environment:
   - name: win64
     platform: amd64
     qt: 5.6\msvc2015_64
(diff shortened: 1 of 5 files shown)


tsteven4:
This now installs the VC redistributable if needed.

tsteven4:
The installer is running in 32 bit mode regardless of the installed files.  Thus if we generate 64 bit exectuables they end up in the wrong directory, i.e. "Program Files (x86)".

PR #153, opened 2017-12-18
make use of QStringLiteral with QXmlStreamWriter.

tsteven4:


robertlipe (commented):
I'm OK with merging this, but man does this get bulky.

Stylistically, in the future we should probably move these to local class constants just to trust that the QStringLiteral constructors don't do anything stupid at runtime (I get that's its purpose) since so many of these appear in loopy places.

tsteven4:
I agree it is unfortunately verbose.  Furthermore, I cooled on this when I couldn't measure a performance improvement in my VM.  I can go either way on this PR.

PR #160, opened 2018-03-01
Increase NMEA lat/lon resolution to 7 significant digits to support c…

sshade:
…m level position accuracy

The Lat/Long in the GGA and RMC output messags is currently truncated to 3 digits of minutes.  This translates to roughly 2m of resolution.  High-precision GNSS receivers are capable of cm level precision in Lat/Lon so gpsbabel should support higher resolution for Lat/Long in the NMEA messages.

### DIFF
--- a/nmea.cc
+++ b/nmea.cc
@@ -1313,7 +1313,7 @@ nmea_trackpt_pr(const Waypoint* wpt)
   }
 
   if (opt_gprmc) {
-    snprintf(obuf, sizeof(obuf), "GPRMC,%010.3f,%c,%08.3f,%c,%09.3f,%c,%.2f,%.2f,%06d,,",
+    snprintf(obuf, sizeof(obuf), "GPRMC,%010.3f,%c,%12.7f,%c,%13.7f,%c,%.2f,%.2f,%06d,,",
              (double) hms + (wpt->GetCreationTime().time().msec() / 1000.0),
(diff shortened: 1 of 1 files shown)


sshade:
I need to fix some of the unittests.  I will fix those an submit a new pull request.

goamberg:
Did this get included in gpsbabel somewhen? 
It looks like there are still only 3 digits in minutes...

GPSBabelDeveloper:
The change was withdrawn in 2018 and the issue closed. Since our target audience is consumer grade GPS and this is beyond their precision, we'd decline this change anyway.
If you want a custom one-off, you're free to integrate something like the above into your local copy.

sshade:
Ok.  Sure.   I already tested locally so I'm ok for my needs.

With the Android GNSS measurement interface anyone can use the measurements
from their phones to do RTK/PPP.  ION papers have shown cm level precision
using consumer grade chips.  Typical open sky accuracy for consumer grade
chips is around 2 meters so the 3 degrees is very near their current
capabilities when no augmentation is used.  If any augmentation is used
(even sbas) accuracy will improve to around 1m.  PPP for consumer devices
is right around the corner and sub-meter will be routine for consumer grade
devices that support PPP.

Anyways, you have the change ready whenever you decide it is warranted.
Thanks for the reply.

Regards,
Steven

PR #162, opened 2018-03-18
Qt5 is required

cweiske:
Qt4 cannot be used anymore - see https://sourceforge.net/p/gpsbabel/mailman/message/36264877/


GPSBabelDeveloper (approved):
Thanx.  Does Ubuntu no longer split the development (headers, debug libs) from the runtime (.so) packages? 

I sat out to repair/rewrite this entire document (it's a victim of the great Qt Battle of 2014) a few months ago, but only made it worse so I'll take "obvious" fixes like this until I can write once again.

GPSBabelDeveloper (commented):
Please note that it will take a while for this to propagate to the web. It'll perc from -development and then to -nextversion.

cweiske:
> Does Ubuntu no longer split the development (headers, debug libs) from the runtime (.so) packages?

I always though that `-dev` packages do only contain the header files. The package description says so:

>  This package contains the header development files used for building Qt 5 applications.

But `dpkg -L qtbase5-dev` also shows some `.so` files - I have no idea why.

robertlipe:
Just from the naming, I'd expect -dev to depend on -core and I'd expect the
.so's to be in -core.

Any Linux devs out there know if we should be recommending both or if there
are enough dependencies for Christian's suggestion to be sufficient in
current distributions?

Issue #90, opened 2017-10-02
Patch: garmin_fit: warn instead of fail on unsupported protocol version


frief:
Files conforming to newer protocol versions should be (at least partly) decodable with programs that support older versions. 
Successfully converted Garmin Edge 520 FW 12.20 fit files protocol version 2.0 to gpx with lat, lon, ele, time, speed, wpt.

Thanks for gpsbabel!

[0001-garmin_fit-warn-instead-of-fail-on-unsupported-proto.patch.txt](https://github.com/gpsbabel/gpsbabel/files/1349616/0001-garmin_fit-warn-instead-of-fail-on-unsupported-proto.patch.txt)

robertlipe:
This seems very dangerous as fit has changed in incompatible ways before. I
think I'd rather raise the level than turn off the error.

To do this, please modify your patch to include a representative (trimmed
is preferable) .fit file for reference/track/ and an entry in
test.d/fit.test so we can test this going forward.

Thanx,
RJL

frief:
Thanks, please find the updated patch with the requested changes here
https://gist.github.com/frief/fc7fffb0b855d94df5ce645a5db4374b
The .fit file includes laps and was recorded with heartrate, power (with cadence), speed and Shimano Di2 gear shifting data. (Just drop a note if not ok)

xmedeko:
I vote for this bug since more and more devices generates FIT 2.0.

tormet:
Hi,
I have some code ready to support FIT 2.0. At least in the way that it read all the data we already get for FIT version 1 files. For the tests I still have to record a track again. The one I have, started at my front door. I want to avoid this. Then I have to do a last code review by myself and then I can do the pull request. This week I am very short of time.  But I hope I get the things done before the weekend.

In some way I understand your reasoning with the warning. If the old code can handle the data (because it does not use "special things" from version 2.0) you get a result. If it cannot handle the data while converting it,  it is the same as it give you the error that it does not support version 2.0. 
But if it gives you a warning you also have the possibility that it will write wrong data. Who knows this exactly, even if I do not expect it in practice. And having code that write wrong data is something one wants to avoid.

For me e.g. the old code throws an error while converting data because the Wahoo Element Bolt uses so called developer fields which are not handled in the current implementation.

I hope this helps a little bit.

xmedeko:
@tormet It's great you have a patch and thanks for the analysis that developer fields may cause a crash.

[...]

PR #163, opened 2018-03-20
Added support for FIT version 2.0

tormet:
With FIT version 2.0 developer fields where added to the definition
messages. These fields extends the definiton messages and due to this the
number of records which must be parsed afterwards.

FIT files containing developer fields could not be parsed before
this changes.

Simple support was added by just parsing the data but without using them.
It is not expected that any of the currently used data like the
position or the elevation is coded inside the developer fields
because for this we have the already existing records from
FIT version 1.0

Added a test with a fit test file coming from the Wahoo Element Bolt
which uses developer fields from FIT version 2.0


xmedeko:
Fixes #90 

@tormet a wicked FIT 2.0 for a test may be found at https://drive.google.com/open?id=1CruSthjf33NjEr0SfRzbOCg1sQg-yvqX 
(It's from https://github.com/dtcooper/python-fitparse/issues/62).

tormet:
@xmedeko thanks, it is no problem to parse and convert this FIT 2.0 test file.

Remark: I did not change how FIT 1.0 data is handled. E.g. this FIT 2.0 test file has e.g. 5 lap messages which are already handled but also one session message with one starting point. Session messages are currently not handled in the FIT->GPX converter.

xmedeko:
This RunScribe FIT file has more unusual things: some developer fields without name and some fields with multiple values (e.g. `enhanced_altitude`). 👍 that your patch can parse it.

@gpsbabel I vote for this path to the upcoming release, please.

tormet:
Hi @xmedeko,
it is very good to have additional test files. BTW I only do the test under Linux.

Normally I do not expect problems with (complicated) developer fields. This reason is described in the comments on my source code changes.

I do not use the developer fields, getting data from them. This means I am just parsing them (\*) and throw them away and the 2 special global data messages (Developer Data ID Message and Field Description message) are normal data messages as before (for parsing). So there is nothing special with them, the old code could just do this. 
(\*) This is new, the definition message was extended in FIT version 2.0.

garmin_fit.cc
```diff
@@ -265,20 +267,72 @@ fit_parse_definition_message(uint8_t header)
...
   // remainder of the definition message is data at one byte per field * 3 fields
-  def->fields = (fit_field_t*) xmalloc(def->num_fields * sizeof(fit_field_t));
-  for (i = 0; i < def->num_fields; i++) {
-    def->fields[i].id = fit_getuint8();
-    def->fields[i].size = fit_getuint8();
-    def->fields[i].type = fit_getuint8();
-    if (global_opts.debug_level >= 8) {
-      debug_print(8,"%s: record %d  ID: %d  SIZE: %d  TYPE: %d  fit_data.len=%d\n",
-                  MYNAME, i, def->fields[i].id, def->fields[i].size, def->fields[i].type,fit_data.len);
+  if (def->num_fields > 0) {
```
robertlipe, line 273: It's never < 0, is it? Without this, if it's ==0, we should get a zero byte xmalloc and an immediate break out of the loop and it'll all work, right?  Maybe an idea for future exploration.

Definitely opportunities here to make fit_field_t a class that allows id, size, type to be easily handled in the constructor and an operator new[].  (THis is true of most of our xmalloc's...)

garmin_fit.cc
```diff
@@ -265,20 +267,72 @@ fit_parse_definition_message(uint8_t header)
...
+  if (hasDevFields) {
+    int numOfDevFields = fit_getuint8();
+    if (global_opts.debug_level >= 8) {
+      debug_print(8,"%s: definition message contains %d developer records\n",MYNAME, numOfDevFields);
+    }
+    if (numOfDevFields == 0) {
+      return;
+    }
+
+    int numOfFields = def->num_fields+numOfDevFields;
```
robertlipe, line 318: So if we had 20 old style records and 30 new style records, we'll fields[20] above and then realloc here (which is a potentially large alloc + copy) to get fields[0-19] of old and [20-49] of new style, but set them to a record type that we'll never use anyway?

If that's true, why not skip the realloc and just loop and ignore the three fit_getuint8() calls via Q_UNUSED(fit_getuint8()) or (void) fit_get... - which tends to make the C-style cast haters angrier than the C preprocessing macro haters for the first.

garmin_fit.cc
```diff
@@ -265,20 +267,72 @@ fit_parse_definition_message(uint8_t header)
...
+      def->fields[i].type = fit_getuint8();
+      if (global_opts.debug_level >= 8) {
+        debug_print(8,"%s: record %d  ID: %d  SIZE: %d  TYPE: %d  fit_data.len=%d\n",
+                    MYNAME, i, def->fields[i].id, def->fields[i].size, def->fields[i].type,fit_data.len);
+      }
+
     }
   }
+
+  // If we have developer fields (since version 2.0) they must be read too
```
robertlipe, line 287: Stylistically, I'd have made this another function with if (!has_dev_fields) at the top and a fast return.

robertlipe (commented):
It may also be worth a glance at xmldoc/formats/garmin_fit.xml to see if we can list what models, versions, or devices use this new form.  

If you're not good with Docbook, just provide the words and one of us can help. Heck, if you're not good with words, help us list affected gear and we'll help with the words, too. :-)

robertlipe:
Torsten, I'll go ahead and approve this just to keep things rolling along since I dropped the inertia on this.  It'll make it easier for you to 'git pull' back to the head and consider the cosmetic changes. Really, the doc change is the only deal breaker if you think about the others and have a strong case (which you don't have to present fully: "I've thought about it and this is the most awesome it can possibly be"  Im OK with that. Allocating a spare gig (whatever) of memory isn't as big of a deal as it used to be.

Thanx!

garmin_fit.cc
tormet: No after the changes supporting FIT version 2 we cannot  leave the function immediately as before.  Even if def->num_fields == 0 we can have developer fields. So we have to handle this case.

tormet: I forgot I will take a look at xmldoc/formats/garmin_fit.xml

garmin_fit.cc
tormet: The reallocation does not make me nervous because normally the operating system should handle this efficiently.  

Your assumption that we do not use the content of the new fields is not correct. These fields contains the description of the newly developer fields which are stored inside the data message.

So yes, we just parse the developer fields (inside the data message) if they exist and throw them away. But for doing this we need the size of this fields and this size is stored in def->fields[i].size.

This is the same as for the already existing fields in the data message.  As mentioned  I use the existing infrastructure for this. 

And yes I think I was not clear enough with my comments. In this part of the code we just read the definition message with contains the information about the fields in the data message. In FIT version 2 we can have a longer definition message and also new fields in the data message. Sorry for the confusion.

garmin_fit.cc
tormet: I do not understand this comment. 
We are inside fit_parse_definition_message() parsing a definition message and this is part of the definition message. Perhaps my comments above made this more clear.

Remark: The definition message describes the next data message

PR #165, opened 2018-04-26
Shapefile doc and reader fixes.

tsteven4:
Don't cram different parts of polylines into one route, they
may not be connected.
Add documentation on transformation of coordinate systems that
may be required for GPSBabel usage of shapefiles.


robertlipe:
Looks good. Thanx.

PR #167, opened 2018-04-28
Move Qt floor to 5.7

tsteven4:
- modify travis build to use docker with an image that meets the minimum Qt floor.  The image is based on Ubuntu bionic (18.04 LTS) and has Qt 5.9.5.
- modify appveyor build to use MSVC 2015 and Qt versions 5.7.1and 5.9.4.  A 32 bit release candidate installer is built with MSVC 2015 and Qt 5.9.4.  MinGW builds are dropped as the GUI can't build with newer Qt versions due to issues with webkit and webengine on MinGW.
- update reference files for rounding changes in conversions from binary to characters that were introduced in Qt 5.7.
- update reference files for the more reasonable number of digits printed with newer Qt versions in the geojson tests.
- update xol.test to allow for observed variation in messages when a file is not found.
- update configure to enforce the Qt version floor.
- add Dockerfile used to build the image used by travis.


testo.d/xol.test
```diff
@@ -18,7 +18,8 @@ ${PNAME} -i xol -f ${REFERENCE}/doesnotexist -o gpx -F ${TMPDIR}/xol-sample_si.g
 # check error message is what we expected
 # Note that the reference file nonexistent.err includes the file path,
 # so this will fail when starting from a directory other than gpsbabel.
-compare ${REFERENCE}/nonexistent.err ${TMPDIR}/nonexistent.err
+# the message can vary depending on the c runtime, .i.e. msvc and gcc are slightly different.
+compare_with_alternate ${REFERENCE}/nonexistent.err ${REFERENCE}/nonexistent_alternate.err ${TMPDIR}/nonexistent.err
```
robertlipe, line 22: It's probably OK to just verify that we exited with something other than 0. Is checking $? -gt 0 enough?

configure.ac
```diff
@@ -166,12 +166,26 @@ AC_CHECK_TOOLS(QMAKE, [qmake-qt5 qmake], "no")
 AC_CHECK_TOOLS(LUPDATE, [lupdate-qt5 lupdate])
 AC_CHECK_TOOLS(LRELEASE, [lrelease-qt5 lrelease])
 
+MIN_QT_VERSION=5.7
```
robertlipe: Please make a note here and in main.cc and gui/main.cc to tie the QT_VERSION_CHECK there and MIN_QT_VERSION here together to keep them in sync.

[ Talking aloud. ] Why are both necessary? This lets the UNIXy "everything is ./configure && make' users fail fast and the other is needed for the Qt "load project, press build" users.

testo.d/xol.test
robertlipe: I don't see compare_with_alternate() in my tree. Does this depend on some other local change?

robertlipe (approved):
Solid work over a large number of systems that are only tangentially related to us (docker, appveyeor, autoconf, etc.) Thanx for tackling this!

testo.d/xol.test
tsteven4: It went into testo last year: https://github.com/gpsbabel/gpsbabel/commit/fbfcbe3d6ad59e2da5a677b435eb0c15bf77e285#diff-69e153e4d7add22f245e24de590eec21

configure.ac
tsteven4: A version check could go into GPSBabel.pro and gui/app.pro, but it would be nice to use versionAtLeast() and that may not exist in the version the developer is trying to use.

Also, we have/had other methods besides configure and .pro, e.g. xcode, and perhaps some old msvc files somewhere?

robertlipe:
Fully patched MacOS 10.13.4 with a totally stock Qt 5.9.5 successfully
builds the Gui and the CLI, which successfully completes testo, using both
the configure/make and qmake/make paths. No mysteries. CL forthcoming to
eliminate some silly warnings, but I want it to go through the presubmit
queue.

Issue #81, opened 2017-08-20
GPX to Naviguide returns an empty file


HarelM:
Running the following command line returns an empty file (0 bytes) - exit code 0:
```
gpsbabel.exe -N -i gpx,gpxver=1.1 -f "C:\Users\xxx\AppData\Local\Temp\tmp3C12.tmp" -o naviguide -F "C:\Users\xxx\AppData\Local\Temp\tmp3C22.tmp"
```
Here's the relevant input file.
[מסלול 1.txt](https://github.com/gpsbabel/gpsbabel/files/1237311/1.txt)

Running on Windows 10.
GPSBabel doesn't have a version in the file so it's dated from 02-jan-2016 (worth adding a version to the file...)

GPSBabelDeveloper:
Your inputs are tracks. Our naviguide format supports only waypoints
according to https://www.gpsbabel.org/capabilities.html

You could convert the tracks to waypoint, but I don't know how well that'll
work on the device or progra


1.5.4 was from New Year this year, but I really don't think we've had any
non-mechanical changes in that format in years. I don't know that we have a
lot of Naviguide users on the dev team as it seemed to be of regional
interest. Perhaps post to gpsbabel-misc may turn up more.

HarelM:
Thanks for the super fast response!
A user posted on our facebook group that he was unable to export in naviguide format and I now remember that we only imported this format up until now.
I'll try with Waypoints never-the-less.

tsteven4:
This will produce a non-empty file with 1078 points from your sample 
input.  the gpxver option you used is used on write, not read, so it is 
unnecessary.

./gpsbabel -N -i gpx -f  1.gpx -x transform,wpt=trk -o naviguide -F 2.twl

HarelM:
The syntax is confusing but I'm sure you know what you are talking about :-)
Thanks!! Not sure I'll use it as I prefer to manipulate the data before passing it to GPSBabel, but it sure can be useful.

robertlipe:
From the department of Stupid GPSBabel Tricks, one of the things I do a lot
is feed our own output back to us. (Please don't think too much about
naming that...)  It won't prove that our output can be read by whatever,
but it can prove that at least our reader and writers are consistent with
each other.

For example:

gpsbabel -t -i gpx -f ~/Downloads/1.txt -xtransform,wpt=trk -o naviguide -F
- | gpsbabel -t -i naviguide -f - -o gpx -F -

[...]

PR #169, opened 2018-05-01
Add Naviguide tracks support.

zstadler:
Following https://github.com/gpsbabel/gpsbabel/issues/81 and https://github.com/IsraelHikingMap/Site/issues/689

Add option for specifying the input data type when writing a Naviguide file.


naviguide.cc
```diff
@@ -67,36 +67,52 @@ typedef struct {
...
+    trk - read file as track
 
-static char temp_short_name[5];
+ */
+static char* read_as = NULL;
+gpsdata_type read_type = unknown_gpsdata;
 
+/*
+    File writing options:
+    First non-empty data type, as a string with the letters:
```
robertlipe, line 89: Unfortunately, our existing art on some formats are confusing.

It's strongly encouraged that we write any type of data we have and not really use the -w/-t/-r flags to control the writing. Can both exist at the same time in this format? If not, I'd much rather we use the doing_rte/doing_trk style code instead of adding suboptions.

naviguide.cc
```diff
@@ -67,36 +67,52 @@ typedef struct {
...
 
 static
 arglist_t ng_args[] = {
   {
-    "output", &process, "'wp' - Create waypoint file , 'rte' - Create route file",
+    "output", &read_as, "'wp' - Read file as waypoints, 'rte' - Read file as route, 'trk' - Read file as track",
     "rte", ARGTYPE_STRING, ARG_NOMINMAX, nullptr
   },
+  {
+    "input", &write_as, "Select the source type for writing, with optional priority.",
```
robertlipe, line 113: This probably wasn't well though through. Let's try to identify the files on read and just do the right thing without making users select anything at all here.

naviguide.cc
```diff
@@ -285,30 +301,67 @@ header_write()
...
 wr_init(const QString& fname)
 {
-  file_out = gbfopen_le(fname, "wb", MYNAME);
-  ng_fill_header_default();
-  if (NULL != reorder)
-    if (!case_ignore_strcmp(reorder, "y")) {
-      reorder_wp = 1;
+  int i;
+
+  for (i=0; write_as[i] && !nof_wp && i<3; i++) {
```
robertlipe, line 330: "for (int i = 0; ..."
It'd be way less fragile if we could just write everything. Will that work?

naviguide.cc
```diff
@@ -326,16 +379,19 @@ rd_init(const QString& fname)
...
   ng_read_file_header();
 
-  if (NULL != process) {
-    if (!case_ignore_strcmp(process, "wp")) {
-      process_rte = 0;
-    }
-    if (!case_ignore_strcmp(process, "rte")) {
-      process_rte = 1;
-    }
+  if (!case_ignore_strcmp(read_as, "wp")) {
```
robertlipe, line 382: These could be QString::compare(s1, s2, Qt::CaseInsensitive) and chained elses, but it's probably better to just add everything all the time independent of options.

robertlipe (commented):
Sorry for the delay in the review. This just got stuck.

We have a strong preference to not special case the data types in a suboption. We can use doing_FOO if we have to, but it's much better if we can just silently do the right thing for all cases - read any data in the file and write a file with all the data we have.

zstadler:
Stale

PR #168, opened 2018-04-30
Fix warning in .Pro build by forcing HAVE_UNISTD

robertlipe:
I'm not sure if it's new in Qt 5.9, but it's trivial (I think) to fix and harmless. When building files like gzread and gzwrite, MacOS would complain that we were missing declarations to extremely well known system functions (write, close, read, seek, ...) Those live in <unistd.h>, so we nudge the Makefile generator to trigger its inclusion for MacOS and Linux.

PR #175, opened 2018-05-12
add config tests in GPSBabel.pro.

tsteven4:
specifically, we look for stdarg.h and unistd.h for zlib.

In #168 a warning was fixed related to unistd.h by assuming HAVE_UNISTD_H.  This PR actually tests for unistd.h and conditionally sets HAVE_UNISTD_H. It also tests for stdarg.h, and conditionally sets HAVE_STDARG_H.  Both of these macros are used by zlib.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,9 @@
 /gpsbabel.html
 /gpsbabel.pdf
 /Makefile
+/.qmake.cache
+/.qmake.stash
+/GPSBabel
 *.o
 /GPSBabel[0-9]*.[0-9]*.[0-9]*/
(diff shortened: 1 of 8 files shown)


robertlipe (approved):
Nice find. 

I wouldn't go too far down this rabbit hole; I think it's perfectly reasonable to distinguish Windows and everything else POSIX-ish for stuff like this, but this seems pretty lightweight. 

TIL that qmake has a lot of tests like qtcompiletest. Maybe we can make it take two minutes for the first build while we check for eight bit bytes like configure does. :-)

PR #179, opened 2018-05-19
Add mac build to travis CI

### LINKED ISSUE #178: mac gui build requires config.h
The mac gui build requires config.h, but it is not generated using the qmake flow, i.e. "qmake GPSBabel.pro && make && cd gui && qmake app.pro && make"

Why do we define HAVE_CONFIG_H?

https://github.com/gpsbabel/gpsbabel/blob/82c969bf024673f33ac5d525a86314c65ff89406/gui/app.pro#L12

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang++ -c -pipe -stdlib=libc++ -O2 -std=gnu++11  -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk -mmacosx-version-min=10.10 -Wall -W -fPIC -DHAVE_CONFIG_H -DHAVE_WEBENGINE -DQT_NO_DEBUG -DQT_WEBENGINEWIDGETS_LIB -DQT_WEBENGINECORE_LIB -DQT_QUICK_LIB -DQT_PRINTSUPPORT_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_WEBCHANNEL_LIB -DQT_QML_LIB -DQT_NETWORK_LIB -DQT_XML_LIB -DQT_POSITIONING_LIB -DQT_CORE_LIB -I. -I../Qt/5.9.5/clang_64/lib/QtWebEngineWidgets.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtWebEngineCore.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtQuick.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtPrintSupport.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtWidgets.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtGui.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtWebChannel.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtQml.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtNetwork.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtXml.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtPositioning.framework/Headers -I../Qt/5.9.5/clang_64/lib/QtCore.framework/Headers -Iobjects -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/OpenGL.framework/Headers -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/System/Library/Frameworks/AGL.framework/Headers -Itmp -I../Qt/5.9.5/clang_64/mkspecs/macx-clang -F/Users/travis/build/tsteven4/gpsbabel/Qt/5.9.5/clang_64/lib -o objects/upgrade.o upgrade.cc
upgrade.cc:28:10: fatal error: '../config.h' file not found
#include "../config.h"
         ^
1 error generated.
make: *** [objects/upgrade.o] Error 1

tsteven4:
1. This adds a build on macOS to travis CI.  The travis osx queues seem to be short at this time.
2. This adds a check target to GPSBabel.pro that just runs testo.
3. This attempts to decouple the GUI from config.h by making changes to app.pro and upgrade.cc.  These changes should be validated, and if valid this PR could resolve #178.
4. The new script, tools/travis_script_osx, attempts to build the CLI and test it, build the GUI, and then generate a dmg.  This was created based on some targets in Makefile.in.  There is an attempt at the end of the script to examine the resulting dmg file, there is likely a better way to accomplish this.

~~TO BE ADDED LATER:~~

5. The dmg could be uploaded to https://github.com/gpsbabel/gpsbabel/releases as a "continuous" Pre-release.  See for example https://github.com/tsteven4/gpsbabel/releases.  Although the example doesn't have the dmg you can see what a continuous Pre-release would look like.  It would be updated on every successful build.  There isn't a way to keep build artifacts on travis like there is on appveyor.


tsteven4:
sometimes the mac travis log ends prematurely even though the build succeeds.  Other times the log looks complete.  In either case it ends with 

> /Users/travis/.travis/job_stages: line 166: shell_session_update: command not found

It is unclear to me what this is.  It would be good to verify that travis_script_osx is exiting with status 0 on a real mac, and not throwing an error in some command (which we would expect to be caught by -e, but maybe not on travis)

tsteven4:
The dmg is being uploaded.  For pushes it goes to https://github.com/gpsbabel/gpsbabel/releases as a continuous pre-release where it lives until the next push.  For PRs it goes to https://transfer.sh where it can live for up to 14 days.  Some fine tuning is probably required to not upload pushes to branches.  Or we may always want to use transfer.sh and skip github.

.travis.yml
```diff
@@ -1,7 +1,33 @@
...
+    - os: linux
+      dist: trusty
+      sudo: required
+      services: docker
+      compiler: clang
+    - os: osx
+      compiler: clang
+
+install:
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then QT_CI_PACKAGES=qt.595.clang_64,qt.595.qtwebengine PATH=$PATH:$PWD/tools ./tools/qt-5.9.5-osx; fi
```
robertlipe: Does the yaml allow some of these tests to be collapsed? Ignore if not.

gui/upgrade.cc
```diff
@@ -81,86 +75,10 @@ bool UpgradeCheck::isTestMode()
   return testing;
 }
 
-// In Qt 5.4, QSysInfo got classes that do this better. Travis is stuck
-// on older version of Qt, so keep it building there for now.
+// In Qt 5.4, QSysInfo got classes that do this better.
```
robertlipe: Are we OK with the check in gui/main.cc handling the test for QT_VERSION_CHECK?  Maybe we move it into one of the .h's that's included everywhere so it'll fail earlier.


I'm OK with the new way being the only way and being free of that OS-specific clutter that we had.  It's always been a maintenance pain for us.

tools/qt-5.9.5-osx
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+# Reference:
+#   https://github.com/musescore/MuseScore/blob/master/build/travis/job_macos/install.sh
+
+echo Downloading Qt
+set -v
+set -e
+
+DOWNLOAD_URL=https://download.qt.io/archive/qt/5.9/5.9.5/qt-opensource-mac-x64-5.9.5.dmg
```
robertlipe: It's a bummer we have to download this on every build, but pretty impressive that the extraction and environment build all works. Wow!

tools/travis_script_osx
```diff
@@ -0,0 +1,44 @@
...
+$QMAKE app.pro && make
+$LUPDATE app.pro
+$LRELEASE app.pro
+popd
+
+# package the app
+mkdir -p gui/objects/GPSBabelFE.app/Contents/MacOS/translations
+cp gui/*.qm  gui/objects/gpsbabelFE.app/Contents/MacOS/translations
+cp GPSBabel gui/objects/GPSBabelFE.app/Contents/MacOS/gpsbabel
+# this help seems usless, why bother?
```
robertlipe: Agreed. Start of a bad idea that was never taken further. Please drop it.

tools/travis_script_osx
```diff
@@ -0,0 +1,44 @@
...
+cp -r gui/help/*.html  gui/objects/GPSBabelFE.app/Contents/MacOS
+cp gui/gmapbase.html gui/objects/GPSBabelFE.app/Contents/MacOS
+# what about the COPYING.txt file?
```
robertlipe: That's an excellent point. Since we never run an installer, we never show the license on MacOS. How have we not noticed this in the last ten years? We probably should have a license in our help screen and show it there.
Obviously, if this has been wrong for hundreds of thousands of installs, this can be a separate CL with a different priority. :-)

tools/travis_script_osx
```diff
@@ -0,0 +1,44 @@
...
+pushd gui/objects
+$MACDEPLOYQT GPSBabelFE.app -executable=GPSBabelFE.app/Contents/MacOS/gpsbabel -dmg -verbose=2
+popd
+
+# what is in there?
+hdiutil attach -noverify gui/objects/GPSBabelFE.dmg
+find /Volumes/GPSBabelFE -ls
+hdiutil detach /Volumes/GPSBabelFE
+
+mv gui/objects/GPSBabelFE.dmg gui/objects/GPSBabel-${VERSIONID}.dmg
```
robertlipe, line 46: Are we OK with these version numbers looking like "b9ddf503" and not really sortable in the 'semantic versioning' sense or readily human-recognizable?

gui/app.pro
```diff
@@ -9,7 +9,6 @@ CONFIG(debug, debug|release) {
 # For Mac, x86 and x64, but not PPC binary.   Ignored on other OSes.
```
robertlipe: Let's drop this hunk while we're here. PPC has been dead for a long time.

tools/travis_script_osx
```diff
@@ -0,0 +1,44 @@
...
+MACDEPLOYQT=${MACDEPLOYQT:-macdeployqt}
+VERSIONID=${VERSIONID:-$(git rev-parse --short HEAD)}
+
+# build and test the CLI
+$QMAKE GPSBabel.pro && make && make check
+
+# build the GUI
+pushd gui
+$QMAKE app.pro && make
+$LUPDATE app.pro
```
robertlipe, line 24: n.b. lupdate and lrelease used to be in the Mac build because that was what I built most. This ensured the translations were up to date when I released.

robertlipe (commented):
Nice cleanups and use of external tooling. Thanx.

Minor nits, but please submit as you see fit.

tools/qt-5.9.5-osx
tsteven4: I enabled caching of the Qt install.  The cache has to be downloaded from offsite just like the qt installer, but caching is about 11x faster than installing from scratch (650s -> 60s).

tools/travis_script_osx
tsteven4: I deleted the help/gpsbabel.html from the mac dmg, the windows installer, and makelinuxdist.

gui/app.pro
tsteven4: ok

.travis.yml
tsteven4: yes you can use a yaml pipe (|) on a sequence (-), i.e "- |", but it seems to be very seldom used in .travis.yml files.  A more standard approach is to move all but trivial things into scripts in your repo, I have taken this approach as recommended by https://docs.travis-ci.com/user/customizing-the-build#Implementing-Complex-Build-Steps

tools/travis_script_osx
tsteven4: I added COPYING.txt to the dmg but that is it, i.e. it isn't integrated into the install process with a screen to accept it.

tools/travis_script_osx
tsteven4: I added the travis build number, so the dmg is named something like GPSBabel-295-b787e2c.dmg and the appveyor one is GPSBabelTest-1.0.205-b787e2c-Setup.exe where b787e2c is the first 7 characters of the git 40 character revision identifiers (SHA-1 hash).  The hash component ties us to the exact source version in the installer.

tsteven4: You can see an example at https://github.com/tsteven4/gpsbabel/releases

tools/travis_script_osx
tsteven4: They will be up to date when travis creates the dmg, but the updated files are not committed.

gui/upgrade.cc
tsteven4: I just took the version check out of upgrade.cc and fixed up the comment.  We still have gui/main.cc enforcing 5.7 minimum.

robertlipe:
Awesome. The ...7e2c.dmg image seems to run for me. Upgrade check fails
because of the funky number and there's a spare vertical separator at the
bottom that I need to chase down, but this is super promising.

tsteven4:
Glad to know the dmg actual works, i can't try it!

The github uploads aren't working on github.com/gpsbabel/gpsbabel.  I am 
chasing that down, they worked fine on github.com/tsteven4/gpsbabel.

PR #185, opened 2018-05-29
clang-tidy cleanup

tsteven4:


tsteven4:
I will resubmit this after a minor process cleanup.

PR #190, opened 2018-06-07
Create a filter base class and make all our filters derived classes of this base class.

tsteven4:
All tests pass when travis and docker cooperate, even a new test for the previously untested swap filter.

All the filters are always constructed in filter_vecs.cc much like they always existed before.  It should be possible to just dynamically construct the needed filter on use, but I haven't pursued this.

All the filter_vecs_t vectors defined in each filter are gone.  The only useful thing that remained in those vectors was the pointer to the args, and they are getting fetched with a getter.  I note that in the case of formats the vecs are likely also being used to determine if a format can read and/or write.  We can always call init, process, deinit, exit functions.  If they aren't defined in the derived class we will just use the no-op copy from the base class.

Implementation Issues:
1. The link between non-static member function pointers and function pointers.  I have glued this together through the classic wrapper using static member functions.  This works because we don't try to run two copies of a filter at once.

2.  initializer for flexible array member.  I am having to explicitly set the length of the args arrays.


tsteven4:
I have a potential solution to Implementation Issue 1:
I template route_disp_all, track_disp_all, waypt_disp_all so they can accept a function pointer or functor.
Then I create functor(s) as nested classes for of each the filters that use *_disp_all.  In the constructor for the functor I pass a reference to the filter so the functor can access non-static methods and non-static members of the filter.
I have it running for the height filter.

tsteven4:
This looks pretty sweet now, although:

1. It is a bit disappointing to have to overload common_disp_all so many times to catch nullptr (at compile time).
2. The filter argument lengths are hardcoded as noted in the original 2. above.

PR #200, opened 2018-06-18
cleanup defs.h orgainizing route and waypt elemements.

tsteven4:
The following prototypes were deleted, they had no definitions:
waypt_free
waypt_dupe
waypt_new
route_add
waypt_alloc_gc_data
waypt_empty_gc_data

The following functions were never used, and they had no prototypes:
void routes_to_tracks();
void tracks_to_routes();

The following function had a repeated prototype:
route_flush

The following function is missing a prototype but is explictly listed by the using file:
update_common_traits

In addition to resolving the above issues, some sections of defs.h
were moved based on if they came from waypt.cc or route.cc,
when they were required, and grouping of similar elements.

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -39,6 +39,7 @@
 # include "src/core/datetime.h"
 
 #define CSTR(qstr) (qstr.toUtf8().constData())
+#define CSTRc(qstr) (qstr.toLatin1().constData())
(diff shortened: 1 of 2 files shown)


tsteven4:
A method to verify the intended prototytpe changes listed above is to compare the sorted files before and after.

PR #220, opened 2018-07-05
Add CMakeLists.txt.

tsteven4:
This is experimental at this stage.

This is useful with CLion.

This has not been tested on MacOS.

This will create additional maintenance work as we have
1. configure
2. qmake
3. cmake .
4. xcode?
5. vcxproj.  Highly recommend qmake -tp vc GPSBabel.pro to create vcxproj file.
The first two are currently tested in CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,260 @@
+
+# It isn't clear what the minimum required version is.
+# tested against 3.10.2 on ubuntu bionic 2018/07/05
+# tested against 3.5.1 on ubuntu xenial 2018/07/05
+# tested against MSVC 2017 which included 3.11.* 2018/07/05
+cmake_minimum_required(VERSION 3.5.1)
+
+project(gpsbabel)
+
+set(CMAKE_CXX_STANDARD 11)
+
+# Find includes in corresponding build directories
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,260 @@
...
+  set(LIBS ${LIBS} setupapi)
+endif()
+
+if(MSVC)
+  add_definitions(-D_CRT_SECURE_NO_DEPRECATE)
+  add_compile_options(/MP -wd4100)
+endif()
+
+if (UNIX AND NOT APPLE)
+  add_definitions(-DHAVE_LINUX_HID)
```
GPSBabelDeveloper, line 217: With delbin safely behind us, I don't think anything uses HID. I'd hope nothing does in the future - it's a terrible choice for machine protocols.

GPSBabelDeveloper (commented):
Nice. 

I think there's a space for things like this that aren't fully maintained but (unlike xcode or some of the MSVC mutants) aren't actually hard to just add source files to.  I had a rewrite of the 'building' page in progress once. I can be sure to spin some of these "present, but possibly out of date" build systems. IMO, *.pro and configure are enough...with the second of those just hanging on.

If this makes your (our) life easier, go for it. Thanx!

PR #224, opened 2018-07-09
limit scope of loop vars, prefer prefix operators.

tsteven4:


robertlipe:
Thank you for the steady progress. FWIW, I wouldn't get bogged down in the recursive "you touched this file, you now own everything about it" fussing from Codacy. Totally go with your judgement on whether it's best to do in the same commit, another commit, or not at all. I got bogged down by that recently.

PR #226, opened 2018-07-09
refactor cases of "Declarator is reassigned in all

tsteven4:
paths before being read"
when the declarator and the subsequent assignment can be joined.

### DIFF
--- a/an1.cc
+++ b/an1.cc
@@ -129,12 +129,11 @@ ReadString(gbfile* f, short len)
 static void
 ReadGuid(gbfile* f, GUID* guid)
 {
-  int i = 0;
   guid->l = ReadLong(f);
-  for (i = 0; i < 3; i++) {
+  for (int i = 0; i < 3; i++) {
     guid->s[i] = ReadShort(f);
   }
-  for (i = 0; i < 6; i++) {
+  for (int i = 0; i < 6; i++) {
     guid->c[i] = ReadChar(f);
   }
 }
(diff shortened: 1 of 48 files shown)


an1.cc
```diff
@@ -832,18 +824,13 @@ static void Write_AN1_Waypoints(gbfile* f)
...
-  an1_line_record* rec = nullptr;
-  an1_vertex_record* vert = nullptr;
   route_head* rte_head;
   Waypoint* wpt_tmp;
 
   ReadShort(f);
-  count = ReadLong(f);
-  for (i = 0; i < count; i++) {
-    rec = Alloc_AN1_Line();
+  unsigned long count = ReadLong(f);
```
GPSBabelDeveloper, line 831: Even if count wasn't used, was it needed to seek the read pointer ahead? (On a fone in bed. I may not have the whole picture. I doubt our coverage of an1 is very good.)

GPSBabelDeveloper (commented):
Much better. Thanks!

an1.cc
GPSBabelDeveloper: It was the occurrence above in read waypoints that I meant to comment on.
Sorry.

an1.cc
```diff
@@ -684,15 +679,12 @@ static void Write_AN1_Bitmaps(gbfile* f)
...
-  an1_waypoint_record* rec = nullptr;
   Waypoint* wpt_tmp;
   char* icon = nullptr;
   ReadShort(f);
-  count = ReadLong(f);
-  for (i = 0; i < count; i++) {
-    rec = Alloc_AN1_Waypoint();
+  unsigned long count = ReadLong(f);
+  for (unsigned long i = 0; i < count; i++) {
+    an1_waypoint_record* rec = Alloc_AN1_Waypoint();
```
tsteven4, line 687: old line 687-689 declare the variable and initialize it.  The initial value is never used.
old lines 693-695 rewrite all three variables.  The initial values were never read.
new lines 685-687 declare the variables, and use the rewritten values from 693-695, and duplicate the loop.
It looks good to  me.

robertlipe: Me, too.  I thought I saw a call to ReadLong(f) being dropped, but on a Real Computer, I see it's not. I suspect the diff just rendered badly on the "small" screen .I read it several times. Sorry for the distraction.

PR #232, opened 2018-07-12
clean up some Variable * is reassigned a value before the old one has…

tsteven4:
… been used.

one of these was actually a very old bug.

### DIFF
--- a/gdb.cc
+++ b/gdb.cc
@@ -844,7 +844,7 @@ read_route()
         bounds.max_alt = FREAD_DBL;
       }
       bounds.min_lat = FREAD_LATLON;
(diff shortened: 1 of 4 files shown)


gdb.cc
```diff
@@ -844,7 +844,7 @@ read_route()
         bounds.max_alt = FREAD_DBL;
       }
       bounds.min_lat = FREAD_LATLON;
-      bounds.min_lat = FREAD_LATLON;
+      bounds.min_lon = FREAD_LATLON;
```
robertlipe, line 847: Now THAT is a gem of a find. This class of mistake actually has a name: the Last Line Effect. T

PR #233, opened 2018-07-13
Provide examples of how to set up CSV when processing UK data

Hornbydd:
Wanted to enhance the help file with additional examples to help people construct the correctly formatted CSV file when processing UK data.

### DIFF
--- a/xmldoc/formats/unicsv.xml
+++ b/xmldoc/formats/unicsv.xml
@@ -103,7 +103,27 @@
      GC1A37,36.090683,-86.679550,The Troll by a182pilot &amp; Family
    </screen>
 </para>
+<para>
+   If processing data from the UK, GPSBabel can process coordinates using the full 12 figure alpha numeric
+   or X,Y values (often referred to as Eastings/Northings).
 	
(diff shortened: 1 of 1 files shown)


xmldoc/formats/unicsv.xml
```diff
@@ -103,7 +103,27 @@
      GC1A37,36.090683,-86.679550,The Troll by a182pilot &amp; Family
    </screen>
 </para>
+<para>
+   If processing data from the UK, GPSBabel can process coordinates using the full 12 figure alpha numeric
+   or X,Y values (often referred to as Eastings/Northings).
+	
+   Below are two examples of how to set up the CSV file using both approaches.
+	
+   <emphasis>Example 1 - CSV input for UK data with XY coordinates</emphasis>
```
GPSBabelDeveloper, line 112: If we use <example id="unicsv-uk-xy"><title>CSV input for ... instead of <emphasis>, we'll get automatic references to that in the generated TOC, I think.

Should we call this "British National Grid" here like we do in the explanation in this file for "bng_z" ?

tsteven4:
I incorporated this into #230.  Thanks

PR #218, opened 2018-07-04
Cleanups

robertlipe:
Grrr. Github desktop works exactly like git command line doesn’t for
uncommitted files. :-/

PR #234, opened 2018-07-14
use reinterpret_cast with queue processing

tsteven4:
instead of a static cast to a void pointer followed
by a static cast to a Waypoint*.

### DIFF
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -158,7 +158,7 @@ void ArcDistanceFilter::process()
   foreach (Waypoint* wp, waypt_list) {
 #else
   QUEUE_FOR_EACH(&waypt_head, elem, tmp) {
-    Waypoint* wp = (Waypoint*) elem;
+    Waypoint* wp = reinterpret_cast<Waypoint *>(elem);
 #endif
     extra_data* ed = (extra_data*) wp->extra_data;
     wp->extra_data = nullptr;
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -47,8 +47,7 @@ void ArcDistanceFilter::arcdist_arc_disp_wpt_cb(const Waypoint* arcpt2)
 #else
     queue* elem, *tmp;
     QUEUE_FOR_EACH(&waypt_head, elem, tmp) {
-      void* vwaypointp = static_cast<void*>(elem);
-      Waypoint* waypointp = static_cast<Waypoint*>(vwaypointp);
+      Waypoint* waypointp = reinterpret_cast<Waypoint*>(elem);
 #endif
       double dist;
       extra_data* ed;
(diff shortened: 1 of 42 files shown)


tsteven4:
Before #218 reinterpret cast would have been used in most cases here by Resharper C++ or clang-tidy.  In the remaining cases Resharper C++ and clang-tidy sort of give up and say "C-style casts are discouraged; use static_cast/const_cast/reinterpret_cast".  The lack of a specific suggested fix seems to be related to confusion with constness.  

I also restored the constness of the queue* and waypoint* in igc.cc to the state they were before #218.
I also applied the same fixes in igc.cc to the else case directly below the original two.

GPSBabelDeveloper (approved):
Agreed. This was bad judgement on my part. In most of these cases, I didn't set out to "fix" these at all, I was just blocked by the presubmit because I touched something else in those files. I'd found something that worked and just applied it to other editor tabs. I just plain forgot about reinterpret_cast because it's so rare in "real" C++ - which our "bounce everything through a void *" in queue_t is not. (...which this program has because of my OS background: it's super easy to safely mix list handling in assembly and C this way)

Two points for anyone else playing along.
1) In probably 80% of the cases we're currently using C casts, we simply have the wrong data type and we're forcing it with a hammer.  Most xmallocs should be operator new, casting away unsigned or const is just bad, etc.
2) I'd rather spend programmer time on replacing queue* uses with real C++ containers. This change undoes damage I introduced, so thanks for that. Newer C++ that can automatically parallelize iterators is very nice.

Thanx for the fix.

tsteven4:
https://github.com/gpsbabel/gpsbabel/pull/234/commits/178077a3a93d3971bc9bed6b3e99db0a0f06907d should make queue casting consistent.  Hopefully when NEW_Q or whatever comes along that supplies a queue that can keep track of the type of elements these casts will become redundant and inspection tools will be able to detect that and remove them.

robertlipe:
Most of those should be a QList<Waypoint *>, but that's more than a
mechanical fix.

tsteven4:
Agreed.  The actual usage looks like this.  I think these are all queue related.  A template-based container class like QList should do nicely.
      1 reinterpret_cast<category_t *>
      1 reinterpret_cast<gpx_global_entry *>
      1 reinterpret_cast<mag_rte_elem *>
      2 reinterpret_cast<const exif_ifd_t *>
      2 reinterpret_cast<const exif_tag_t *>
      2 reinterpret_cast<exif_app_t *>
      2 reinterpret_cast<inifile_entry_t *>
      2 reinterpret_cast<session_t *>
      2 reinterpret_cast<uniq_shortname *>
      3 reinterpret_cast<inifile_section_t *>
      4 reinterpret_cast<field_map_t *>
      6 reinterpret_cast<exif_ifd_t *>
      7 reinterpret_cast<exif_tag_t *>
      8 reinterpret_cast<route_head *>
     17 reinterpret_cast<const Waypoint *>
     94 reinterpret_cast<Waypoint *>

Issue #235, opened 2018-07-15
gpx "global" and metadata handling of non-compliant input


tsteven4:
The gpx reader saves gpx 1.0 "global" data and gpx 1.1 metadata.  If this saved data exists the gpx writer uses it, i.e. it forwards it from the input to the output.

The gpx 1.0 "global" data is

> <xsd:element name="name"    type="xsd:string" minOccurs="0"/>   
        <xsd:element name="desc"    type="xsd:string" minOccurs="0"/>   
        <xsd:element name="author"    type="xsd:string" minOccurs="0"/>   
        <xsd:element name="email"   type="gpx:emailType" minOccurs="0"/>  
        <xsd:element name="url"     type="xsd:anyURI" minOccurs="0"/>   
    <xsd:element name="urlname"   type="xsd:string" minOccurs="0"/>
        <xsd:element name="time"    type="xsd:dateTime" minOccurs="0"/>   
        <xsd:element name="keywords"  type="xsd:string" minOccurs="0"/>   
        <xsd:element name="bounds"    type="gpx:boundsType" minOccurs="0"/> 

Note that maxOccurs is not specified, so the default value of 1 is used.  These elements cannot be repeated in valid gpx 1.0.

The gpx 1.1 metadata is

>    <xsd:element name="name"		type="xsd:string"		minOccurs="0">
    <xsd:element name="desc"		type="xsd:string"		minOccurs="0">
    <xsd:element name="author"		type="personType"		minOccurs="0">
    <xsd:element name="copyright"	type="copyrightType"	minOccurs="0">
    <xsd:element name="link"		type="linkType"			minOccurs="0" maxOccurs="unbounded">
    <xsd:element name="time"		type="xsd:dateTime"		minOccurs="0">
    <xsd:element name="keywords"	type="xsd:string"		minOccurs="0">
    <xsd:element name="bounds"		type="boundsType"		minOccurs="0">
    <xsd:element name="extensions"	type="extensionsType"	minOccurs="0">

The only element that can be repeated is link.

We don't currently handle gpx 1.1 metadata elements author, copyright, link, and extensions.  These elements have more complicated content that simple text.

1. It appears that the gpx_global_entry queues are unnecessarily complicated in that they aren't used for anything that can legally exist more than once.  If we handled gpx 1.1 metadata link elements then a solution that could track multiple occurrences of an element would be needed.

[...]

PR #236, opened 2018-07-15
update gpx handling of global/metadata.

tsteven4:
use qt containers instead of queue.
add support for gpx/metadata/link.

### DIFF
--- a/gpx.cc
+++ b/gpx.cc
@@ -189,6 +189,8 @@ typedef enum  {
  * Note that all gpx 1.0 "global data" has a maxOccurs limit of one,
  * which is the default if maxOccurs is not in the xsd.
  * The only gpx 1.1 metadata that has a maxOccurs limit > one is link.
+ * However, multiple gpx files may be read, and their global/metadata
+ * combined, by this implementation.
  */
 struct GpxGlobal {
   QStringList name;
@@ -201,7 +203,7 @@ struct GpxGlobal {
   QList<UrlLink> link;
   /* time and bounds aren't here; they're recomputed. */
 };
-static GpxGlobal* gpx_global;
+static GpxGlobal* gpx_global = nullptr;
 
 static void
 gpx_add_to_global(QStringList& ge, const QString& s)
@@ -250,9 +252,11 @@ gpx_write_gdata(const QStringList& ge, const QString& tag)
 {
   if (!ge.isEmpty()) {
     writer->writeStartElement(tag);
-    // TODO: this seems wrong, it is concatenating element content
-    // from multiple illegally repeated elements into one.
-    // However, this does result in output that complies with the schema.
+    // TODO: This seems questionable.
+    // We concatenate element content from multiple elements,
+    // possibly from muliple input files, into one element.
+    // This is necessary to comply with the schema as
+    // these elements have maxOccurs limits of 1.
     for (const auto& str : ge) {
       writer->writeCharacters(str);
       /* Some tags we just output once. */
--- a/gpx.cc
+++ b/gpx.cc
@@ -190,18 +190,18 @@ typedef enum  {
  * which is the default if maxOccurs is not in the xsd.
  * The only gpx 1.1 metadata that has a maxOccurs limit > one is link.
  */
-static
-struct gpx_global {
-  QStringList* name;
-  QStringList* desc;
-  QStringList* author;
-  QStringList* email;
-  QStringList* url;
-  QStringList* urlname;
-  QStringList* keywords;
-  QList<UrlLink>* link;
+struct GpxGlobal {
+  QStringList name;
+  QStringList desc;
+  QStringList author;
+  QStringList email;
+  QStringList url;
+  QStringList urlname;
+  QStringList keywords;
+  QList<UrlLink> link;
   /* time and bounds aren't here; they're recomputed. */
-}* gpx_global ;
+};
+static GpxGlobal* gpx_global;
 
 static void
 gpx_add_to_global(QStringList& ge, const QString& s)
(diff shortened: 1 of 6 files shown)


tsteven4:
This is related to #235.  As it stands the questionable handling of illegally repeated elements is preserved.

gpx.cc
```diff
@@ -264,24 +246,23 @@ gpx_reset_short_handle()
...
-      break;
+  if (!ge.isEmpty()) {
+    writer->writeStartElement(tag);
+    // TODO: this seems wrong, it is concatenating element content
+    // from multiple illegally repeated elements into one.
+    // However, this does result in output that complies with the schema.
+    foreach (const QString& str, ge) {
+      writer->writeCharacters(str);
+      /* Some tags we just output once. */
+      if ((0 == strcmp(tag, "url")) ||
```
GPSBabelDeveloper: Did we decide that QString == "literal" is OK?

gpx.cc
```diff
@@ -264,24 +246,23 @@ gpx_reset_short_handle()
...
-    /* Some tags we just output once. */
-    if ((0 == strcmp(tag, "url")) ||
-        (0 == strcmp(tag, "email"))) {
-      break;
+  if (!ge.isEmpty()) {
+    writer->writeStartElement(tag);
+    // TODO: this seems wrong, it is concatenating element content
+    // from multiple illegally repeated elements into one.
+    // However, this does result in output that complies with the schema.
+    foreach (const QString& str, ge) {
```
GPSBabelDeveloper: for (auto &str: ge) { ?

gpx.cc
```diff
@@ -461,7 +445,7 @@ tag_gpx(const QXmlStreamAttributes& attr)
    * that use them to the writer.
    */
   const QXmlStreamNamespaceDeclarations ns = reader->namespaceDeclarations();
-  for (const auto & n : ns) {
+  for (const auto& n : ns) {
```
GPSBabelDeveloper, line 452: Let's be consistent with for instead of for_each if we can.

gpx.cc
```diff
@@ -1887,13 +1904,14 @@ gpx_write()
...
 gpx_free_gpx_global()
 {
-  gpx_rm_from_global(&gpx_global->name);
-  gpx_rm_from_global(&gpx_global->desc);
-  gpx_rm_from_global(&gpx_global->author);
-  gpx_rm_from_global(&gpx_global->email);
-  gpx_rm_from_global(&gpx_global->url);
-  gpx_rm_from_global(&gpx_global->urlname);
-  gpx_rm_from_global(&gpx_global->keywords);
+  delete gpx_global->name;
```
GPSBabelDeveloper: We could make gpx_global a class and then all the containers inside it would delete when gpx_global is deleted. One less xmalloc/xfree and better future safety.

GPSBabelDeveloper (commented):
I like it. I've never really had a good plan on handling both GPX 1.0 and 1.1 with their annoying incompatibilities.

This is very much the direction I'm going with the xcsv/csv_util handlers.

gpx.cc
tsteven4: tag is a char string still.

robertlipe:
Change the function signature. We're constructing a QString when calling
writeCharacters() anyway, right?

gpx.cc
tsteven4: foreach (auto& str, ge)
works and preserves const, but I haven't pursued auto too much.
clang-tidy modernize-use-auto misses this, probably because foreach is a qtism.

robertlipe:
I think foreach is qt and "for (auto& str : ge)" is the hipster-approved
way in C++11.

gpx.cc
tsteven4: I like that!  and they will be constructed as well.

tsteven4:
so this begs the big question, should we maintain this questionable handling of input that violates the schema?  We go through a bunch of work to keep track of and concatenate the content of elements that are not suppose to occur more than once.

GPSBabelDeveloper:
In GPX 1.0 they occur once and in 1.1, they can repeat, IIRC.

GPSBabelDeveloper:
I think that, in general, we make a mess of extended schema handling. If you do something like include a schema under a different namespace, we don't recognize it, for example.

tsteven4:
I beg to differ.  See my extracts from the 1.1 schema in #235.  Only link can occur more than once.  All the other metadata elements don't specify a maxOccurs in the 1.1 schema, thus maxOccurs defaults to 1.

tsteven4:
> if you do something like include a schema under a different namespace

That has been a sore spot in my craw for a long time, but lets save that for another day.

GPSBabelDeveloper:
Ah, now I remember the reason. This isn't to say it's a good reason, only that it was A reason. :-)
Merge two GPX 1.1 files, each with different extension schemas. If they were well designed, you could glom them together in <extensions>, include the schema refs, and all would be good. 

In the early days of GPX 1.1, Dan Foster of Topografix (pretty much the driving force for GPX 1.1) was very hip on "microschemas" where there would be one for HR and one for line style and one for point style and and for temperature and such. So his software would create GPX 1.1's with a dozen schemas. Now, merge that with a data file from another vendor and there was a bunch.

I think that URL was one of my original "OMG, this sucks so hard in C" breaking points where the original change from url_link to url_links just was awful.

tsteven4:
> Did we decide that QString == "literal" is OK?

https://woboq.com/blog/qstringliteral.html would lead me to beleive
I think that for a Qstring qstr,
qstr == QLatin1String("literal")
is preferred because the QString operator == specifically handles QLatin1String, although
qstr == "literal"
works if QT_NO_CAST_FROM_ASCII and QT_RESTRICTED_CAST_FROM_ASCII are not defined.
Trashing through the Qt source seems to suggest there may be an advantage to the QStringLiteral approach, but I suspect we hardly care.

tsteven4:
> Ah, now I remember the reason.

I gather there is justification then to maintain the status quo.  It is certainly easier to do with Qt than with queues, and that is what I implemented anyway.

PR #238, opened 2018-07-20
update exif format replacing queues.

tsteven4:
So far I have chosen not to pursue a shift from pointers to references to minimize the changes in the diff.


exif.cc
```diff
@@ -440,24 +414,25 @@ exif_read_ifd(exif_app_t* app, const uint16_t ifd_nr, gbsize_t offs,
...
 #endif
 
-    ENQUEUE_TAIL(&ifd->tags, &tag->Q);
-
     tag->id = gbfgetuint16(fin);
     tag->type = gbfgetuint16(fin);
     tag->count = gbfgetuint32(fin);
     tag->size = exif_type_size(tag->type) * tag->count;
-    tag->data = &tag->value;
+    tag->data.reset(reinterpret_cast<char *>(xcalloc(4, 1)), xfree);
```
robertlipe: ick

exif.cc
```diff
@@ -440,24 +414,25 @@ exif_read_ifd(exif_app_t* app, const uint16_t ifd_nr, gbsize_t offs,
...
+    
 
     if (BYTE_TYPE(tag->type) && (tag->count <= 4)) {
-      gbfread(tag->data, 4, 1, fin);
+      gbfread(tag->data.data(), 4, 1, fin);
     } else {
       tag->value = gbfgetuint32(fin);
+      memcpy(tag->data.data(), &tag->value, 4);
```
robertlipe: I'd have to think about this really hard for big-endian machines, but I'm not sure we care about them much any more.

exif.cc
```diff
@@ -570,15 +544,15 @@ exif_read_app(exif_app_t* app)
...
   // Warning hush.
   (void) ifd;
 }
 
 static void
-exif_examine_app(exif_app_t* app)
+exif_examine_app(ExifApp* app)
 {
-  gbfile* ftmp = exif_app->fcache;
+  gbfile* ftmp = app->fcache;
```
robertlipe, line 681: So the original was looking at a global exif_app instead of the function argument? Wow.

exif.cc
```diff
@@ -620,14 +592,13 @@ exif_find_ifd(exif_app_t* app, const uint16_t ifd_nr)
...
+static ExifTag*
+exif_find_tag(ExifApp* app, const uint16_t ifd_nr, const uint16_t tag_id)
 {
-  exif_ifd_t* ifd = exif_find_ifd(app, ifd_nr);
+  ExifIfd* ifd = exif_find_ifd(app, ifd_nr);
   if (ifd != nullptr) {
-    queue* elem, *tmp;
-    QUEUE_FOR_EACH(&ifd->tags, elem, tmp) {
-      exif_tag_t* tag = reinterpret_cast<exif_tag_t *>(elem);
+    for (auto& tag_instance : ifd->tags) {
```
robertlipe, line 726: This could probably be a smarter type to avoid a linear scan, but you've already poked this pretty hard in this CL so that can wait another day.

exif.cc
```diff
@@ -844,15 +814,15 @@ exif_waypt_from_exif_app(exif_app_t* app)
   tag = exif_find_tag(app, EXIF_IFD, EXIF_IFD_TAG_USER_CMT); /* UserComment */
   if (tag && (tag->size > 8)) {
     QString str;
-    if (memcmp(tag->data, "ASCII\0\0\0", 8) == 0) {
-      wpt->notes = QString::fromLatin1((char*) tag->data + 8, tag->size - 8);
-    } else if (memcmp(tag->data, "UNICODE\0", 8) == 0) {
+    if (memcmp(tag->data.data(), "ASCII\0\0\0", 8) == 0) {
+      wpt->notes = QString::fromLatin1(tag->data.data() + 8, tag->size - 8);
+    } else if (memcmp(tag->data.data(), "UNICODE\0", 8) == 0) {
```
robertlipe: Both of these tests could probably be just plain ole operator ==. I'm not sure I care about testing trailing junk in a data structure that's both null terminated and fixed length. "ASCII\0\0377\07" is probably a legit match, no?

(I also totally get the mindset of faithfully converting existing code instead of rethinking it.)

exif.cc
```diff
@@ -969,41 +939,38 @@ exif_put_value(const int ifd_nr, const uint16_t tag_id, const uint16_t type, con
       return nullptr;
     }
 
-    tag = (exif_tag_t*) xcalloc(sizeof(*tag), 1);
+    ifd->tags.append(ExifTag());
```
robertlipe, line 1095: Would it be more natural to have a local that's populated and appended instead of one that's created and immediately stuffed into a queue and then pulled back out?

exif.cc
```diff
@@ -969,41 +939,38 @@ exif_put_value(const int ifd_nr, const uint16_t tag_id, const uint16_t type, con
...
       return nullptr;
     }
     if (tag->count < (index + count)) {
-      if (! tag->data_is_dynamic) {
-        void* tmp = xmalloc(tag->size < 4 ? 4 : tag->size);
-        memcpy(tmp, tag->data, tag->size);
-        tag->data = tmp;
-        tag->data_is_dynamic = 1;
-      }
+      auto* tmp = reinterpret_cast<char *>(xmalloc(tag->size < 4 ? 4 : tag->size));
```
robertlipe: This makes my head hurt. This whole mess makes me think that tag->data should be a std::vector<uint8_t> or something that's just resized as we go or something. Maybe that surgery is too radical for now...But QSharedPointers wrapping xmalloc/xfree seems like the wrong long-term solution.

exif.cc
```diff
@@ -1251,9 +1208,9 @@ exif_write_apps()
         exif_put_long(IFD1, IFD1_TAG_JPEG_OFFS, 0, len);
       }
 
-      QUEUE_FOR_EACH(&app->ifds, e1, t1) {
-        exif_ifd_t* ifd = reinterpret_cast<exif_ifd_t *>(e1);
-        sortqueue(&ifd->tags, exif_sort_tags_cb);
+      for (auto& ifd_instance : app->ifds) {
```
robertlipe, line 1439: Could this loop be combined with the one above or does GPS_IFD interfere with the sort? Thinking about readability here(if that's not a lost cause in this module :-) not the performance const of repeating it.)

exif.cc
```diff
@@ -1263,16 +1220,10 @@ exif_write_apps()
       gbfputuint16(0x2A, ftmp);
       gbfputuint32(0x08, ftmp);	/* offset to first IFD */
 
-      QUEUE_FOR_EACH(&app->ifds, e1, t1) {
-        exif_ifd_t* ifd = reinterpret_cast<exif_ifd_t *>(e1);
-        exif_ifd_t* ifd_next = reinterpret_cast<exif_ifd_t *>(t1);
-        char next;
+      for (int i=0; i<app->ifds.size(); ++i) {
```
robertlipe: Whitespace

robertlipe (commented):
I wonder if .data really should be a std::vector<uint8_t> or a QByteArray or something - anything - other than a char* that's reallocated and copied endlessly. Maybe it's an array of QVariants so you can get bools and doubles and chars in and out without endianness and alignment hassles.

But to the point of this CL, eradicating queue and using QList, I think that made this much more readable, so thanx!

exif.cc
tsteven4: This is a bug, and it manifests itself not based on the endianness of the machine, but based on the endianness of the image as stored in the TIFF header.  The big endian byte order isn't too common, but I found a sample image we can use to test.

exif.cc
tsteven4: These exact strings, including all the nulls, are specified in Table 9 and 35 of the EXIF specification.  Also, the user comment itself may not be null terminated, although the tag->data may be depending on how we implement it.  
(http://www.cipa.jp/std/documents/e/DC-008-Translation-2016-E.pdf)

tsteven4: There is a bug just below this, unicode can be UTF-16BE or UTF-16LE depending on file endianness.

exif.cc
tsteven4: It would be more natural to me as well, and I actually started down that path.  But in the end I went with less disruptive more literal translation.  Sometimes a pointer to a list element is returned and we need to get it back from the list anyway.

exif.cc
tsteven4: done.

exif.cc
tsteven4: QByteArray drops in for QSharedPointer<char> almost transparently.  There is still a ton of char manipulation through .data() or .constData().  QByteArray does clean up the messy replacement for xrealloc cited here though.

exif.cc
tsteven4: There is also a different preexisting bug here!

exif.cc
tsteven4: sorry, bigger fish to fry.

exif.cc
tsteven4: maybe, but it will have to wait.

tsteven4:
thanks for the review, there was one especially nasty old bug near the "ick" in exif_read_ifd.  It is fixed and tested now.  image endianess and machine endianess combine to create lots of opportunities for bugs that are hopefully all resolved now.

I replaced QSharePointer<char> with QByteArray.  It is a little better, especially when we need to resize on write, but there is still a ton of casting the underlying char* depending on the tag type.

GPSBabelDeveloper (approved):
Even with the grubbiness we've agreed to walk away from, this is *so much* more type-safe and generally readable, that it's a poster child fo such cleanup. Thanx for going the extra mile and improving test coverage and having the tenacity to dig up the specs and find that it really is a fixed-length AND zero terminated buffer, for example.

Thanx for a nice batch of fixes!

GPSBabelDeveloper:
BTW, for that bunch of casting you mention, I think that constructing a QVariant foo(theByteArray) and then using foo.toInt(), foo.toDouble(), foo.toLongLong, and so on is the approved way to avoid that.  I think that std::variant came in around C++11 or 17, but we can safely depend on QVariant which is battle-tested.

tsteven4:
Note that GPSBabel.pro was enabling assertions for release and debug CONFIG.  For MSVC the assertions got turned off for release builds during the translation process to a project file, but for other compilers assertions remained on.
On my linux system the default build is release, which is unfortunate in that assertions are turned off unless you do something extra, i.e. "qmake CONFIG+=debug".  It would be preferable to enable assertions unless you did something special, i.e. "qmake CONFIG+=release".

I have another version of exif.cc where tag->data becomes QVector<QVaraint>.  While this eliminates some casting it adds about 100 lines of code out of 1500.  There is an argument to be made that exif.cc doesn't need such detailed knowledge of most tags, it just needs to preserve them from read to write.  I think the QVariant version is less tolerant of unkown tag types as well.

tsteven4:
One problem with the QByteArray implementation was that alignment for different types is not guaranteed.  So I bit the bullet and went for QVector<QVariant>.   QByteArray is only used for byte data now.

PR #239, opened 2018-07-21
eliminate queues in inifile.

tsteven4:
check out the changes to inifile_done.

### DIFF
--- a/inifile.cc
+++ b/inifile.cc
@@ -34,7 +34,9 @@
 
 #define MYNAME "inifile"
 
-struct InifileSection {
+class InifileSection
+{
+public:
   QString name;
   QHash<QString, QString> entries;
 
--- a/inifile.cc
+++ b/inifile.cc
@@ -18,25 +18,28 @@
     Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111 USA
 */
 
-#include "defs.h"
+#include "defs.h"              // for fatal, ugetenv, warning
 #include "inifile.h"
-#include <QtCore/QDir>
-#include <QtCore/QFile>
-#include <QtCore/QList>
-#include <QtCore/QSharedPointer>
-#include <cstdio>
-#include <cstdlib>
+#include "src/core/file.h"     // for File
+#include <QtCore/QByteArray>   // for QByteArray
+#include <QtCore/QChar>        // for operator==, QChar
+#include <QtCore/QDir>         // for QDir
+#include <QtCore/QFile>        // for QFile
+#include <QtCore/QFileInfo>    // for QFileInfo
+#include <QtCore/QHash>        // for QHash
+#include <QtCore/QIODevice>    // for QIODevice::ReadOnly, QIODevice
+#include <QtCore/QTextStream>  // for QTextStream
+#include <QtCore/Qt>           // for CaseInsensitive
+#include <QtCore/QtGlobal>     // for qPrintable
 
 #define MYNAME "inifile"
 
-struct inifile_entry_t {
-  QSharedPointer<char> key;
-  QSharedPointer<char> val;
-};
+struct InifileSection {
+  QString name;
+  QHash<QString, QString> entries;
 
-struct inifile_section_t {
-  QSharedPointer<char> name;
-  QList<inifile_entry_t> entries;
+  InifileSection() = default;
+  explicit InifileSection(QString nm) : name{nm} {}
 };
 
 /* internal procedures */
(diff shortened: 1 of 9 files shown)


inifile.cc
```diff
@@ -22,29 +22,25 @@
...
 #include <cstdlib>
 
 #define MYNAME "inifile"
 
-typedef struct inifile_entry_s {
-  queue Q;
-  char* key;
-  char* val;
-} inifile_entry_t;
+struct inifile_entry_t {
```
robertlipe: Isn't this basically now a QMap or QHash? If preserving the case insensitivity matters, we can override operator==  with a QString::compare(...Qt::CaseInsensitive)

inifile.cc
```diff
@@ -109,7 +105,7 @@ inifile_load_file(gbfile* fin, inifile_t* inifile, const char* myname)
     char* cin = lrtrim(buf);
```
robertlipe: If you just make cin a QString(buf).trimmed(), I think we save some xstrdup chatter below.

inifile.cc
```diff
@@ -132,35 +128,31 @@ inifile_load_file(gbfile* fin, inifile_t* inifile, const char* myname)
...
     } else {
       if (sec == nullptr) {
         fatal("%s: missing section header in '%s'.\n", myname,
               qPrintable(gbinipathname));
       }
 
-      inifile_entry_t* entry = (inifile_entry_t*) xcalloc(1, sizeof(*entry));
-      ENQUEUE_TAIL(&sec->entries, &entry->Q);
-      sec->ientries++;
+      sec->entries.append(inifile_entry_t());
```
robertlipe: Would it be more natural to just construct entry, use it, and append it instead of constructing it, shoving it into the queue, and then immediately pulling the last one back out?

inifile.cc
```diff
@@ -169,23 +161,18 @@ inifile_load_file(gbfile* fin, inifile_t* inifile, const char* myname)
 static char*
 inifile_find_value(const inifile_t* inifile, const char* sec_name, const char* key)
```
robertlipe: Ideally, the meat of this becomes a entry[val].data()  lookup. Maybe it's a two-stage thing to deal with each section, but maybe the internal storage becomes "Section:Entry" so you glom section and name together and can do a single lookup, at the cost of not being able to have a colon (or whatever) in the section or key names. But I'd bed this whole function could be a few lines long and exhibit way better than O(sections*N) lookups and stores. (Duly noting that readability and not performance is my actual point...)

inifile.cc
```diff
@@ -132,35 +128,31 @@ inifile_load_file(gbfile* fin, inifile_t* inifile, const char* myname)
...
-      entry->key = xstrdup(cin);
+      entry->key.reset(xstrdup(cin), xfree);
 
       if (cx != nullptr) {
         cx = lrtrim(++cx);
-        entry->val = xstrdup(cx);
+        entry->val.reset(xstrdup(cx), xfree);
       } else {
-        entry->val = xstrdup("");
+        entry->val.reset(xstrdup(""), xfree);
```
robertlipe: That goes to a gross length to make entry->val = QString() - and whatever code looking at this that's prepared to make the distinction of an empty string and a null string pointer probably becamse accidentally smarter when we moved to QString anyway.

inifile.h
```diff
@@ -22,12 +22,13 @@
...
 
-typedef struct inifile_s {
-  int isecs;			/* number of sections */
-  queue secs;			/* sections */
-  uint8_t unicode:1;
-} inifile_t;
+struct inifile_section_t;
+struct inifile_t {
+  QList<inifile_section_t> secs;			/* sections */
+  bool unicode{false};
```
robertlipe: Is .unicode ever actually used anywhere meaningfully here? It seems to be vestigal calls to cet_convert()

robertlipe (commented):
This is so much easier on the eyes...Thanx!

inifile.cc
tsteven4: I think this is all in char* land still, note that 
char* xstrdup(const char* s);
 is in defs.h, as is
char* xstrdup(const QString& s);

So we set up a non-null QSharedPointer to point to a char string that is null, i.e. it contains only '\0'.

robertlipe:
There were certainly overloads added to unblock NEW_STRING so we didn't
have to conditionalize another bazillion things.  Let's not get wrapped up
pulling at details. If this isn't easy, walk away for now. The last version
was groovy with me. Go for it. Your judgement is good with me.

inifile.h
tsteven4: In all use cases the unicode flag seems superfluous.

tsteven4:
updated:
1. QHash used for key/values, but not for sections.  Keys and lookups forced to lower case to implement case insensitive matching.
2. If a key isn't found you get a null QString, If a key is found you get a non-null but possibly empty value.  This should preserve historical behavior.
3. QTextStream used to read the file.

tsteven4:
I am a little nervous about implementing case insensitive matching by forcing to lower case.   toLower() is done in the C locale, so I would expect  converting À to lower may not give à.

My attempt to override operator== didn't work on the first lookup, but all the subsequent lookups worked.  I haven't figured that out.

PR #246, opened 2018-08-30
introduce optional class

tsteven4:
for things that may or may not have a value.
demonstrate usage with computed_trkdata.


GPSBabelDeveloper (approved):
I like this a lot. I've thought about this off and on for years - it was one of the areas that made me move to C++.

Thanx for tackling it!

PR #255, opened 2018-10-09
Parallelize CLI and GUI make.

GPSBabelDeveloper:
Try to knock those 30 minute build times down a little.
 testo and vtesto are harder


GPSBabelDeveloper:
1:22 to 1:06. Not a bad start knowing that the valgrind tests are the long road. Well, building the doc four times is a bit silly, too.

PR #259, opened 2018-10-15
eliminate queues in csv style handling.

tsteven4:
also:
fix bug with style option encoding.
pass some parameters by const reference.
eliminate some unused time fiddling.

Note that the field maps stores key, value and printf data as QByteArrays.  This allows xcsv_parse_val() and xcsv_waypt_pr() to continue to use character strings via constData(), which should be quite efficient.  It simplifies memory allocation/deallocation by reducing use of xstrdup.  It avoids repeated conversion with fromUtf8() that is required to get from QStrings to character strings.  I didn't have the fortitude to convert these functions to use QStrings.  see csv_util.h for some comments on this.

### DIFF
--- a/csv_util.cc
+++ b/csv_util.cc
@@ -20,9 +20,11 @@
 
  */
 
-#include <QtCore/QRegExp>
-#include <QtCore/QTextStream>
-#include <QtCore/QDebug>
+#include <QtCore/QChar>            // for QChar
+#include <QtCore/QDateTime>        // for QDateTime, QDate
+#include <QtCore/QString>          // for QString, QCharRef
+#include <QtCore/QTextStream>      // for QTextStream
+#include <QtCore/QtGlobal>         // for foreach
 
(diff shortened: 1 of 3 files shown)


GPSBabelDeveloper:
Awesome. Thanx. I have a tree like this that I'm happy to retire. This is very much what I had in mind as the ending point for my recent work in this area. 

This is definitely one of those places in the code where the C version was a forced fit and having access to grown-up containers is so much nicer.

I really wonder if our excessively clever hashed_key and in the full gpx tag token stuff is really a win with these in place. (That's a good starter project for someone that's not steven in case we have any readers...)

It's my fantasy for xstrdup() to go away; we've used it as a pain killer in a lot of places. We've already removed xmalloc from most of our interesting modules; most of the places that remain are either low value or real head-screw code.

GPSBabelDeveloper (approved):

PR #264, opened 2018-11-01
support kml 2.3 track element.

tsteven4:
Also relax ordering requirement for gx:Track kml:when and
gx:coord children.  This allows gx:Track elements that violate
the schema as far as the sequencing of coord and when
children to be processed.  Note that the number of coord and when
elements still must be equal.

### DIFF
--- a/kml.cc
+++ b/kml.cc
@@ -35,6 +35,7 @@
 #include <cmath>
 #include <cstdio>
 #include <cstdlib>
+#include <tuple>
 
 // options
 static char* opt_deficon = nullptr;
@@ -73,6 +74,7 @@ static QString posnfilenametmp;
 
 static route_head* gx_trk_head;
 static QList<gpsbabel::DateTime>* gx_trk_times;
+static QList<std::tuple<int, double, double, double>>* gx_trk_coords;
 
 static gpsbabel::File* oqfile;
 static gpsbabel::XmlStreamWriter* writer;
@@ -276,6 +278,14 @@ xg_tag_mapping kml_map[] = {
   { gx_trk_e,  	cb_end, 	"/Placemark/*gx:Track" },
   { gx_trk_when,  cb_cdata, "/Placemark/*gx:Track/when" },
   { gx_trk_coord, cb_cdata, "/Placemark/*gx:Track/gx:coord" },
+  { gx_trk_s,  	cb_start, 	"/Placemark/Track" }, // KML 2.3
+  { gx_trk_e,  	cb_end, 	"/Placemark/Track" }, // KML 2.3
+  { gx_trk_when,  cb_cdata, "/Placemark/Track/when" }, // KML 2.3
+  { gx_trk_coord, cb_cdata, "/Placemark/Track/coord" }, // KML 2.3
+  { gx_trk_s,  	cb_start, 	"/Placemark/MultiTrack/Track" }, // KML 2.3
+  { gx_trk_e,  	cb_end, 	"/Placemark/MultiTrack/Track" }, // KML 2.3
+  { gx_trk_when,  cb_cdata, "/Placemark/MultiTrack/Track/when" }, // KML 2.3
+  { gx_trk_coord, cb_cdata, "/Placemark/MultiTrack/Track/coord" }, // KML 2.3
   { nullptr,	(xg_cb_type) 0, 		nullptr }
 };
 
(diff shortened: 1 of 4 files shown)


kml.cc
```diff
@@ -448,15 +458,51 @@ void gx_trk_s(xg_string, const QXmlStreamAttributes*)
...
+  delete gx_trk_coords;
+  gx_trk_coords = new QList<std::tuple<int, double, double, double>>;
 }
 
 void gx_trk_e(xg_string, const QXmlStreamAttributes*)
 {
+  // Check that for every temporal value (kml:when) in a kml:Track there is a position (kml:coord) value.
+  // Check that for every temporal value (kml:when) in a gx:Track there is a position (gx:coord) value.
+  if (gx_trk_times->size() != gx_trk_coords->size()) {
+    fatal(MYNAME ": There were more coord elements than the number of when elements.\n");
```
GPSBabelDeveloper, line 478: Optional: We should probably display the numbers we're fussing about here.

kml.cc
```diff
@@ -73,6 +74,7 @@ static QString posnfilenametmp;
 
 static route_head* gx_trk_head;
 static QList<gpsbabel::DateTime>* gx_trk_times;
+static QList<std::tuple<int, double, double, double>>* gx_trk_coords;
```
GPSBabelDeveloper, line 85: Interesting to choose tuple over struct. Makes sense in this case, but it's awkward that we have implied ordering in a few places.

kml.cc
```diff
@@ -448,15 +458,51 @@ void gx_trk_s(xg_string, const QXmlStreamAttributes*)
...
 
 void gx_trk_e(xg_string, const QXmlStreamAttributes*)
 {
+  // Check that for every temporal value (kml:when) in a kml:Track there is a position (kml:coord) value.
+  // Check that for every temporal value (kml:when) in a gx:Track there is a position (gx:coord) value.
+  if (gx_trk_times->size() != gx_trk_coords->size()) {
+    fatal(MYNAME ": There were more coord elements than the number of when elements.\n");
+  }
+
+  // In KML 2.3 kml:Track elements kml:coord and kml:when elements are not required to be in any order.
```
GPSBabelDeveloper, line 481: As an aside, I don't think that was as much as conscious decision as much as it was difficult to explain it in the schema tools used.

kml.cc
```diff
@@ -448,15 +458,51 @@ void gx_trk_s(xg_string, const QXmlStreamAttributes*)
...
+    double lat, lon, alt;
+    int n;
+    std::tie(n, lat, lon, alt) = gx_trk_coords->takeFirst();
+    // An empty kml:coord element is permitted to indicate missing position data;
+    // the estimated position may be determined using some interpolation method.
+    // However if we get one we will throw away the time as we don't have a location.
+    // It is not clear that coord elements without altitude are allowed, but our
+    // writer produces them.
+    if (n >= 2) {
+    trkpt->latitude = lat;
```
GPSBabelDeveloper: Indention?

kml.cc
```diff
@@ -448,15 +458,51 @@ void gx_trk_s(xg_string, const QXmlStreamAttributes*)
...
+  while (!gx_trk_times->isEmpty()) {
+    Waypoint* trkpt = new Waypoint;
+    trkpt->SetCreationTime(gx_trk_times->takeFirst());
+    double lat, lon, alt;
+    int n;
+    std::tie(n, lat, lon, alt) = gx_trk_coords->takeFirst();
+    // An empty kml:coord element is permitted to indicate missing position data;
+    // the estimated position may be determined using some interpolation method.
+    // However if we get one we will throw away the time as we don't have a location.
+    // It is not clear that coord elements without altitude are allowed, but our
```
GPSBabelDeveloper, line 493: That's the KML equivalent of (and at some level, the inspiration for) our "unknown altitude" values. In Earth, it mostly means 'clamped to ground' or 'draped over terrain' to stop things like boat rides dipping below water when the number is likely to be close to zero vs. being legitimately unknown. We (GPSBabel) don't handle this consistently.

GPSBabelDeveloper (approved):
This is a really nice update. Thanx for tackling it. No changes necessary, just taking notes.

(Not that I think anybody but you and I read these...)

kml.cc
tsteven4: In general \<sequence> was changed to \<all> in the schema, moving the burden of ordering onto the reader.

KML 2.3 seems DOA, despite google prominently appearing among the submitters of the 2.3 OGC standard, and 3 years passing, GE is still using kml 2.2 and google extensions (xmlns:gx="http://www.google.com/kml/ext/2.2).  It was very hard to find any 2.3 samples.

PR #266, opened 2018-11-13
Incorporate improvements found by Qt oriented code checker clazy

tsteven4:
https://github.com/KDE/clazy tag v1.4 was used to build clazy.

Fixes are manual except for clazy-qstring-ref.

When a for loop was flagged, I followed the recommendations on
http://doc.qt.io/qt-5/qtglobal.html#Q_FOREACH
regarding qforeach vs. c++11 range-for.
I note that these recommendations seem to be a bit different than those on
http://doc.qt.io/qt-5/containers.html#the-foreach-keyword

### DIFF
--- a/bcr.cc
+++ b/bcr.cc
@@ -447,9 +447,9 @@ bcr_route_header(const route_head* route)
     }
 
     if (sout.isEmpty()) {
-      sout = QString("%1,%2,@,0").arg(s1).arg(s1);
+      sout = QString("%1,%2,@,0").arg(s1, s1);
     } else {
-      sout = QString("%1,%2,@,0").arg(s1).arg(s2);
+      sout = QString("%1,%2,@,0").arg(s1, s2);
     }
 
     bcr_write_line(fout, "STATION", &i, sout);
--- a/csv_util.cc
+++ b/csv_util.cc
@@ -1593,7 +1593,7 @@ xcsv_waypt_pr(const Waypoint* wpt)
   }
 
   int i = 0;
-  for (const auto& fmp : xcsv_file.ofields) {
+  for (const auto& fmp : qAsConst(xcsv_file.ofields)) {
     double lat = latitude;
     double lon = longitude;
     /*
@@ -2183,7 +2183,7 @@ xcsv_data_write(void)
   waypt_out_count = 0;
 
   /* output prologue lines, if any. */
-  for (const auto& line : xcsv_file.prologue) {
+  for (const auto& line : qAsConst(xcsv_file.prologue)) {
    QString line_to_write = xcsv_replace_tokens(line);
     *xcsv_file.stream << line_to_write <<  xcsv_file.record_delimiter;
   }
@@ -2199,7 +2199,7 @@ xcsv_data_write(void)
   }
 
   /* output epilogue lines, if any. */
-  for (const auto& line : xcsv_file.epilogue) {
+  for (const auto& line : qAsConst(xcsv_file.epilogue)) {
     QString line_to_write = xcsv_replace_tokens(line);
     *xcsv_file.stream << line_to_write << xcsv_file.record_delimiter;
   }
--- a/csv_util.cc
+++ b/csv_util.cc
@@ -894,7 +894,6 @@ writetime(const char* format, time_t t, bool gmt)
 
   char tbuff[1024];
   strftime(tbuff, sizeof tbuff, format, stmp);
-  QDateTime dt = QDateTime::fromTime_t(t);
   return QString(tbuff);
 }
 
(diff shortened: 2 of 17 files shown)


bcr.cc
```diff
@@ -447,9 +447,9 @@ bcr_route_header(const route_head* route)
     }
 
     if (sout.isEmpty()) {
-      sout = QString("%1,%2,@,0").arg(s1).arg(s1);
+      sout = QString("%1,%2,@,0").arg(s1, s1);
```
GPSBabelDeveloper, line 450: That's a weird way to write this code anyway. But BCR isn't a format I worry much about.

csv_util.cc
```diff
@@ -1421,7 +1420,7 @@ xcsv_data_read(void)
      * pre-read the file to know how many data lines we should be seeing,
      * we take this cheap shot at the data and cross our fingers.
      */
-    foreach(const QString& ogp, xcsv_file.epilogue) {
+    for(const auto& ogp : qAsConst(xcsv_file.epilogue)) {
```
GPSBabelDeveloper, line 1423: In general, we're going to have a lot of noise here because Qt's foreach predates C11's range for and C17's const foreach iterators that semi-automatically create a thread pool and parallelize.

gpx.cc
```diff
@@ -20,20 +20,28 @@
...
-#include <QtCore/QRegExp>
-#include <QtCore/QXmlStreamReader>
+#include <QtCore/QDateTime>            // for QDateTime, QDate, QTime
+#include <QtCore/QHash>                // for QHash
+#include <QtCore/QList>                // for QList
+#include <QtCore/QString>              // for QString, QStringLiteral, QStringRef, QStaticStringData, QLatin1String, operator+
+#include <QtCore/QStringList>          // for QStringList
+#include <QtCore/QXmlStreamAttributes> // for QXmlStreamAttributes
+#include <QtCore/QXmlStreamNamespaceDeclarations> // for QXmlStreamNamespaceDeclarations
+#include <QtCore/QXmlStreamReader>     // for QXmlStreamReader, QXmlStreamReader::TokenType::Characters, QXmlStreamReader::TokenType::EndDocument, QXmlStreamReader::TokenType::EndElement, QXmlStreamReader::TokenType::Invalid, QXmlStreamReader::TokenType::StartElement
```
GPSBabelDeveloper, line 38: You must have some tool that's better at this than what I'm using. This is great if it's auto-generated.

kml.cc
```diff
@@ -792,33 +792,33 @@ void kml_output_trkdescription(const route_head* header, const computed_trkdata*
   }
   const char* distance_units;
   double distance = fmt_distance(td->distance_meters, &distance_units);
-  kml_td(hwriter, QStringLiteral("Distance"), QStringLiteral(" %1 %2 ").arg(QString::number(distance, 'f', 1)).arg(distance_units));
+  kml_td(hwriter, QStringLiteral("Distance"), QStringLiteral(" %1 %2 ").arg(QString::number(distance, 'f', 1), distance_units));
```
GPSBabelDeveloper, line 795: The original guidance for localizataion was to have user-visible generated strings as .arg so that translators could reorder them if needed. We've generally made such a mess of that this is probably not helpful now.

mmo.cc
```diff
@@ -259,9 +260,9 @@ mmo_register_object(const int objid, const void* ptr, const gpsdata_type type)
 static int
 mmo_get_objid(const void* ptr)
 {
-  foreach(int key, objects.keys()) {
-    if (objects.value(key)->data == ptr) {
-      return key;
+  for (auto o = objects.constBegin(); o != objects.constEnd(); ++o) {
```
GPSBabelDeveloper, line 263: This code has probably just been mechanically converted a few too many times. Is this:
if objects[ptr].has_value() return object[ptr]
This looks like some of Olaf's home grown map code.

GPSBabelDeveloper (approved):
Good mod. thanx! I've been meaning to point clazy and our code and gui for a while. I'm surprised at how little it found, but we use pretty little (and pretty conservative) Qt in our main app.

gpx.cc
tsteven4: include-what-you-use is the tool, with mapping files for Qt 5.9 that I generated.  But it is imperfect, so I have been using it in a piecemeal fashion and massaging the results by hand.  If you want to generate Qt mapping files I have a script or two to help automate that step.

mmo.cc
tsteven4: Your translation doesn't look good to me.

Codacy shows we do have coverage on those lines.

If this web site is related to this format, then I wonder about the usefulness of us supporting it.
http://memory-map.com/mm-PC-help2/index.htm#t=Maptech_io.htm
shows they support csv, gpx, shape files(with professional license), and loc files besides their proprietary format.

robertlipe:
Doing it from a fone was risky. It just caught my eye that sequentially
iterating over a container looking for a key wasn't very awesome.

Other than an open issue in github, I don't see evidence of this format in
mail so just making it disappear (or certainly not spending time on it)
seems wise.

Do as you see fit.

Issue #248, opened 2018-09-11
error parsing Lowrance USR version 4 waypoints


bjkowalski:
Using version 1.5.4 (both on Windows and Linux) attempts to use USR4 file as input results in error "Process exited unsuccessfully with code 1".

![image](https://user-images.githubusercontent.com/42647415/45365158-81645b00-b5a1-11e8-9261-27b05d1b61b2.png)

![image](https://user-images.githubusercontent.com/42647415/45365171-8c1ef000-b5a1-11e8-8be2-94d2933d159a.png)

PR #268, opened 2018-11-14
Added support for USR version 5 and 6 and corrected USR version 4 support.

bjkowalski:
Eliminated separate processing for USR4 to simplify user interface.  Now processing detects the USR version and does the appropriate processing.  Newer Lowrance units generate different USR2 than what was supported.  Corrected old USR4 processing and added support for USR5 and USR6. 

Added tests for USR4, USR5, and USR6 and the ability to compare results between them in some cases.
Renamed test files to follow a lowrance-*.* naming scheme.  Significant documentation updates.


tsteven4:
Can you please resolve the merge conflict so CI can run.  Thanks.

Are the changes to garmin_fit.cc, strptime.c and zip.c extraneous to the purpose of this PR?

bjkowalski:
The changes to garmin_fit.cc, strptime.c and zip.c were to address compiler warnings.  I should have submitted those separately.

tsteven4:
Thanks for this unifying work, and even figuring out our doc system.  Here are some requests that are peripheral to your PR.  I still need to review the meat in lowranceusr.cc

please back out strptime.c, garmin_fit.cc, zip.c.  Let's consider those separately.  There is a higher bar for library changes like zip.c as we don't really want to own them.  In the case of MINIZIP it isn't even built for builds other than config (.e.g GPSBabel.pro), and we have no test coverage.  I think MINIZIP was work  @gpsbabel was doing that never went live.

Please update CMakeLists.txt re lowranceusr4.cc.  We don't often use cmake, but it does help with CLion.

If we are going to improve precision which requires regenerating a bunch of reference files lets use
#define DEGREESTORADIANS (M_PI/180.0)

#define SECSTO2000 is unused, kill it.

does your PR resolve #248?  If so indicate this as described https://help.github.com/articles/closing-issues-using-keywords and the issue will get closed when this PR merges.

tsteven4:
Another nit.  We try to enforce a consistent style using astyle with the supplied style file astylerc.
astyle --options=astylerc lowranceusr.cc

lowranceusr.cc
```diff
@@ -1,238 +1,474 @@
...
+      format data.
+    - Combined processing of USR 2/3/4/5/6 formats into a single
+      file. The previous implementation did not support conversion
+      between USR 2/3 and 4/5/6 or between 4/5/6 and 2/3 beause of 
+      the seperation in processing between lowranceusr.cc and 
+      lowranceusr4.cc.
+
+  USR Background Information
+
+  Collected from various WEB sources and Lowrance HOOK2 User Manual.
```
tsteven4, line 46: For maintenance it is very nice to have the sources of information available.  Can you list the URLs?

lowranceusr.cc
```diff
@@ -489,167 +1230,320 @@ lowranceusr_parse_icons()
...
+      }
       waypt_add(wpt_tmp);
     }
   }
 
 }
 
 static void
-lowranceusr_parse_trails()
+lowranceusr_parse_trail(int *trail_num)
```
tsteven4: There appears to be no intent to modify the callers value of trail num.  Pass as int?

lowranceusr.cc
```diff
@@ -489,167 +1230,320 @@ lowranceusr_parse_icons()
...
+      }
     }
+  } else {
+    /* remove the trail since it's empty */
+    track_del_head(trk_head);
+  }
+}
+
+static void
+lowranceusr4_parse_trail(int *trail_num)
```
tsteven4: There appears to be no intent to modify the callers value of trail num. Pass as int?

lowranceusr.cc
```diff
@@ -1,238 +1,474 @@
...
+  Collected from various WEB sources and Lowrance HOOK2 User Manual.
+
+    USRv3 supported trails with a maximum of 10,000 trail-points.
+    USRv4 and above support a maximum of 20,000 trail-points(actually 24K and change).
+    USRv4 and above & GPX support trails with trail-segments.
+    USRv6 supports trail characteristics speed & temperature.
+
+    USR Data File Description from Lowrance Manual
+    ----------------------------------------------
+      User Data File version 6 - USRv6
```
tsteven4, line 54: copyright issues?

testo.d/lowranceusr.test
```diff
@@ -1,36 +1,150 @@
...
-# awkward  to test.
+# Lowrance USR
+# ============
+# 
+# Binary data format, and slightly lossy because of the math to
+# convert lat/long.
+#
+# File Name		  	File Contents
+# ===============		========================
+# lowrance.usr			USR version 2 format file.  Generated by GPSBabel from geocaching.loc input..
```
tsteven4, line 10: Its very nice to document where reference files came from, and which are golden, i.e. directly from the device.  They way we store reference files and translations of reference files in the same directory is confusing at best.

testo.d/lowranceusr.test
```diff
@@ -1,36 +1,150 @@
...
+#
+# Perform USR version 5 test
+#         -------------
+#
+gpsbabel -i lowranceusr -f ${REFERENCE}/lowrance-v5.usr -o gpx -F ${TMPDIR}/lowrance-v5.gpx
+compare ${REFERENCE}/lowrance-v5.gpx ${TMPDIR}/lowrance-v5.gpx
+
+# Should have generated the same GPX file as USR version 4 but the order is different for some reason
+# so use another way to compare to ensure have EXACTLY the same data
+#
```
tsteven4, line 129: Interesting.  Maybe we shouldn't be storing rte_head->rte_num values at all if the order we retrieve the routes/trails doesn't mean anything.

If didn't store rte_num values then I have enhancements to the sort filter that could resort the routes & tracks by name, description or number.  The sort filter can already sort the waypoints by time, etc.

lowranceusr.cc
tsteven4: my bad, break trail into segments modifies trail_num.

lowranceusr.cc
```diff
@@ -489,167 +1230,320 @@ lowranceusr_parse_icons()
...
-      /* position coord lat & long */
-      (void) gbfread(&buff[0], 4, 2, file_in);
-      /* symbol */
-      (void) gbfread(&buff[0], 4, 1, file_in);
-    } else {
+    double latitude    = lat_mm_to_deg(gbfgetint32(file_in));
+    double longitude   = lon_mm_to_deg(gbfgetint32(file_in));
+    int    icon_number = gbfgetint32(file_in);
+
+    if (opt_ignoreicons == NULL) {          /* Option not specified if NULL */
```
tsteven4: nullptr instead of NULL.

lowranceusr.cc
```diff
@@ -894,18 +1884,75 @@ lowranceusr_route_hdr(const route_head* rte)
...
+{
+  if (global_opts.debug_level >= 1) {
+    printf(MYNAME " writing route #%d (%s) with %d waypts\n",
+           route_uid, qPrintable(rte->rte_name), rte->rte_waypt_ct);
+  }
 
+  /* UID unit number */
+  if (opt_serialnum_i > 0)
+    gbfputint32(opt_serialnum_i, file_out);   // use option serial number if specified
+  else if (rte->fs != (format_specific_data *)NULL)
```
tsteven4: nullptr instead of "(format_specific_data *) NULL" ?

lowranceusr.cc
```diff
@@ -758,24 +1652,120 @@ lowranceusr_waypt_disp(const Waypoint* wpt)
...
+    printf("\n");
 }
 
+static void
+lowranceusr4_waypt_disp(const Waypoint* wpt)
+{
+  /* UID unit number */
+  if (opt_serialnum_i > 0)
+    gbfputint32(opt_serialnum_i, file_out);   // use option serial number if specified
+  else if (wpt->fs != (format_specific_data *)NULL)
```
tsteven4: nullptr instead of "(format_specific_data *) NULL" ?

lowranceusr.cc
```diff
@@ -894,18 +1884,75 @@ lowranceusr_route_hdr(const route_head* rte)
...
+        printf(MYNAME " wrote route leg with waypt '%s'\n", qPrintable(wpt->shortname));
+      }
+      break;
+    }
+  }
+}
+
+
+static void
+lowranceusr4_route_trl(const route_head* rte)
```
tsteven4: "const route_head* rte" -> "const route_head* /* rte */" or "const route head*" will silence warnings about unused parameter.

bjkowalski:
Closes #248

lowranceusr.cc
bjkowalski: Left this in for consistency with lowranceusr_parse_trail().  Like lowranceusr_parse_trail() this may need to update if I find that trails need to be broken into segments.

lowranceusr.cc
bjkowalski: I wondered about that as well, hence my reference to the source.  This is also available electronically and will include the URL per your earlier suggestion.  I've included similar text in the doc updates.

lowranceusr.cc
bjkowalski: Changed all NULL references to nullptr.
Thanks.

lowranceusr.cc
bjkowalski: Adopted change.

lowranceusr.cc
bjkowalski: Adopted suggestion.

lowranceusr.cc
bjkowalski: Added URLs per request.

lowranceusr.cc
bjkowalski: Thanks for the info.

lowranceusr.cc
```diff
@@ -1,238 +1,478 @@
...
+static short          num_section_points;
+static char*          merge;
+static int            reading_version;
+static int            writing_version;
 
 #define MYNAME "Lowrance USR"
 
-#define MAXUSRSTRINGSIZE	256
-#define SEMIMINOR		   6356752.3142
-#define DEGREESTORADIANS	0.017453292
```
tsteven4, line 220: I would still like to see this:
#define DEGREESTORADIANS (M_PI/180.0)
If not defined elsewhere defs.h will supply M_PI.
The compiler should do the division at compile time.
This will reduce the chance of reference file thrashing in the future as there isn't the temptation to add another digit.

bjkowalski: I was planning to make the change but I need to be able to spend enough time so I can make the change, regenerate the reference files, and then desk check to make sure I didn't mess anything up.  I am hoping to get that done in the next few days.

bjkowalski: Suggested changed adopted.

testo.d/lowranceusr.test
bjkowalski: Being able to sort would definitely simplify the verification tasks, and I could see where there would be advantages when moving from one vendor to another.  Right now there is so much loss of Waypoint, Trail and Route information moving from USR to GPX, or even USRx to USRy, that the change would seem to only benefit the validation process.  I'd definitely support adding the option but without significant support from Lowrance on decoding the USR formats it wouldn't be something that many folks would use at least for sorting data (which is not an option they have on their units).

tsteven4: I merged #273.  I think it is of general usefulness, and if you want to use it for verification of lowranceusr it's available.  However, what you have works fine, and there is not need to change it.
For example
gpsbabel -i lowrancusr -f ${REFRENCE}/lowrance-v5.usr -x sort,rtename -o gpx -F lowrance-v5.gpx

However, because the route number (rte_num) values reflect the order in the usr file, while the route names don't, the gpx files still won't match between usr versions.

lowranceusr.cc
tsteven4: Please remove any and all text from copyrighted sources that we don't have permission to use.  This applies to all files, including source and documentation files.  You may substitute a phrase like:
"The recipe for green eggs and ham may be found in the book "Green Eggs and Ham" by Dr. Seuss, which is available at
http://www.seussville.com/books/book_detail.php?isbn=9780394800165

In the document you can use ulink elements, for an example see xmldoc/formats/kml.xml

bjkowalski: Replaced copyrighted material with authored comments.

robertlipe:
I've not scoured this in detail as it seems to be proceeding well without
me - which makes me very happy. :-)

GPX and GPSBabel's original internal data structures are both strongly
influenced by handheld GPS receivers of the previous century. Both the
Magellan and Garmins of that era used an integer for a "route number", but
it was really never well defined other than inside the unit itself. If you
copied from one device to another, and had overlapping route numbers, were
they guaranteed to be unique? (I already had a route "3" and copied in your
route with a number of "3" - is that OK? /shrug  - nobody's complained
much...)  Magellan's serial protocol (including their devices that used
their serial protocol stuck in files like Explorist x00 and Meridian) is
pretty dead these days. It looks like it's been dropped from the more
recent Garmin route transfer protocols. In short, I don't think we should
think too much about it between formats and just try to do something same
when going back to the same device, such as for backup and restore across a
reset.

minizip was needed for a format I started and then hit a stone wall with.
Garmin's GGZ format is a zip file that contains something like a .loc that
contains file/byte offsets into gpx files.  Worked great when everything
was ASCII, but our Qt helpers really really wanted to return character
offsets and not byte offsets so the offsets were bogus and crashed the
units. I also screwed up the implementation in a way that broke GPX and the
geocaching world pretty much lost interest in GGZ files when we realized
that they weren't that much more convenient than GPX files anyway in
practice.

format_specific_data could probably someday be a subclass of Waypoint. I
just really haven't tried in practice to see how well that would work out.
If you could dynamic case a Waypoint to a GarminWaypoint  (whatever) you
could probably carry that around, add it to lists, etc. Now what happens if
you have fitness data + temperature data + Lowrance data and you try to
dycast that a GarminWaypoint? There's probably some actual design work
there to do.

Carry on!  I'm back to bed...

tsteven4:
The test to see if we can write back the same data can be enabled with minor changes as shown in the following patch.  This patch uses rounding instead of truncation when converting back to the integer Lowrance Mercator meter format for latitudes and longitudes, which would seem to be desirable in any case.  However, this does ripple into other reference files.  

[p.txt](https://github.com/gpsbabel/gpsbabel/files/2637377/p.txt)

tsteven4:
unfortunately coverage reporting isn't working with pull requests, but running it locally shows the coverage is 62%.  This could be greatly improved by a test where we write a version 4 usr file.

Generation of local html coverage reports will be coming in #281.

tsteven4:
There is at least one problem with the >=v4 writer:
./gpsbabel -i geo -f geocaching.loc -o lowranceusr,wversion=4 -F x.usr
Lowrance USR: Could not write 4294893579 bytes to x.usr (result 73716)!

bjkowalski:
I happened to try some USR4 output as well this weekend, ran into the same
problem and also discovered the ICONs were not translating properly.  I ha
e a few USR4 output changes and will look at the other suggested changed..

Thanks..

tsteven4:
I noticed that the day increments if we read version 4 and write it out.

I have some code I can pass along (via a PR to bjkowalsk/gpsbabel ?) that

1. avoids unnecessary conversions between time_t and gpsbabel::DateTime or QDateTime
2. uses Qt for julian day conversions
These solve the incrementing day problem, but I am suspicious of our interpretation of time:

>   /* Creation date/time; the date is a Julian day number, and the
     time is a unix timestamp. */
  uint create_date = gbfgetint32(file_in);
  uint create_time = gbfgetint32(file_in);

The JD interpretation seems to be consistent with the date string in the beginning of the newer usr files, followed by a 32 bit JD, although the date string sometimes seems to be MM/dd/yyyy and other times dd/MM/yyyy.

However, using the interpretation of the time values as unix timestamps yields times that are much longer than a day.  I think instead the time values might be milliseconds within a JD.  The largest value I see is 84,026,356 and there are 86,400,000 milliseconds in a day.

Can you decode the date and time of some waypoints using your device software so we know what the correct answer is suppose to be, e.g. for some points from lowrance-v4.usr or another one of the reference files with version >= 4?

Since JD start at noon UT the same JD may convert to either of two Gregorian dates depending on if we add more or less than 12 hours for the time.  Ideally we would have waypoints on either side of midnight UT in our test case.

tsteven4:
Assuming my hypothesis is correct, i.e. usr v4 waypoint datetimes are stored as a julian day number followed by a number of msec within a day so that the julian date = julian day number + number of msec/8,6400,000 I get the following datetimes for lowrance-v4.usr.  Can you verify if these datetimes match those shown on the device?

[lowrance-v4.txt](https://github.com/gpsbabel/gpsbabel/files/2644642/lowrance-v4.txt)

bjkowalski:
That was a rather disappointing exercise.

My Lowrance unit is unable to read the waypoints on the lowrance-v4.usr
reference file.  Lowrance appears to have little quality control or
motivation to ensure USR file compatibility is maintained between product
generations.  In my attempt to update GPSBabel I have discovered my unit
creates USR 2/3 files with an additional data element that is not found in
the reference files and I have not yet found a way to detect the difference
in formats without brut force attempts to read one way, fail, and then try
the other.  Now I find a USR4 file that my unit cannot read fully.  This
experience echos frustration I have read about on some of the support
bulletin boards.

Since I only have a single unit to test with, and it is a recent
generation, I am less than satisfied with the confidence level in my end
product.  No matter what we release, short of dropping Lowrance support, I
believe users may encounter issues.  While I think we will decode USR files
correctly I am less confident that the USR output I generate will work in
all units.

Sigh..  I will keep pursuing this and have already started to reach out to
other Lowrance owners to see if I can borrow older generation units for
testing.

But I digress from the topic at hand.  Using a "small" (15 waypoint) file I
had created for ICON test purposes I have used GPSBabel to output unicsv
and then compared the results to my unit.  Your hypothesis about BAD time
conversion was spot on.

You comment on date format in the USR file is also correct - at least in my
unit, I have the ability to select either the MM/dd/yyyy or the dd/MM/yyyy
format as a user preference.  The date is also stored in binary format so
the ASCII is just for human consumption and not used by the unit.

I've attached both the PDF showing the conversion results and the info from
my unit.  I've also attached the USR file, the unicsv, and the PDF showing
the comparison.

Also I performed the same test with the USR2 and USR3 files generated by my
unit using the same input.  With 2/3 conversion the only difference in time
values is the localization of DST adjustments in the unit output and some
rounding differences.  Something to keep in mind when we address the
USR4/5/6 time conversion correction.

USR 2/3 time conversion uses a single 32-bit value (# seconds since Jan 1,
2000) but USR 4/5/6 uses two 32-bit values as you described above.  Can you
provide the patch so I can test it with these new files?

tsteven4:
I never got your attachments to verify the timestamp properties, but here is my patch anyway.  I did find some files on score-international.com that had v4 usr and gpx, but there seemed to be a 12 hour shift between their gpx and the one I generate with this patch from their usr.  There is confusion as a julian date starts at a calendar time of noon.  I carefully added and subtracted this 12 hour offset, so maybe whatever they used to create the gpx had a bug.  I do note that the julian date conversions this patch provides do match what one gets from https://aa.usno.navy.mil/data/docs/JulianDate.php. If the patch generates gpx files from a usr from your unit that has timestamps that agree with what your unit says then I think we are good.

This doesn't include the rounding recommendations from the previous patch which I still recommend.
It has some qDebug statements you will want to take out, but I left them in there because the "fixed << qSetRealNumberPrecision(9)" bit took me a bit to figure out.  Note that debug statements have output that can be compared with the julian date converter cited above.
It ifdef'd out some code in lowranceusr4_writestr so I could test.


[timestamp.txt](https://github.com/gpsbabel/gpsbabel/files/2649582/timestamp.txt)

bjkowalski:
I just grabbed one of those USR files and they seem to indicate there is a
problem with Altitude for USR2, at least I don't think they are racing at
200ft below sea level.

I've implemented your time patch.  I had to take out the 12 hour bias as
the times being generated were off by 12 hours.  Makes me think that
Lowrance has this issue.

lowranceusr.cc
```diff
@@ -423,28 +960,397 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
+    } else {
+      printf(MYNAME " parse_waypt: waypt_type = %d\n",waypt_type);
+    }
   }
 
   // Version 3 has a depth field here.
-  if (reading_version >= 3) {
+  if (reading_version == 3) {
     float depth_feet = gbfgetflt(file_in);
     if (std::abs(depth_feet - 99999.0)  > .1) {
```
tsteven4, line 944: L973 implies the field is stored as a float.
L974 implies that the magic value +99999.0 indicates no data.
compare this with the comment at L1808-1809.

lowranceusr.cc
```diff
@@ -758,24 +1802,140 @@ lowranceusr_waypt_disp(const Waypoint* wpt)
   /* USER waypoint type */
   short int WayptType = 0;
   gbfputint16(WayptType, file_out);
+
+  if (writing_version == 3) {
+    float depth = WAYPT_HAS(wpt, depth) ?
+                  METERS_TO_FEET(wpt->depth) : -99999.0;
+    gbfputint32(depth, file_out);
```
tsteven4, line 1813: L1808 implies the magic value -99999.0 indicates no data.
L1809 converts the float depth to an int, and saves it as an int, not a float.
Compare to comment at L973-974.
These operations should be symmetric between read and write.

lowranceusr.cc
```diff
@@ -423,28 +960,397 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
+  if (create_date > 2440587) {
+    wpt_tmp->SetCreationTime(lowranceusr4_get_timestamp(create_date,
+                             create_time));
+  }
+
+  /* Unused byte */
+  char unused_byte = gbfgetc(file_in);
+
+  /* Altitude/Depth in feet */
+  wpt_tmp->altitude = FEET_TO_METERS(gbfgetflt(file_in));
```
tsteven4: I am not sure altitude is what you want.  I see lots of ele elements now with values of 0.0 in gpx output generated from usr input.  This suggests that 0.0 values are being falsely interpreted as valid sea level.  Perhaps there is a flag somewhere in the usr file that indicates if this field is valid (like in the unused byte just before this?).
Internally we use the magic value unknown_alt to indicate invalid altitude data in a waypoint. We are not consistent in how we handle unknown values (magic values like unknown_alt vs. flags like wpt_flags bit vs. gpsbabel_optional::optional.)

The definition of depth in defs.h regarding "thickness" of a waypoint may not be used consistently.

What was the case you mentioned with issues 200ft below sea level?  Is that related to this change?

In any event read and write need to be symmetric, here you currently read into altitude, but in lowranceusr4_waypt_disp you write from depth.

tsteven4: 2016 baja 1000 has usr versions 2, 4 and gpx available.  A few of the points in the prerun gpx file have non-zero elevations, but their gpx has mostly ele elements with 0 values which are likely incorrect (the element is optional and shouldn't be included if the data is unknown).  From the position we should be able to confirm the elevations from DEMs.

tsteven4: I think usr v4 elevation is stored in what we call Flags.  I suspect what you now call Altitude/Depth is something else, perhaps depth.  We think usr v3 has both altitude and depth.

  /* Long/Lat */
  wpt_tmp->longitude = lon_mm_to_deg(gbfgetint32(file_in));
  wpt_tmp->latitude = lat_mm_to_deg(gbfgetint32(file_in));

  **/* Flags, discard for now */
  fsdata->flags = gbfgetint32(file_in);**

tsteven4: Scratch that, its not in Flags.  somewhere else.

bjkowalski: I was able to find a USR2 file on an iFinder HUNTc which had waypoints with lat, lon, elev, date, time values.  With the updates now committed the output using unicsv generates the same values as displayed on the iFinder device.  In that process I discovered another way Lowrance indicates unknown elevation in USR2 files.  There are now two different check methods when reading elev in USR2 and USR3 files.

lowranceusr.cc
```diff
@@ -1,238 +1,537 @@
...
-  { 10042, "Disabled Cache" },                // black box with red X
-
-  {	 -1, nullptr }
+  { 10038,    "Geocache" },             // flag buoy
+  { 10016,    "Geocache Found" },       // exclamation
+  { 10043,    "Micro-Cache" },          // small red dot
+  { 10065,    "Virtual cache" },        // ghost
+  { 10051,    "Multi-Cache" },          // 4 dots in rhombus shape
+  { 10068,    "Unknown Cache" },        // ? mark
+  { 10045,    "Locationless (Reverse) Cache" },  // hiding hunter
```
GPSBabelDeveloper, line 260: Nobody's noticed in all these years, but is this perhaps a "Letterbox cache"?

Described at https://www.geocaching.com/about/cache_types.aspx

lowranceusr.cc
```diff
@@ -1,238 +1,537 @@
...
+  const int      value;
+  const char*    icon;
+  const char*    color[7];
+} lowranceusr4_icon_mapping_t;
+
+const lowranceusr4_icon_mapping_t lowranceusr4_icon_value_table[] = {
+
+  /*  USR     GPX Symbol                COLOR1     COLOR2     COLOR3    COLOR4     COLOR5    COLOR6      COLOR7         HOOK2 Displays */
+
+  {     1,    "diamond 1"           , { "blue",   "magenta", "orange", "yellow",  "greem",   "aqua",    "white" }},   // diamond
```
GPSBabelDeveloper: Throughout you use gree*m*. Did you mean "green"? If all 7 elements of this really are the same everywhere, we should reconsider a more concise data structure...

lowranceusr.cc
```diff
@@ -1,238 +1,537 @@
...
+           (A->latitude == B->latitude) &&
+           (A->longitude == B->longitude));
+}
+
+static void
+register_waypt(const Waypoint* ref)
+{
+  Waypoint* wpt = const_cast<Waypoint*>(ref);
+
+  for (int i = 0; i < waypt_table_ct; i++) {
```
GPSBabelDeveloper, line 442: Should waypt_table (and _ct and _sz) be something like a QVector<Waypoint*> or maybe a QList? I don't this is really a high-running path, but these linear walks of the structure and the manual resizing below make me cringe.
This CL is already large. If we defer this to a future point, so be it.

lowranceusr.cc
```diff
@@ -1,238 +1,537 @@
...
+    waypt_table_sz += 32;
+    if (waypt_table) {
+      waypt_table = (Waypoint**) xrealloc(waypt_table, waypt_table_sz * sizeof(wpt));
+    } else {
+      waypt_table = (Waypoint**) xmalloc(waypt_table_sz * sizeof(wpt));
+    }
+  }
+
+  if (global_opts.debug_level >= 2) {
+    printf(MYNAME " adding waypt %s (%s) to table at index %d\n",
```
GPSBabelDeveloper, line 460: throughout this patch, there's a lot of struggle between C and C++/Qt. (I know this didn't start this week. I'm just talking from bed...)
qDebug() << MYNAME << "adding waypt" << wpt->shortname<<  "to index" << waypt_table_ct;
At some point, I plan to provide a custom class that unifies our formatting, debugging levels(byte level, logical) but it's not here yet. This patch just spends a LOT of effort formatting deb code and at some point, we should ask if we're doing the debugger's work...

lowranceusr.cc
```diff
@@ -248,12 +547,93 @@ lowranceusr_readstr(char* buf, const int maxlen, gbfile* file)
         buf[i] = '*';
       }
     }
+  }
+  return len;
+}
+
+//  Starting with USR 4, adopted a UTF-16 character string format
+static int
+lowranceusr4_readstr(char* buf, const int maxlen, gbfile* file, int bytes_per_char)
```
GPSBabelDeveloper: Bytes per_character is always either 1 or 2, right? It's either ASCII or UTF-16, right?
Can this whole function become something like:
QString lowrance4_readstr(gbfile, int bytes_per_char)  {
int org = len = gbfgetint32(file);
buf = xmalloc(len) // Probably something more clever here to be done
bytesread = gbfread(buf, bpc, len, file);
QString rval;
if (bpc == 1) {
  rval = QString::fromUtf8(buf, len);
} else {
  rval = QString::FromUtf16(buf, len)
}
return rval;

Maybe we give up on gbfread (word size, endian safety, ...) and make everything QFiles and then we can stream QByteArrays directly.

lowranceusr.cc
```diff
@@ -248,12 +547,93 @@ lowranceusr_readstr(char* buf, const int maxlen, gbfile* file)
...
+       to write more than 2^32 bytes */
+    len = (int)(0x7fffffff / bytes_per_char);
+  }
+
+  gbfputint32(len*bytes_per_char, file_out);
+
+  if (bytes_per_char == 1) {
+    (void) gbfwrite(CSTR(buf), 1, len, file);
+  } else {
+    for (int i = 0; i < len; ++i) {
```
GPSBabelDeveloper: Maybe writestr() also gets a similar toUtf() and toUcs16() beating.

lowranceusr.cc
```diff
@@ -355,34 +842,63 @@ lon_mm_to_deg(double x)
...
-  if (METERS_TO_FEET(wpt_tmp->altitude) <= -10000) {
-    wpt_tmp->altitude = unknown_alt;
+  /* 
+   * From available data, looks like stream version = 0 used float to store for altitude in feet
+   * and version 1 used int32.  If that is found to not be the case may need to add flag for input.
+   * Only have a small sample size.
+   */
+  if (rstream_version == 0) {
+    float read_alt = gbfgetflt(file_in);
+    char buf[16];
```
GPSBabelDeveloper: this hunk is weird. Using sprintf and strcmp to determine isnan isn't great. Is this basically
read_alt = getflt() (may rely on safe upcast of float todouble)
if isNaN(read_alt)) {
wpt_tmp->altitude = unknown_alt; (Steve, did we just not get optional to alt because of the zillion places it's used?)
} else {
  wpt_tmp->altitude = FTOMeters()
} else {
 wpt_tmp->altitude = ...
}

lowranceusr.cc
```diff
@@ -391,9 +907,25 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
     wpt_tmp->shortname = buff;
   }
 
-  if (global_opts.debug_level >= 1)
-    printf(MYNAME " parse_waypt: Waypt name = %s Lat = %f Lon = %f alt = %f\n",qPrintable(wpt_tmp->shortname), wpt_tmp->latitude,
-           wpt_tmp->longitude, wpt_tmp->altitude);
+  if (global_opts.debug_level > 1) {
+    if (global_opts.debug_level == 99) {
+      printf(" %16.16s %+15.10f %+15.10f",
```
GPSBabelDeveloper: This is an awful lot of finicky formatting in debug code. Are we doing the debuggers's job for it? Could the finicky interleaving between string constants and debug values just be done with qDebug() ? Note http://doc.qt.io/qt-5/qdebug.html#formatting-options if you really want to display tenths of feet.
Maybe more readable to bust all this type of code into different functions and just pass the pointer to the wpt around.

lowranceusr.cc
```diff
@@ -406,15 +938,35 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
+  if (global_opts.debug_level > 2) {
+    if (global_opts.debug_level == 99) {
+      QDateTime dt;
+      QTime t;
+      dt = wpt_tmp->GetCreationTime();
+      t  = wpt_tmp->GetCreationTime().time();
+      printf(" %s %s", qPrintable(dt.toString("yyyy/MM/dd")), qPrintable(t.toString("hh:mm")));
+    } else {
+      printf(MYNAME " parse_waypt: creation time %d, base_time %d waypt_time %d\n",
+             (int)wpt_tmp->creation_time.toTime_t(),
```
GPSBabelDeveloper: Notably qDebug knows how to print a QDateTime in a way that's human-readable. qDebug() << "created" << wpt_tmp->creation_time << "base time" << base_time_secs should do something reasonable.
It's not really suited to the writing of the files, but it saves that ickiness in both of these two printf lines.

lowranceusr.cc
```diff
@@ -406,15 +938,35 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
   /* Symbol ID */
-  wpt_tmp->icon_descr = lowranceusr_find_desc_from_icon_number(gbfgetint32(file_in));
+  int icon_number = gbfgetint32(file_in);
+  if (global_opts.debug_level == 99) {
+    printf(" %08x (%d)", icon_number, icon_number);
+  }
+  wpt_tmp->icon_descr = lowranceusr_find_desc_from_icon_number(icon_number);
   if (wpt_tmp->icon_descr.isNull()) {
     char nbuf[10];
     snprintf(nbuf, sizeof(nbuf), "%d", le_read32(buff));
```
GPSBabelDeveloper: wpt_tmp->icon_descr = QString::number(le_read(buff));

lowranceusr.cc
```diff
@@ -423,28 +975,399 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
 
+  if (global_opts.debug_level == 99) {
+    printf("\n");
+  }
 }
 
+static void
+lowranceusr4_parse_waypt(Waypoint* wpt_tmp)
+{
+  char name_buff[MAXUSRSTRINGSIZE + 1];
```
GPSBabelDeveloper: Put these closer to their actual use. Top-of-function quit being a rule long ago. (Yes, a lot of the code in GPSBabel is really old...)

lowranceusr.cc
```diff
@@ -423,28 +975,399 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
+    snprintf(nbuf, sizeof(nbuf), "%d", le_read32(name_buff));
+    wpt_tmp->icon_descr = nbuf;
+  }
+
+  /* Color ID */
+  fsdata->color = gbfgetint16(file_in);
+  fsdata->color_desc = lowranceusr4_find_color_from_icon_number_plus_color_index(fsdata->icon_num, fsdata->color);
+
+  /* Waypoint descr; input is 2 bytes per char, we convert to 1 */
+  int desc_len = lowranceusr4_readstr(&desc_buff[0], MAXUSRSTRINGSIZE, file_in, 2);
```
GPSBabelDeveloper: The repeated blocks of boilerplate are a good indication that this calling convention needs "fixed" to return  useful value.

lowranceusr.cc
```diff
@@ -423,28 +975,399 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
+             waypoint_version, name_len, name_buff);
+      if (reading_version > 4) {
+        printf("  %10u ", fsdata->uid_unit2);
+      }
+      printf(" %+15.10f %+15.10f", wpt_tmp->longitude, wpt_tmp->latitude);
+      printf(" %08x %4d %4d %7s", fsdata->flags, fsdata->icon_num, fsdata->color, qPrintable(fsdata->color_desc));
+      printf(" %6d %16s", desc_len, desc_buff);
+      QDateTime dt;
+      QTime t;
+      dt = wpt_tmp->GetCreationTime();
```
GPSBabelDeveloper: Changing all the printfs to qDebugs gets the variable creation and manipulation out of this nested if debug/if more debuglevel.

lowranceusr.cc
```diff
@@ -423,28 +975,399 @@ lowranceusr_parse_waypt(Waypoint* wpt_tmp)
...
+        printf(" Unit Number2");
+      }
+      printf(" Latitude        Longitude       Flags    ICON Color        Length Description     ");
+      printf(" Date       Time  Unused   Depth    LoranGRI LoranTda LoranTdb\n");
+
+      printf(MYNAME " parse_waypoints: ");
+      if (reading_version > 4) {
+        printf("-------- -------- -------- -------- ");
+      }
+      printf("----------- -------- -------- -------- ------ ----------------");
```
GPSBabelDeveloper, line 1127: This will admittedly be difficult/fragile with qDebug(), but it's probably fragile here, too. Maybe it all becomes a qtextstream and http://doc.qt.io/qt-5/qtextstream.html#setFieldWidth and let that handle the columnization.

lowranceusr.cc
```diff
@@ -455,226 +1378,361 @@ lowranceusr_parse_routes()
...
+
+    if (opt_ignoreicons == nullptr) {          /* Option not specified if NULL */
       Waypoint* wpt_tmp = new Waypoint;
-      lowranceusr_parse_waypt(wpt_tmp);
-      route_add_wpt(rte_head, wpt_tmp);
+
+      /* position coord lat & long */
+      wpt_tmp->latitude = latitude;
+      wpt_tmp->longitude = longitude;
+      wpt_tmp->altitude = 0;
```
GPSBabelDeveloper: Don't explicitly set alt if we don't know it. (That'll be more important later.(

lowranceusr.cc
```diff
@@ -455,226 +1378,361 @@ lowranceusr_parse_routes()
...
+      snprintf(buff, sizeof(buff), "Event Marker %d", i+1);
```
GPSBabelDeveloper: ...shortname = "Event marker" + QString::number (i+1);

In general, avoid s[n]sprintf when we can.

lowranceusr.cc
```diff
@@ -455,226 +1378,361 @@ lowranceusr_parse_routes()
...
-      waypt_add(wpt_tmp);
-    }
+  /* Color ID, discard for now */
+  trail_color = gbfgetint32(file_in);
+
+  /* Comment/description */
+  int desc_len = lowranceusr4_readstr(&desc_buff[0], MAXUSRSTRINGSIZE, file_in, 2);
+  if (desc_len) {
+    desc_buff[desc_len] = '\0';
+    trk_head->rte_desc = desc_buff;
```
GPSBabelDeveloper: WIth foo_readstr() returning QStrings, we'd be out of the business of manually terminating and testing them before we used them.About every caller would benefit.

GPSBabelDeveloper (commented):
I had to stop, but it seemed I had the biggest part reviewed. Well, that and it's 3am.

Looks good. Most of my comments can to in a future commit. This format is approaching .gdb for complexity and unknown fields. :-/

lowranceusr.cc
tsteven4: I already have a patch that does this.

lowranceusr.cc
tsteven4: I already have a patch that uses codecs for usr 4 readstr and writestr.

lowranceusr.cc
bjkowalski: They should all be green.

While there are only 7 values (0-6) used to represent colors, looking only at what the unit displays I couldn't derive a method for consistently representing what was being displayed without separate, explicit values.  Some like anchor and rock appear to be the same icon displayed but the unit allows me to pick from seven icons and each one has a different color value.  It also appears to me, from looking at the unit, that there are actually 8 colors (blue, magenta, orange, yellow, green, aqua, white, red) represented by the 7 values (0-6).

I need access to a higher quality screen than the 7" 800x480 display on my unit.

lowranceusr.cc
bjkowalski: Historic code.  I changed it to be "unknown_alt" as that seems more appropriate given that we are converting an Event Marker (icon/lat/lon only) to a waypoint.

tsteven4:
Here is the codec patch.  I can't count how many times I have merged this.  I was holding off as this PR is huge, but maybe we should just add it in.  It is very nice to have you working on this with your access to a real device.  Thanks for sticking with this sticky wicket.

Please see the FIXME comments in any event.

[codec.txt](https://github.com/gpsbabel/gpsbabel/files/2658037/codec.txt)

tsteven4:
When reading lowrance-v4.usr and writing gpx most waypoints show <ele> tags of 0.  The ANCHOR1 and ANCHOR2 waypoints have <ele> tags with values of ~ 5 or 15m.  Are these really elevations, or are the depths?  These points are in the sea off the coast of Malaysia.  Maybe a better rosetta stone than the random things we have found on the web is to use a lowrance device to create the translation.  I would imagine the software developers of the lowrance software have access to the usr specification while 3rd party developers probably do not.  Can you use your hook2 and load lowrance-v4.usr and then have the hook2 write a gpx?  Alternatively, if you have waypoints with elevations and or depths already in your hook2 can you write out usr v2, v3, v4, v5, v6, and gpx?

bjkowalski:
Using my HOOK2 was how I got as far as I did.  Now that I have revived my iHunt Color (hand held) that has actual elevations I can verify my previous assertion that the HOOK2 generates a USR2 file that is not compatible with older devices.  After your previous comments I spent more time looking at how the lowrance code was using depth and altitude.  Not good.  I will be re-working the legacy code to use format specific for depth and that should correct some of the bad behavior when moving elevations to GPX.  I have already made changes to USR4 but haven't pushed everything yet.

The GPX files the HOOK2 generates are very minimal in content.

Sorry this has been as dragged out as it has been.  I though there were just a few bugs but the more I dig the more pop up.

I REALLY appreciate everyone's comments and suggestions.  Hopefully we are on the downside of this mountain but until I have completed more exhaustive testing with the iHunt, and generated tracks and trails to convert between the devices I will not be done.  Also the Baja data motivated me to figure out how to leverage Google earth to see if my trails/routes make sense.

lowranceusr.cc
bjkowalski: Adapted the codec patch.

lowranceusr.cc
bjkowalski: Issues with inconsistent use of altitude and depth have been addressed.

lowranceusr.cc
bjkowalski: Adopted in the latest QString changes.

lowranceusr.cc
bjkowalski: Consolidated to a single lowranceusr_readstr() using QStrings.

robertlipe:
Please no new sprintfs.that are about to go into.a.qstring. "bla" +QString:
Number(I+1)  here should do.

I'm sure you can fund a thousand counterexamples in our code. I'm just
trying to evolve to newer practices.

robertlipe:
Prefer not setting the alt over setting it to zero. A fake zero annoys
people near coasts or in Death Value.

Some day alt.(and its cousins) could become a std::optional

bjkowalski:
Sorry - my comment was without the associated commit.  I eliminated the snprintf() per your earlier suggestion.

bjkowalski:
Regarding alt - I am only setting to zero if the device reports it as zero now.  Found at least two different ways lowrance units report it as UNKNOWN.  So now I see many more UNKNOWNs in reading USR files than I see zeros.

tsteven4:
QDateTime ctors use Qt::LocalTime by default.
Waypoint::GetCreationTime() returns a datetime which could be any Qt::TimeSpec, i.e. Qt::LocalTime, Qt::UTC, Qt::OffsetfromUTC or Qt::TimeZone.

I think it is safest to use toLocalTime(), toUTC(), etc. to get what you want on output, and not count on the Qt::TimeSpec being a particular value for a waypts creation_time field.

I note that lowranceusr4_get_timestamp intentionally uses Qt::UTC (my understanding was lowranceusr4 stored UTC not local times),  and
 `wpt_tmp->SetCreationTime(lowranceusr4_get_timestamp(create_date,
create_time));`
merely copies the QDateTime or gpsbabel::DateTime to creation_time without changing the Qt::TimeSpec.

tsteven4:
@bjkowalski, Thanks for all your investigation and improvements.  Do you have anything pending?

bjkowalski:
I have nothing additional to add at this time.

PR #271, opened 2018-11-15
Start new track segments on timer events (start/stop/...)

gromit1811:
Insert track segment separators when encountering a timer start event in a
FIT file (usually caused by manually pushing the start button). This way,
pauses in a FIT file don't appear as continuous tracks.

Also adapt reference tracks by adding suitable <trkseg> tags to please "make
check"/testo.

### DIFF
--- a/garmin_fit.cc
+++ b/garmin_fit.cc
@@ -30,6 +30,7 @@
 const int kIdDeviceSettings = 0;
 const int kIdLap = 19;
 const int kIdRecord = 20;
+const int kIdEvent = 21;
 
 // constants for message fields
 // for all global IDs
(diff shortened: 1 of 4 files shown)


gromit1811:
The reference FIT files with start/stop timer events all have one additional track point following the stop event. The FIT files generated by my Garmin Edge 130 which I tested while implementing this feature don't have any track points between stop and start events.
The way this is implemented now will always cause one additional track segment with a single track point right after each stop event for FIT files with the same message sequence as the reference files. This doesn't look particularly nice.
This could be fixed by moving track points after a stop event but with the same timestamp as the stop event logically before the stop event i.e. into the previous track segment. Should I implement this?

GPSBabelDeveloper:
Hi, Martin. The conversation here went to the mailing list and there were some minor changes discussed. Any updates here or should this go in as it's standing? It's not quite "last call", but I'd like to start heading that direction and burning down the queue. Thanx. RJL

gromit1811:
Hi Robert,

from my point of view, it's ready for merging. My initial pull request had
the problem that it created unwanted single-point track segments for some of
the reference files in the gpsbabel source tree. But I've fixed this in the
meantime (by starting new track segments only at FIT start events, not at
any event) and updated the pull request. I can't remember anything else
having been discussed on the mailing list.

Executive summary: Go!

Thanks,
Martin

PR #275, opened 2018-11-24
simplify conversion from utm.

tsteven4:
add test cases for Norway and Svalbard exceptions.

The changes in gpsmath.cc do NOT change the test results, i.e. the old and new code is equivalent.


jeeps/gpsmath.cc
```diff
@@ -2169,32 +2169,14 @@ static int32 GPS_Math_UTM_Param_To_Mc(int32 zone, char zc, double* Mc,
                                       double* E0, double* N0, double* F0)
 {
 
-  if (zone>60 || zone<0 || zc<'C' || zc>'X') {
+  if (zone>60 || zone<0 || zc<'C' || zc>'X' || zc=='I' || zc=='O') {
```
GPSBabelDeveloper, line 2177: Please consider this a warning if not a fatal. I see only two callers of it and the callers of those are inconsistent in checking for error  case. Let's just splash a warning or a fatal so someone knows their data is bogus.

Thanx for making this less painful to read. UTM really is icky.

GPSBabelDeveloper (approved):

tsteven4:
The lastest changes in gpsmath.cc do change the test results, an old bug in GPS_Math_LatLon_To_UTM_Param has been fixed.
The changes in GPS_Math_UTM_Param_To_Mc are equivalent and don't influence the test results.

jeeps/gpsmath.cc
tsteven4: your correct, but the problem with jeeps callers ignoring the return value is systematic.  I left that unresolved and moved on.

PR #279, opened 2018-11-27
obsolete DEBUG_MEM

tsteven4:


GPSBabelDeveloper (approved):
Good. This was awesome at the turn of the century but valgrind is so much better. Thanks for vacuuming this up.

PR #287, opened 2018-12-27
fix up style file OFIELD option processing.

tsteven4:
The fourth field of a OFIELD line in the style file, which represent options, may or may not be split into multiple tokens.
Note that there were two issues here:
1) we only looked for options when we had 5 or more tokens, not 4.
2) we only looked at the fourth token.


GPSBabelDeveloper:
That no no delimiter stuff is something I regret. Patch approved.

PR #288, opened 2018-12-27
enhance xcsv reader to not ignore LAT_DIR,LON_DIR.

tsteven4:
this came up on [Gpsbabel-misc] Globalsat DG-500.

### DIFF
--- a/csv_util.cc
+++ b/csv_util.cc
@@ -169,6 +169,8 @@ static route_head* csv_track, *csv_route;
 static double utm_northing, utm_easting, utm_zone = 0;
 static char utm_zonec;
 static UrlLink* link_;
+static gpsbabel_optional::optional<bool> lat_dir_positive;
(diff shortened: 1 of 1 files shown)


robertlipe:
This is order sensitive, right? Should we keep these until the full line is
parsed and instead flip the sign before the _add()?

tsteven4:
You are correct, and I believe the PR already does this.  That is why 
the correction is done after each line is fully parsed in 
xcsv_data_read, and not in xcsv_parse_value.  xcsv_parse_value saves the 
direction in the static global variables lat_dir_positive, 
lon_dir_positive.  xcsv_data_read uses these after all fields in a line 
are parsed.  I just flipped LON_NMEA and LON_DIR to verify this, and it 
works as it should regardless of the order of LON_NMEA and LON_DIR.

GPSBabelDeveloper:
I see now you are correct, sir. Apologies for the distraction. Carry on!

PR #291, opened 2018-12-29
Corrected handling elevation and speed

aslak47:
When converting fit-files from Oregon 700 activities to gpx-files, speed and elevation (altitude) were not converted because the fit-files have both standard and enhanced values and only the enhanced value have a not null value.<br>
Added tests for null values for both speed and alt.


tsteven4:
Thanks.  Can you provide a reference file and a short testcase that fails before your fix and passes afterwards?  See testo.d/garmin_fit.test.  You can run the test with "./testo garmin_fit".   The addition of a test case makes maintenance easier and ensures that your fix continues to work.

PR #292, opened 2019-01-11
fix leaks when deleting wpts in filters.

tsteven4:


discard.cc
```diff
@@ -91,17 +91,19 @@ void DiscardFilter::fix_process_wpt(const Waypoint* wpt)
     switch (what) {
     case wptdata:
       waypt_del(waypointp);
+      delete waypointp;
```
GPSBabelDeveloper, line 94: Offhand, what do you think of releasing the object in what was "obviously" meant to be the destructors instead of leaving it live and being released in the callers? It seems like a hazard that these are still dangling and live after calling something named "delete"

GPSBabelDeveloper (commented):
Lgtm, but let's reconsider the semantics

PR #293, opened 2019-01-11
modernization of track filter

tsteven4:
This replaces legacy code (time_t, qsort, atoi, xmalloc'd dynamic arrays) with Qt classes and std lib.  It the process some bugs are fixed.   bool and auto are used more.  some redundant casting is eliminated.  Test cases are enhanced.  Please see the individual commit messages.


tsteven4:
It is possible to continue the elimination of duplicate information by eliminating the use of track_list.  This requires access to my_track_head similar to what the sort filter does.  Stay tuned.

trackfilter.cc
```diff
@@ -57,90 +72,54 @@ int TrackFilter::trackfilter_opt_count()
...
+  qint64 result;
+
+  QRegularExpression re("^([+-]?\\d+)([dhms])$", QRegularExpression::CaseInsensitiveOption);
+  assert(re.isValid());
+  QRegularExpressionMatch match = re.match(arg);
+  if (match.hasMatch()) {
+    bool ok;
+    result = match.captured(1).toLong(&ok);
+    if (!ok) {
+      fatal(MYNAME "-time: invalid quantity in move option \"%s\"!\n", qPrintable(match.captured(1)));
```
GPSBabelDeveloper, line 86: Optional throughout: Use Fatal() and Warning() to avoid qPrintable
Somewhere I have a patch that allows you to create a Logger object that accepts MYNAME ...

trackfilter.cc
```diff
@@ -57,90 +72,54 @@ int TrackFilter::trackfilter_opt_count()
...
-    if (c >= '0' && c <= '9') {
-      t1 = (t1 * 10) + (c - '0');
-      continue;
+  qint64 result;
+
+  QRegularExpression re("^([+-]?\\d+)([dhms])$", QRegularExpression::CaseInsensitiveOption);
+  assert(re.isValid());
+  QRegularExpressionMatch match = re.match(arg);
+  if (match.hasMatch()) {
+    bool ok;
```
GPSBabelDeveloper, line 83: Do our warning tools know that ok won't be used uninitialized? Interesting.

trackfilter.cc
```diff
@@ -719,86 +664,82 @@ void TrackFilter::trackfilter_synth()
...
 
-  return mkgmtime(&time);
+  return result;
 }
 
-int TrackFilter::trackfilter_range()		/* returns number of track points left after filtering */
+void TrackFilter::trackfilter_range()
 {
-  time_t start, stop;
+  QDateTime start, stop; // constructed such that isValid() is false, unlike gpsbabel::DateTime!
```
GPSBabelDeveloper, line 696: Sorry. That was totally a compatibility crutch.

GPSBabelDeveloper (commented):
This is a big improvement in readability, helping make our 40 days and nights through the desert seem worth it. 
Thanx for being so thorough.

trackfilter.cc
tsteven4: I haven't found any complaints, but there is always another checker to check.

trackfilter.cc
tsteven4: Maybe later.  I think you already commited this in src/core/logging.h, e.g.
stmsdf.cc:      Warning() << MYNAME << "Invalid internal field type" << what;

PR #295, opened 2019-02-08
bug fixes for xcsv reader.

tsteven4:
For some lines data must be accumulated from several fields
in arbitrary field order.
Previously this was done with global variables.  Now a variable to
hold the accumulated data for a line is constructed/destructed for
each line.

This eliminates the possiblity of unintended communication between
lines.

This also eliminated some unintended communication between the reader
and the writer through global variables that they previously shared
(csv_track, csv_route).

An ordering sensitivty to TRACK_NEW and TRACK_NAME is removed.

For the reader, conditional allocation of route_heads is removed
from xcsv_parse_val, and centralized in xcsv_data_read.
This eliminates some undetected
memory leaks which hid as "still reachable" due to the pointer
being saved in a global variable (csv_track, csv_route).

Undesirable reading of the route_head->Q variable is
eliminated improving encapsulation.  I beleive the intent was
never to decide if the csv_track->Q was empty, i.e. it hadn't been
added to the global track_head yet.  I beleive what was desired was
to decide if csv_track had any waypoints. In any event the
centralization of route_head allocation and head addition makes new
track handling much more straight forward and robust.

The xcsv test case is enhanced to increase coverage.


tsteven4:
Robert,
There was considerable history around TRACK_NEW, TRACK_NAME, ROUTE_NAME.  If I didn't appreciate something, perhaps related to the use of QUEUE_EMPTY, please let me know.
Thanks
https://github.com/gpsbabel/gpsbabel/blob/0f6c9ecb41b6e964ac3cb2f22d1b9aa2a1dfd06e/csv_util.cc#L1303-L1310

csv_util.cc
```diff
@@ -167,11 +167,19 @@ static double oldlat = 999;
 
 static int waypt_out_count;
 static route_head* csv_track, *csv_route;
-static double utm_northing, utm_easting, utm_zone = 0;
-static char utm_zonec;
-static UrlLink* link_;
-static gpsbabel_optional::optional<bool> lat_dir_positive;
-static gpsbabel_optional::optional<bool> lon_dir_positive;
+
```
GPSBabelDeveloper, line 170: Those lines above might go into this "class members" (that we can't quite call class data yet..) and then these become foo_ instead of foo_ but I like this as a stepping stone because it calls things out.

GPSBabelDeveloper (commented):
I like this. Thanx for untangling that.

PR #296, opened 2019-02-09
Introduce a RouteList class.

tsteven4:
Introduce a RouteList class to define the interface to our lists of routes.

Initially this is backed by QueueList, which adds iterator support to our legacy queues.

As a separate commit, QueueList is replaced by QList, eliminating the use of legacy queues for route lists.


bend.cc
```diff
@@ -158,17 +158,17 @@ void BendFilter::process_route_orig(const route_head* route_orig)
...
     process_route_orig(route_orig);
   }
 }
 
 void BendFilter::deinit()
 {
-  route_flush(routes_orig);
-  xfree(routes_orig);
+  routes_orig->flush();
+  delete routes_orig;
```
GPSBabelDeveloper, line 170: Can routes_orig be a unique_ptr, perhaps passed in at creation time?

defs.h
```diff
@@ -652,7 +652,6 @@ struct computed_trkdata {
 class route_head
 {
 public:
```
GPSBabelDeveloper, line 651: Hooray. Death to one of our most deeply entrenched ugly Qs!

defs.h
```diff
@@ -667,12 +666,69 @@ class route_head
...
+  void disp_all(T1 rh, T2 rt, T3 wc);
+  template <typename T2, typename T3>
+  void disp_all(std::nullptr_t /* rh */, T2 rt, T3 wc);
+  template <typename T1, typename T3>
+  void disp_all(T1 rh, std::nullptr_t /* rt */, T3 wc);
+  template <typename T3>
+  void disp_all(std::nullptr_t /* rh */, std::nullptr_t /* rt */, T3 wc);
+
+  // Only expose methods from our underlying container that won't corrupt our private data.
+  // Our contained element (route_head) also contains a container (waypoint_list),
```
GPSBabelDeveloper, line 709: Those maintain an explicit (and easy to get out of sync) count only because they were linked lists. It was always the intent that they'd become foo.size() or foo.count() some day.
Please set your editor to wrap at 80 characters. (I know, I know, punched card, vt100, etc. :-) )

gtm.cc
```diff
@@ -575,20 +577,28 @@ gtm_read()
...
     track_add_wpt(trk_head, wpt);
   }
 
   /* Tracklog styles */
-  trk_head = first_trk_head;
-  for (i = 0; i != ts_count && i != real_tr_count; i++) {
-    trk_head->rte_name = fread_string(file_in);
+  // TODO: The format document states there are ts_count tracklog style entries,
+  //       and tr_count tracklog entries.
+  //       Some tracklog entries may be contiuation entries, so we turn these into
```
GPSBabelDeveloper: Typo nit: continuation

gtm.cc
```diff
@@ -575,20 +577,28 @@ gtm_read()
...
+  //       real_track_list.size() <= tr_count tracks.
+  //       If ts_count != real_track_list.size() we don't know how to line up the tracklogs,
+  //       and the real tracks, with the tracklog styles.
+  if (ts_count != real_track_list.size()) {
+    warning(MYNAME ": The number of tracklog entries with the new flag set doesn't match the number of tracklog style entries.");
```
GPSBabelDeveloper: Is this really a warning? Is there anything prescriptive we can suggest a user can DO about this case? Maybe it's weird enough we'd like to know if it happens.

gtm.cc
```diff
@@ -575,20 +577,28 @@ gtm_read()
...
+  }
+  for (i = 0; i != ts_count; i++) {
+    QString tname = fread_string(file_in);
     fread_discard(file_in, 12);
-    trk_head = reinterpret_cast<route_head *>QUEUE_NEXT(&trk_head->Q);
+    if (i < real_track_list.size()) {
+      trk_head = real_track_list.at(i);
```
GPSBabelDeveloper, line 603: Could the loop max be the std::min(ts_count, real_track_List.size() or is there some hope of resynchronizing things in this body?

main.cc
```diff
@@ -213,8 +213,10 @@ main(int argc, char* argv[])
   int opt_version = 0;
   int did_something = 0;
   const char* prog_name = argv[0]; /* argv is modified during processing */
-  queue* wpt_head_bak, *rte_head_bak, *trk_head_bak;	/* #ifdef UTF8_SUPPORT */
-  signed int wpt_ct_bak, rte_ct_bak, trk_ct_bak;	/* #ifdef UTF8_SUPPORT */
+  queue* wpt_head_bak;	/* #ifdef UTF8_SUPPORT */
```
GPSBabelDeveloper: I don't see UTF*_SUPPORT elsewhere in this patch or my tree. Is this automation gone wild?

main.cc
```diff
@@ -417,9 +417,10 @@ main(int argc, char* argv[])
            */
```
GPSBabelDeveloper, line 363: Meta question: is the remaining presence of global_opts.charset and this test of cet_cs_vec_utf8 now a bug? I don't think it works sensibly in garmin.cc, I'd drop gdb and mps on a dare and html.cc seems like a pretty silly use. Should we just drop this whole block and the entire use of it? I'm OK with that being the same or a different commit.

queue.h
```diff
@@ -61,6 +63,197 @@ queue* dequeue(queue* element);
 		(element) != (listhead); \
 		(element) = (tmp))
 
+// FIXME: g++ 7.3.0, -O2, with T=route_head yields warnings.
```
GPSBabelDeveloper, line 66: Wow. So much for our three opcode implementations. :-)
This looks gnarly to have worked out. Have one of your favorite beverages...

queue.h
```diff
@@ -61,6 +63,197 @@ queue* dequeue(queue* element);
...
+    {
+      return ptr_ != other.ptr_;
+    }
+
+  private:
+    queue* ptr_{nullptr};
+  };
+  //friend class Iterator;
+
+  class ConstIterator
```
GPSBabelDeveloper, line 141: I've run into this before of having to have a complete copy-paste block of 60 lines just to have a non-const iterator. I thought it was just my weak C++-fu...
Still, it feels kinda bad to have to have this many LOC just to describe an "obvious" container. Where did we go so wrong and should we try to fix that instead? Again, a meta question.

route.cc
```diff
@@ -18,161 +18,96 @@
...
+#include "src/core/datetime.h"  // for DateTime
+#include "src/core/optional.h"  // for optional, operator>, operator<
+#include <QtCore/QDateTime>     // for QDateTime
+#include <QtCore/QList>         // for QList<>::iterator
+#include <QtCore/QString>       // for QString
+#include <QtCore/QtGlobal>      // for foreach
+#include <algorithm>            // for sort
+#include <cstddef>              // for nullptr_t
+
+RouteList* global_route_list;
```
GPSBabelDeveloper, line 33: Can these  be static in route.cc? ("No" is an OK answer. Just trying to help keep things from leaking back out.)

route.cc
```diff
@@ -201,55 +129,21 @@ track_add_wpt_named(route_head* rte, Waypoint* wpt, const QString& namepart, int
...
-    Waypoint* wpt_next = reinterpret_cast<Waypoint *>QUEUE_NEXT(&wpt->Q);
-    wpt_next->wpt_flags.new_trkseg = 1;
-  }
-  wpt->wpt_flags.new_trkseg = 0;
-  dequeue(&wpt->Q);
-  rte->rte_waypt_ct--;
-  if (ct) {
-    (*ct)--;
-  }
+  // FIXME: It is misleading to accept namepart and number_digits parameters which
```
GPSBabelDeveloper, line 132: It's actually a bit weird that these are part of the add_wpt art list at all instead of methods that potentially set them on the entire list.  Having what's essentially methods in the route class that exist strictly to support the transform filter makes me wonder if we should make this just be the track filter's problem...

route.cc
```diff
@@ -262,255 +156,105 @@ void
 route_reverse(const route_head* rte_hd)
```
GPSBabelDeveloper: Should this just go into reverse_route.cc ?

route.cc
```diff
@@ -262,255 +156,105 @@ void
...
 }
 
 void
-route_flush_all()
+route_deinit()
 {
-  route_flush_all_tracks();
   route_flush_all_routes();
+  route_flush_all_tracks();
+  delete global_route_list;
```
GPSBabelDeveloper, line 184: If these were unique_ptrs, a set to nullptr or a route_list.reset() would both free them and make the pointers unusable.

route.cc
```diff
@@ -262,255 +156,105 @@ void
...
-  QUEUE_INIT(&my_route_head);
-  QUEUE_MOVE(&my_route_head, head_bak);
-
-  common_restore_finish();
+  global_route_list->sort(cmp);
 }
 
 void
-track_backup(signed int* count, queue** head_bak)
+track_backup(RouteList** head_bak)
```
GPSBabelDeveloper, line 225: The double dispatch in all these makes me wonder if GlobalTrackList and GlobalRouteList should be instances of a public PathList and let the compiler's vtable work it out. Let's think on that a bit more. Maybe this is a stepping stone to that.

route.cc
```diff
@@ -661,3 +404,140 @@ route_head::~route_head()
...
+// added themselves yet.
+void
+RouteList::add_wpt(route_head* rte, Waypoint* wpt, bool synth, const QString& namepart, int number_digits)
+{
+  ENQUEUE_TAIL(&rte->waypoint_list, &wpt->Q);
+  rte->rte_waypt_ct++;	/* waypoints in this route */
+  ++waypt_ct;
+  if (synth && wpt->shortname.isEmpty()) {
+    wpt->shortname = QString().sprintf("%s%0*d", CSTRc(namepart), number_digits, waypt_ct);
+    wpt->wpt_flags.shortname_is_synthetic = 1;
```
GPSBabelDeveloper, line 440: Now that we're not micro-managing free fs. points to const data, I'm not sure that shortname_is_synthetic makes sense any more. Maybe we can drop that completely.

route.cc
```diff
@@ -661,3 +404,140 @@ route_head::~route_head()
...
+// in the RouteList AND any routes that have had waypoints added but haven't been
+// added themselves yet.
+void
+RouteList::add_wpt(route_head* rte, Waypoint* wpt, bool synth, const QString& namepart, int number_digits)
+{
+  ENQUEUE_TAIL(&rte->waypoint_list, &wpt->Q);
+  rte->rte_waypt_ct++;	/* waypoints in this route */
+  ++waypt_ct;
+  if (synth && wpt->shortname.isEmpty()) {
+    wpt->shortname = QString().sprintf("%s%0*d", CSTRc(namepart), number_digits, waypt_ct);
```
GPSBabelDeveloper: Is this more readable as:
wpt->shortname = namepart + QString::number(waypt_ct).rightJustified(number_digits, '0');
?
Maybe
QString number =  QString(namepart + "%1").arg(waypt_ct, number_digits, 10, QChar('0'));
but the .arg isn't something well recognized outside of Qt.
Maybe not. I just dislike keeping another CSTR()

route.cc
```diff
@@ -661,3 +404,140 @@ route_head::~route_head()
...
+    *dst = new RouteList;
+  }
+
+  const char RPT[] = "RPT";
+  foreach (const route_head* rte_old, *this) {
+    route_head* rte_new = route_head_alloc();
+    rte_new->rte_name = rte_old->rte_name;
+    rte_new->rte_desc = rte_old->rte_desc;
+    rte_new->rte_urls = rte_old->rte_urls;
+    rte_new->fs = fs_chain_copy(rte_old->fs);
```
GPSBabelDeveloper, line 499: How much of our iterator pain is actually caused by the format-specific stuff? We only have a few modules that use it and we could make it more C++-like easier than we could protect ourselves from pointer abuse. in it.

GPSBabelDeveloper (approved):
This is a huge win for readibility. This is probably the last of the really funky (and intrinsic) data structures to tackle.
I don't think any of my comments are deal-breakers or even insisting on change; mostly things for us (to all) think through and if it were committed today unchanged, it'd still be a win for us.
Thank you for taking on the dark arts of the application of struct queue.

gtm.cc
tsteven4: If the condition is true then either
1) the input file hasn't been properly created, or
2) we don't understand the format.
The code below is now hardened against losing it's place in the file, it always reads ts_count * ( string + 12 bytes).  Since ts_count stored in the file I think we shouldn't read too many or too few bytes.

However, if the the condition is true, then we don't know how to line up the tracks and the track styles, and we are at risk of naming the tracks incorrectly.

With a warning we continue to read the file.  We could throw an error and give up, but the user might prefer something instead of nothing.  We could expand on the warning message and state "As a result the track names may be associated with the wrong tracks."

I think it is weird enough we want to know about it.

Suggestions?

route.cc
tsteven4: and QString::sprintf is obsolete.
I went with
wpt->shortname = QString("%1%2").arg(namepart).arg(waypt_ct, number_digits, 10, QChar('0'));

main.cc
tsteven4: I just copied those comments.  I don't know where they came from.  I will delete them.

queue.h
tsteven4: It's not just us, see qlist.h.

At this point class QList isn't used.  I left it in because it was a nice stepping stone in this PR, and I think it will be a nice stepping stone with a waypoint list.  Shortly after that queue.h and queue.cc go to deprecated/queue* or get removed.

route.cc
tsteven4: They can't be static because they are used by the template route_disp_all which gets included all over.

But we do have a bunch of global routines, e.g. route_add_head, ..., that just do one liners on global_route_list/global_track_list.  These could be eliminated in a massive edit that exposed global_route_list/global_track_list to the world instead.  That would  open up more of the power already in RouteList, e.g. iterators.  So far these aren't used on global_route_list/global_track_list because they are semi-hidden.  I'm not sure I want to do that, but I certainly don't want to do it in this PR.

gtm.cc
tsteven4: done.

gtm.cc
tsteven4: As I explained elsewhere the loop count should really be ts_count because the loop is used to read the entire track style section.  What you suggest, and the original code
`for (i = 0; i != ts_count && i != real_tr_count; i++) {`
risks not reading the entire track style section and getting out of synch for the following routes section.

gtm.cc
GPSBabelDeveloper: Convincing. We're potentially creating a slightly damaged file. We don't have a lot of GTM users represented, so we may not know.

GPSBabelDeveloper:
LGTM Merge at will.

route.cc
tsteven4: I agree.  Done.

PR #297, opened 2019-02-12
Relocate xcsv format functions.

tsteven4:
Many funtions for the xcsv format were in csv_util.cc
instead of xcsv.cc.  This commit moves the xcsv format
specific functions to xcsv.cc, and leaves the possibly
shared utility functions in csv_util.cc.

xcsv.h is added to communicate shared xcsv style handling
with vecs.cc.

This re-partitioning should allow some movement between the xcsv write init, write, and write deinit functions.  As it stands the prologue and epilogue are printed for every waypt in real time positioning.

This PR doesn't make changes other than re-partitioning except for the includes, some functions changed between static and non-static, and the utility functions in csv_util.cc now are not conditional on CSVFMTS_ENABLED, previously some where.


tsteven4:
We are thrashing a bit on generating the dependencies for Makefile.in.  I regenerated because of the addition of xcsv.h.  I generated on linux with "make dep", but the results are slightly different on macOS.  I could probably build and archive the dependencies on linux and macOS on travis if that would help.

GPSBabelDeveloper:
Lovely. I've been meaning to do this. Thanx.

PR #303, opened 2019-02-16
Parallel test

tsteven4:
with the configure flow and "make -j2 check-vtesto".
This will be utilized on travis.

If we weren't depending on GNU make we are with this.


robertlipe:
Valgrind is probably obscure enough that's OK. Greg from one of the BSDs
fusses at us when we use anything that's not in their make...even though
we've long since left POSIX and System V make-land.
Let's fight that if we have to. I've  wanted parallel valgrind for a long
time. Cool!

PR #308, opened 2019-02-23
update language processing for macos.

tsteven4:
merge relevant Qt .qm files into one qt_xx.qm file as
recommended by Qt.

deploy these meged .qm files with macos builds.
Since Qt5 we only deployed the meta catalog, but not it's
dependencies.  This should have led to the dependencies not
being available without Qt being installed.

whitespace corrections to locversion.plist skeleton.

eliminate double copying of our .qm files for mac. 

gpsbabel.qm and gpsbabelfe.qm are no longer included in the dmg, as it should be.

### DIFF
--- a/tools/convert_qt_translations
+++ b/tools/convert_qt_translations
@@ -1,0 +1,28 @@
+#!/bin/bash -e
+# It is recommended to combine Qts .qm files, this script does that for
+# linux and macos, windeployqt does this for windows.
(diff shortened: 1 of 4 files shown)


tsteven4:
I am refactoring this.  The translation file updates are needed, but in addition I am trying to bring all the gui packaging bits through the ages up to date and tie them into qmake.

tsteven4:
This has been superseded by #310.

PR #310, opened 2019-02-26
Refactor application packaging for linux/macos.

tsteven4:
Packaging of the GUI for linux/macos can now be done
either by building the make target "package" created by
"qmake app.pro", or by executing a bash script "package_app".
Packaging with the legacy configure flow will build the
package target.
Travis CI scripts will build the package target.

The obsolete script mac-localize is deleted.  This
functionality is included in package_app.

The obsolete script makelinuxdist.sh is deleted.  We have not
distributed a linux package in some time.  Furthermore this
scirpt hadn't worked in some time.  package_app does not
create a standalone linux app, however it packages all the
bits from GPSBabel in GPSBabelFE.  This makes creating a
Docker app, which includes the Qt installation, for GPSBabel easy.

Merge relevant Qt .qm files into one qt_xx.qm file for each language
as recommended by Qt.
Deploy these meged .qm files with linux/macos builds.
Since Qt5 we only deployed the meta catalog, but not it's
dependencies. This should have led to the dependencies not
being available without Qt being installed.

gpsbabel.qm and gpsbabelfe.qm are no longer included in the packaged
apps for windows/linux/macos, as it should be.


tsteven4:
I have downloaded and extracted the files from the dmg on linux, comparing them with a dmg before this change.  The differences are intentional and expected.  However, I don't have a macos machine to actually try the install on.  It would be best to try it on one without Qt installed, and using a locale that doesn't use the english language.  We support de, es, fr, it, hu, ru.  Some of the translations are from gpsbabel, and some are from Qt.  The close button translations come from Qt, the File translation in the upper left Input section comes from gpsbabel.  If both these are correct it demonstrates both translation files are being accessed.  I did this under linux with "LANG=fr_FR.utf8 ./gpsbabelfe".  I had the fr_FR.utf8 locale installed.

tsteven4:
Test dmg from travis, as shown in the log, is here
https://transfer.sh/y0GQN/GPSBabel-20190226T2013Z-9aef1c6.dmg

gui/app.pro
```diff
@@ -117,6 +118,9 @@ TRANSLATIONS += gpsbabelfe_hu.ts
 TRANSLATIONS += gpsbabelfe_it.ts
 TRANSLATIONS += gpsbabelfe.ts
 
-
-
-
+macx|linux{
```
GPSBabelDeveloper, line 121: We're now packaging QtPositioning, QtQml, and QtPrintSupport and a bunch of libraries that we haven't always packaged on Mac. We're up 30MB since our last release. Maybe Qt really does need them all now...it's easy to imagine that bringing in the webview brings in the kitchen sink.
I confirmed that the command line gpsbabel isn't blown up by debug info; the extra pounds are in Frameworks/

gui/package_app
```diff
@@ -0,0 +1,120 @@
...
+# which Qt modules we use.
+# In our case these are qtbase_*.qm, qtdeclarative_*qm and qtserialport_*.qm.
+#
+# Note with Qt5 the Qt distributed qt_xx.qm files are metacatalogs, and just
+# copying or converting them won't copy the dependencies.
+
+  if [ "${machine}" = "Mac" ]; then
+    resourcedir="${APPDIR}/Contents/Resources"
+
+    # caution, mktemp here is macos specific version
```
GPSBabelDeveloper, line 26: Historical note: mktemp was actually a Berkeley thing. The GNU coreutils team knew that it was handy enough they include it in coreutils, which is one of the more popular packages in Homebrew. Macs owned by "programmery types" will thus often have a GNU mktemp as both mktemp and gmktemp that are in directories that can be $PATHed before or after /usr/bin, where the native BSD mktemp lives. This basic battern follows for lots of things and it can be a bit of a landmine of "works on mine" type problems.
Fortunately, here we're just using -t which is the same in both GNU and BSD.

gui/package_app
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash -e
```
GPSBabelDeveloper, line 1: I think the BSD guys prefer "/bin/sh" and "set -e" instead. They probably don't use our packaging.

gui/package_app
```diff
@@ -0,0 +1,120 @@
...
+#
+# Note with Qt5 the Qt distributed qt_xx.qm files are metacatalogs, and just
+# copying or converting them won't copy the dependencies.
+
+  if [ "${machine}" = "Mac" ]; then
+    resourcedir="${APPDIR}/Contents/Resources"
+
+    # caution, mktemp here is macos specific version
+    resourceskel="$(mktemp -t gpsbabel_package_app)"
+    echo '<?xml version="1.0" encoding="UTF-8"?>' > "$resourceskel"
```
GPSBabelDeveloper, line 28: If this is code we ever have to look at again, a heredoc might be easier to read, but if this is code we don't touch for years, there's no value in changing it. Editing $LANGUAGE in might be easier. Testing this stuff is such a drag and we've had depressingly little feedback about translations.

gui/package_app
```diff
@@ -0,0 +1,120 @@
...
+MACDEPLOYQT="$(${QMAKE} -query QT_INSTALL_BINS)/macdeployqt"
+
+# update our translations and compile them.
+"${LUPDATE}" app.pro
+"${LRELEASE}" app.pro
+
+case "$(uname -s)" in
+  Linux*)     machine=Linux;;
+  Darwin*)    machine=Mac;;
+  *)          exit 1;;
```
GPSBabelDeveloper: echo a dying gasp, please.

GPSBabelDeveloper (approved):
Minor thinking aloud while looking through it, but LGTM. The size issue is something we can tackle - or not - in the future.

gui/package_app
tsteven4: done

gui/package_app
tsteven4: The skeleton had never changed since the first commit in 2010.

gui/app.pro
tsteven4: Qt seems to drag more stuff in all the time.  The Qt install is growing at an alarming rate:
/opt/Qt$ du -sm 5*
1013	5.12.1
381          5.9.7

tsteven4: Among others QtWebEngineWidgets pulls in QtPositioning, QtQml, QtPrintSupport.

ldd libQt5WebEngineWidgets.so | grep libQt
	libQt5WebEngineCore.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5WebEngineCore.so.5 (0x00007f1f5b87c000)
	libQt5Quick.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Quick.so.5 (0x00007f1f5b1e2000)
	libQt5PrintSupport.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5PrintSupport.so.5 (0x00007f1f5af70000)
	libQt5Widgets.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Widgets.so.5 (0x00007f1f5a724000)
	libQt5Gui.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Gui.so.5 (0x00007f1f59efe000)
	libQt5WebChannel.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5WebChannel.so.5 (0x00007f1f59ce0000)
	libQt5Qml.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Qml.so.5 (0x00007f1f59640000)
	libQt5Network.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Network.so.5 (0x00007f1f5929a000)
	libQt5Positioning.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Positioning.so.5 (0x00007f1f59005000)
	libQt5Core.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5Core.so.5 (0x00007f1f58872000)
	libQt5QuickWidgets.so.5 => /opt/Qt/5.12.1/gcc_64/lib/./libQt5QuickWidgets.so.5 (0x00007f1f58440000)

PR #312, opened 2019-03-03
clang-tidy fixes for the gui

tsteven4:
hicpp-use-equals-default
performance-unnecessary-value-param
readability-redundant-member-init
readability-container-size-empty
readability-delete-null-pointer
modernize-use-nullptr
readability-implicit-bool-conversion.
readability-static-accessed-through-instance
readability-braces-around-statements
readability-simplify-boolean-expr
hicpp-use-override
readability-named-parameter
manual fix suggested by clang-tidy modernize-use-using.
readability-else-after-return

### DIFF
--- a/gui/serial_unix.cc
+++ b/gui/serial_unix.cc
@@ -98,7 +98,7 @@ void MainWindow::osLoadDeviceNameCombos(QComboBox* box)
   const QStringList devices = dynamicDevices();
   box->addItems(devices);
 
-  for (int i=0; deviceNames[i]; i++) {
+  for (int i=0; deviceNames[i] != nullptr; i++) {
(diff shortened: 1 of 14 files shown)


gui/serial_unix.cc
```diff
@@ -90,15 +90,15 @@ static const char* deviceNames[] = {
   "/dev/ttyS3",
   "/dev/ttyUSB0",
   "/dev/rfcomm0",
-  0
+  nullptr
```
GPSBabelDeveloper, line 93: How is this code still here? Didn't we move to QSerialPortInfo::availablePort years ago? (I know this CL was cargo-culted and not introspecting the origins, so not an objection)

PR #313, opened 2019-03-04
clean up gui main and fix a png.

tsteven4:
- fix libpng warning: iCCP: known incorrect sRGB profile
  issue with runit.png with pngcrush.

- fix includes for gui/main.cc, including removal of
  CoreFoundation.h
- modify environmental varialble PATH  processing.
  - use toNativeSeparators.
  - use fromLocal8Bit on output of getenv.
  - use toLocal8Bit on input for putenv.
  - replace allocation of character array, conversion to
    standard string, and conversion to c_str with
    a local QByteArray and toLocal8Bit.
- use QApplication static methods instead of going directly
  to the underlying QCoreApplication static methods.


tsteven4:
I hung back from qputenv, qgetenv after reading https://lists.qt-project.org/pipermail/development/2015-December/023964.html
qEnvironmentVariable was introduced at Qt 5.10.

tsteven4:
This is surprisingly difficult if we want to get it correct everywhere including odd ball characters in paths, even more so with composite characters.  I don't think Qt 5.13 has all the bits necessary on windows, we need the equivalent of qEnvironmentVariable that does a put as well as a get.

gui/main.cc
```diff
@@ -52,15 +89,11 @@ int main(int argc, char** argv)
   QApplication app(argc, argv);
   QApplication::setWindowIcon(QIcon(":/images/appicon.png"));
 
-  QString newPath = "PATH=" + QApplication::applicationDirPath() +
-                    QString(pathSeparator) + getenv("PATH");
-  char* newPathEnv = new char[newPath.length() + 1];
-  strcpy(newPathEnv, newPath.toStdString().c_str());
-  putenv(newPathEnv);
+  set_path();
```
GPSBabelDeveloper: I wonder if we should quit trying so hard with OS specific code that's so fragile-looking. Can we look in the application dir and be done?
I think before we provided installers and encouraged people to use the GUI and the "real" app separately, we invited these issues, but it seems like anyone using the GUI that didn't do it through the installer should be able to work out broken PATH.
But I'm OK with it if it works and you're tired of thinking about this. :-)

GPSBabelDeveloper (commented):
OK

tsteven4:
If we are only trying to find gpsbabel maybe we don't need to mess with 
the PATH at all.  But if dlls/libs are found by the PATH we might need 
to set it. I'll explore but the only test platform I have for mac is travis.

tsteven4:
For windows it looks like the library search path has the first 
possibility as:

The directory from which the application loaded.

robertlipe:
I can check Mac later tonight. The libs are fixed to framework and rpath
relative locations in macdeployqt - that was an early need for it.
Translation libaries are found relative to applicationDirPath.

It doesn't seem right that we have to modify PATH in the GUI to find our
helper executable. Maybe we should put the helper executable in a different
place so it's found.

This is all way beyond the scope of what I'm sure you were hoping to byte
off here....

RJL

tsteven4:
These last 3 commits avoid our OS-specific fragile looking code that attempted to and usually successfully set the PATH.  Instead we pass the full path to gpsbabel to QProcess::start.  This involves ~100x more lines of OS-specific fragile looking code, but it is all inside Qt!  However, it't not our code and QProcess is much more solid and widely used that qgetenv/qputenv.

There are a couple of messages that are now out of date
`"Error reading format configuration.  "
 "Check that the backend program \"gpsbabel\" is properly installed "
 "and is in the current PATH\n\n"
 "This program cannot continue."`
I didn't change them because of the translations.

gui/mainwindow.cc
```diff
@@ -869,7 +895,7 @@ bool MainWindow::runGpsbabel(const QStringList& args, QString& errorString,
 {
   QProcess* proc = new QProcess(nullptr);
   QString name = "gpsbabel";
-  proc->start(name, args);
+  proc->start(QApplication::applicationDirPath() + '/' + name, args);
```
GPSBabelDeveloper, line 898: Consider refactoring this and the one above in 505 to use a simple one-liner that builds the full path to the executable in a single place before commit.

GPSBabelDeveloper (commented):
This eliminates lots of icky stuff. I like it. Please consider making it a single place where we find our own executable, but LGTM. I did verify on MacOS that gpsbabel and gpsbabelfe were siblings in GPSBabelFE.app/Contents/MacOS and that it wasn't buried in resources/

Issue #314, opened 2019-03-04
OZI: character encoding not respected


jdpiguet:
When using gpsbabel to filter OziExplorer track file, the non ASCII characters in track name are lost.
For example a track description "àéè" will be transformed in "???".

Tested on Debian 9 and Mint 19.1

jdp@amroth:/tmp$ gpsbabel -t -i ozi -f 20070813.plt -x track,pack -x simplify,length,error=0.2k -o ozi,pack -F 20070813_out.plt
jdp@amroth:/tmp$ head -5 20070813.plt
OziExplorer Track Point File Version 2.1
WGS 84
Altitude is in Feet
Reserved 3
0,2,255,V�zelay / Cuncy-l�s-Varzy          ,1,0,2,8421376,0
jdp@amroth:/tmp$ head -5 20070813_out.plt
OziExplorer Track Point File Version 2.1
WGS 84
Altitude is in Feet
Reserved 3
0,2,255,V?zelay / Cuncy-l?s-Varzy,0,0,2,8421376
jdp@amroth:/tmp$ gpsbabel -V

GPSBabel Version 1.5.4

jdp@amroth:/tmp$

tsteven4:
I believe this is fixed in the HEAD 
https://github.com/gpsbabel/gpsbabel.  On unix you will have to buid 
from source, for windows and macos there are pre-release builds at 
https://github.com/gpsbabel/gpsbabel/releases

Can you try it?  If not, can you supply a test case and I will verify if 
it is fixed.

jdpiguet:
Files used for:
`gpsbabel -t -i ozi -f 20070813_short.plt -x track,pack -x simplify,length,error=0.2k -o ozi,pack -F 20070813_out.plt
`
Check the line number 5...
Thanks and regards, Jacques-D.

[tracks.zip](https://github.com/gpsbabel/gpsbabel/files/2927372/tracks.zip)

jdpiguet:
Very strange: if I first convert the file to UTF-8 with:
`iconv --from-code=ISO_8859-1 --to-code=UTF-8 20070813_short.plt > 20070813_short_utf8.plt`
The result produced by gpsbabel is back to ISO-8859-1 ???

tsteven4:
This may be a current bug.  I'll look further.

tsteven4:
I once asked

> I am a maintainer of gpsbabel and we are trying to clean up character 
> encoding issues.  I see the page 
[...]

PR #315, opened 2019-03-06
convert ozi io to QTextStream

tsteven4:
and add an option to set the codec.
default the codec to windows-1252, which matches historic
usage but not recent behavior.

replace csv_lineparse.

fix a mistranslation with QString::arg.

enhance test for better coverage of routes and tracks.


ozi.cc
```diff
@@ -20,30 +20,63 @@
...
   int fgcolor;
   int bgcolor;
-} ozi_fsdata;
+};
 
+static struct {
+  gpsbabel::File* file{nullptr};
+  QTextStream* stream{nullptr};
+  QTextCodec* codec{nullptr};
+} ozi_file;
```
GPSBabelDeveloper, line 78: All these could become std::unique_ptr<foo> passed in during construction for more safety. Then when you delete ozi_file, they'd be auto-destructed.
I'm actually OK with this as a stepping stone, though.

GPSBabelDeveloper (approved):
I like how cleanly that swaps in.
Please add an xmldoc/formats/options/ entry for the code.

tsteven4:
This resolves #314.

PR #316, opened 2019-03-07
normalize connections in GUI

tsteven4:
as suggested by clazy.


GPSBabelDeveloper:
The whitespace seems inconsistent, but I'm kind of over arguing with tools
about such things. OK.

(I have a pile of email I oew you, Steve. I'm not ignoring you.)

tsteven4:
The lack of whitespace is an intentional and required part of the normalization.

PR #323, opened 2019-03-11
replace random queues with QLists.

tsteven4:


garmin_gpi.cc
```diff
@@ -781,8 +791,7 @@ compare_strings(const QString& s1, const QString& s2)
 static writer_data_t*
 wdata_alloc()
 {
-  writer_data_t* res = (writer_data_t*) xcalloc(1, sizeof(*res));
-  QUEUE_INIT(&res->Q);
+  auto res = new writer_data_t;
   waypt_init_bounds(&res->bds);
```
GPSBabelDeveloper, line 795: Some day, I think init_bounds() becomes part of the ctor for writer_data_t.

garmin_gpi.cc
```diff
@@ -792,10 +801,7 @@ wdata_alloc()
 static void
 wdata_free(writer_data_t* data)
```
GPSBabelDeveloper, line 802: ...and this becomes the dtor for writer_data about that same day. (I understand gravity and respect not getting sucked into it...)

garmin_gpi.cc
```diff
@@ -820,48 +826,45 @@ wdata_free(writer_data_t* data)
...
 
-  if ((data->ct <= WAYPOINTS_PER_BLOCK) ||
+  if ((data->waypt_list.size() <= WAYPOINTS_PER_BLOCK) ||
       /* avoid endless loop for points (more than WAYPOINTS_PER_BLOCK)
          at same coordinates */
       ((data->bds.min_lat >= data->bds.max_lat) && (data->bds.min_lon >= data->bds.max_lon))) {
-    if (data->ct > 1) {
-      sortqueue(&data->Q, compare_wpt_cb);
+    if (data->waypt_list.size() > 1) {
+      std::stable_sort(data->waypt_list.begin(), data->waypt_list.end(), compare_wpt_cb);
```
GPSBabelDeveloper, line 851: We can use std:: on a QList. Huh. I wouldn't have thought to try that. That's cool!

mapsource.cc
```diff
@@ -134,12 +142,9 @@ mps_wpt_q_deinit(queue* whichQueue)
...
  */
 static Waypoint*
-mps_find_wpt_q_by_name(const queue* whichQueue, const QString& name)
+mps_find_wpt_q_by_name(const QList<Waypoint *>* whichQueue, const QString& name)
 {
-  queue* elem, *tmp;
-
-  QUEUE_FOR_EACH(whichQueue, elem, tmp) {
-    Waypoint* waypointp = reinterpret_cast<Waypoint*>(elem);
+  foreach (Waypoint* waypointp, *whichQueue) {
```
GPSBabelDeveloper, line 147: This could turn into indexOf.

Or we could move mps and gdb into deprecate and never look at this code again. :-/

GPSBabelDeveloper (commented):
I'm still AFK, but skimming it, this is awesome.  I think this is much more accessible to a developer and WAY less fragile. Thanx.

garmin_gpi.cc
tsteven4: Yes, and it is official: https://doc.qt.io/qt-5/containers.html#introduction

> For traversing the items stored in a container, you can use one of two types of iterators: Java-style iterators and STL-style iterators. The Java-style iterators are easier to use and provide high-level functionality, whereas the STL-style iterators are slightly more efficient and can be used together with Qt's and STL's generic algorithms.

PR #325, opened 2019-03-13
fiddle with nmea waypoint status tracking.

tsteven4:
Traditionally nmea looks at the Waypoint data member Q
to decide if the Waypoint has been added on to a global list, and
if it has not it deletes the Waypoint so we don't leak memory.

This is problematic when we change the waypoint lists so queues
aren't used.

To overcome this difficulty nmea now uses a derived class of
Waypoint that tracks if the Waypoint has been added or not.

### DIFF
--- a/nmea.cc
+++ b/nmea.cc
@@ -155,6 +155,11 @@ time I have seen this is when the recording stops suddenly, where the last
 sentence is truncated - and missing part of the line, including the checksum.
 */
 
+struct NmeaWaypoint : Waypoint
+{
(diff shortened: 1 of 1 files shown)


tsteven4:
I thought I opened this a few days ago - but no.  Anyway I will merge it soon as it is now gating WaypointList (which is passing vtesto!).

GPSBabelDeveloper (commented):
Most of our uses of extra_data should probably eventually look like this.

Nice!

PR #326, opened 2019-03-14
 Introduce a WaypointList class to define the interface to our lists of waypoints.

tsteven4:
WaypointList is backed QList, eliminating the use of legacy queues for waypoint lists.


tsteven4:
I note there was some code marked 

> this test doesn't do what I want i.e test if this is a valid route - treat as a placeholder for now

in mapsource and psitrex.
I replaced these tests with if(true).

GPSBabelDeveloper:
This makes me very happy. This is what I envisioned with NEWQ and it eliminates a lot of really fragile, weird things. It makes things work much more normal and I like it a lot. Thanx!

...and if mapsource and psitrex cause any problems, I'd drop them in a heartbeat. The thing that replaced mps (gdb) has been replaced by basecamp has been dropped by the maker. mps is probably 15 years in the mirror and we have a lot of really fragile code for it.

PR #327, opened 2019-03-14
replace queues in mkshort.

tsteven4:
Also, hide the contents of mkshort_handle_imp.


GPSBabelDeveloper (approved):
An example from early in our life of OO  in C.

PR #328, opened 2019-03-16
retire queues.

tsteven4:


GPSBabelDeveloper (commented):
Hooray! Thank you.
Is it time to let the MSVC build files go?

tsteven4:
We could let the GPSBabel.vcxproj*  and GPSBabel.sln files go, but they 
don't bother me.  The *vcxproj* files are generated by "qmake -tp vc 
GPSBabel.pro".  I don't know how to generate the sln file with starting 
visual studio, and visual studio will create it from the vcxproj files 
anyway.  The most out of date file in msvc is the README, were I have 
only bothered to add 10 lines at the start that are up to date.  Anyway, 
I am certainly not promising to keep any of those up to date, but I 
don't feel compelled to delete them either.

robertlipe:
OK, good. As long as they're not generating work for us (you :-) ).

PR #329, opened 2019-03-16
update Qt 5.12 builds to 5.12.2.

tsteven4:
also add enough packages for docker builds so qmake can figure out libudev is available.


GPSBabelDeveloper:
We should eliminate the dependency on libudev.  With Qt5, all our gui/serial* collapses to like a half dozen lines of code that I know we've written. Let me dig for that.

PR #332, opened 2019-03-29
split csv lines as QStrings and return a list of all fields.

tsteven4:
supply csv_linesplit to process lines as QStrings and return a QStringList of all the fields.  This eliminates some oscillation between QStrings and char strings and eliminates a common memory leak when users of csv_lineparse fail to finish parsing a line leaving csv_lineparse holding dynamic memory.

add a RFC4180 based quoting strategy for csv, and use it for unicsv.  I left the historic strategy in place for other users.

add a textstream class to simplify and centralize use of file based textstreams.

convert unicsv to textstreams and csv_linesplit.  add option for specifying a codec with unicsv input and output.  print available codecs when an unrecognized codec is requested.   add a testcase to demonstrate that we can enquote and dequote a unicsv field and get the original field back (which we didn't previously do).

convert ozi to use textstreams and csv_linesplit.  convert main to use csv_linesplit.

other formats continue to use the legacy character based csv_lineparse: compegps, g7towin, garmin_txt, gopal, jtr, maggeo, netstumbler, nmn4, pocketfms_wp, stmsdf, tmpro.  If usages justifies the investment we can convert these as well.


ozi.cc
```diff
@@ -765,11 +743,7 @@ data_read()
   char* trk_name = nullptr;
   int linecount = 0;
 
-  while (true) {
-    buff = ozi_file.stream->readLine();
-    if (buff.isNull()) {
-      break;
-    }
+  while (buff = stream->readLine(), !buff.isNull()) {
```
GPSBabelDeveloper, line 746: Isn't this just while (buff = stream->readLine()) { ?Shouldn't the next return be null anyway?

GPSBabel.pro
```diff
@@ -152,6 +153,7 @@ HEADERS =  \
 	src/core/datetime.h \
 	src/core/file.h \
 	src/core/logging.h \
+  src/core/textstream.h \
```
GPSBabelDeveloper: Is there a tab here that's invisible to you?

ozi.cc
```diff
@@ -267,7 +245,7 @@ ozi_openfile(const QString& fname)
    */
 
   if (fname == "-") {
-    if (ozi_file.file == nullptr) {
+    if (stream == nullptr) {
```
GPSBabelDeveloper, line 248: I know this is mechanical, but I wonder if stream is ever !nullptr here.

unicsv.cc
```diff
@@ -1228,15 +1204,15 @@ unicsv_parse_one_line(char* ibuf)
...
+  QString buff;
 
   if (unicsv_fieldsep == nullptr) {
     return;
   }
 
-  while ((buff = gbfgetstr(fin))) {
-    buff = lrtrim(buff);
-    if ((*buff == '\0') || (*buff == '#')) {
+  while ((buff = fin->readLine(), !buff.isNull())) {
```
GPSBabelDeveloper, line 1213: As above.

GPSBabelDeveloper (approved):
Ambitious and a much needed change. Thank you. There are some cases where we should be using auto_ptr as our idiom, but when "just" doing mechanical transformations, it's hard to hold that as the bar.
Go for it.

GPSBabel.pro
tsteven4: expandtab got me again.  I fixed it and another line with the same issue.  We don't really need any tabs here at all in the pro file, but for consistency I put the tab on the textstream.h and xcsv.h lines.

ozi.cc
tsteven4: No.  buff is a QString, which cannot be automatically converted to a bool.

error: could not convert ‘buff.QString::operator=(QTextStream::readLine(qint64)(0))’ from ‘QString’ to ‘bool’

unicsv.cc
tsteven4: same as above, no auto conversion from QString to bool.

ozi.cc
tsteven4: yes, it can be depending on the number of waypoints, tracks and routes.

when fname == "-" then the first time here stream will be nullptr, but subsequent calls it will be != nullptr.

The first call possibility is from data_write() if we have one or more waypoints.
Subsequent call possibilities are from data_write() if we have tracks, one call per track header.
And a final possibility from data_write() if we have one or more routes.

robertlipe:
Well reasoned. Thanks for humoring me.

PR #342, opened 2019-04-21
hook GUI up to sort filter.

tsteven4:
This resolved an issue pointed out on Gpsbabel-misc by Tim Zenner.  Part of the issue was a bug in previous releases, and part was new in the 1.6.0 beta.

The Data Filters Waypoints widget offered to sort the waypoints, however if this checkbox was selected nothing happened.  The widget also did not ask how the waypoints should be sorted.  This is an old bug.  This option was removed.
Annotation 2019-04-21 060018

New to 1.6.0 is the ability of the sort filter to sort routes and tracks as well as waypoints.  These options can now be controlled on the Data Filters Miscellaneous widget.  Furthermore they actually cause the sort filter to be invoked.
Annotation 2019-04-21 045827

### DIFF
--- a/gui/filterdata.cc
+++ b/gui/filterdata.cc
@@ -210,5 +210,24 @@ QStringList MiscFltFilterData::makeOptionString()
     }
     args << s;
   }
+
+  if (sortWpt_ || sortRte_ || sortTrk_) {
+    args << "-x";
+    QString s = "sort";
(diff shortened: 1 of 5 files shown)


GPSBabelDeveloper (approved):
Wow. Good example of how a simple and obvious change can blow up on us. Thanx for the fix.

PR #344, opened 2019-04-24
Create One click converter ubuntu

SiRiusCb:
This script convert FIT files generated with Garmin or Bryton devices in GPX files and rename it using last modification time. Maybe it should be added to https://www.gpsbabel.org/osnotes.html page


GPSBabelDeveloper (commented):
Withe the specificity and the hardcoded paths, I'm not sure I want to add this in our repo, but it could probably be fluffed up into a page for https://www.gpsbabel.org/os/ I don't do Nautilus, so would you be interested in making a couple of paragraphs describing the general technique and offering a few examples like this?

PR #347, opened 2019-05-03
Create skeletal News

GPSBabelDeveloper:


gdt:
Looks good, with two nits:
 - there is an extra space at the end of each line
 - master has a file 'newnews', the existence of which doesn't make sense to me, as the point is to align with the longstanding GNU conventions that ~everyone knows, which will cause people to look in NEWS, similar to how they know to look in COPYING, AUTHORS, and README*.

PR #348, opened 2019-05-09
Fix gbfputcstr when using non UTF-8 encoding.

tsteven4:
This restores gbfputcstr from 75d8ec00f^.
That change could result in implicit conversion of
char* data that was specifically encoded as required for
a format to UTF-16, and then explicitly converted from UTF-16
to UTF-8.  While this was inefficient it could also
result in corruption when the data to be written was encoded
in something other than UTF-8.
For example, the Latin1 encoded 'ö' is '\xf6' will
be converted to the UTF-16 replacement character, and that will
be converter to the UTF-8 replacement character, not the original
character.

This resolves a mystery that was noted in the gdb test where some
strings where corrupted when writing gdb versions 1 and 2.

Instead of using function overloading I deleted the version of
gbfputcstr that accepted a QString to write, forcing the user
to specifically convert the data as appropriate.


jeeps/gpsmath.cc
```diff
@@ -249,7 +249,7 @@ double GPS_Math_Feet_To_Metres(double v)
 
 int32 GPS_Math_Deg_To_Semi(double v)
 {
-  return ((double)(1U<<31) / 180.0) * v;
+  return round(((double)(1U<<31) / 180.0) * v);
```
GPSBabelDeveloper, line 252: Wow, that's a good bonus find.

GPSBabelDeveloper (approved):
Awesome analysis and explanation. Thank you.

PR #354, opened 2019-05-15
update Free Software Foundation address.

tsteven4:


tsteven4:
All of the changes except strptime.h, strptime.c, COPYING, and COPYING.txt were done with the script:

> #!/bin/bash
for file in $@
do
perl -0777 -i.original -pe 's/59 Temple Place - Suite 330, Boston, MA 02111\n\/\/  USA/51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,\n\/\/  USA./igs' $file
perl -0777 -i.original -pe 's/59 Temple Place - Suite 330, Boston, MA 02111\n\#  USA/51 Franklin Street, Fifth Floor, Boston, MA  02110-1301,\n\#  USA./igs' $file
perl -0777 -i.original -pe 's/59 Temple Place - Suite 330,\n\*\* Boston, MA  02111-1307, USA./51 Franklin Street, Fifth Floor,\n\*\* Boston, MA  02110-1301, USA./igs' $file
perl -0777 -i.original -pe 's/59 Temple Place - Suite 330, Boston, MA 02111 USA/51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA./igs' $file
done

height.cc
```diff
@@ -17,7 +17,7 @@
 
     You should have received a copy of the GNU General Public License
     along with this program; if not, write to the Free Software
-    Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111 USA
+    Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
```
GPSBabelDeveloper, line 20: A bunch of these spill out of the 80 columns, but I'm not sure I care...

strptime.c
```diff
@@ -15,8 +15,8 @@
 
    You should have received a copy of the GNU Library General Public
    License along with the GNU C Library; see the file COPYING.LIB.  If not,
-   write to the Free Software Foundation, Inc., 59 Temple Place - Suite 330,
-   Boston, MA 02111-1307, USA.  */
+   write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
```
GPSBabelDeveloper: Since strptime.[ch] aren't ours[1] do we think this change is OK to make? I'd probably vote yes for the same reasons I'm OKing everything else, but we may lead to future ping-pong if we ever update them in the future.
[1] These should probably be moved to src/thirdparty/strptime (or something) for that very reason. Doing the same for jeeps, however, seems wrong because we've treated it as "ours" now for about 20 years and the author and original distributions are gone. I'd argue we ARE the source for jeeps now - absolutely a derived work, needing to keep credit and license, and such, but with no worry about ever picking up an refresh from upstream.

GPSBabelDeveloper (approved):
LGTM, but please consider reverting strptime.*
Rationale: improves fidelity with source license, doesn't change license terms or intent. This is just administrivia. (And nobody is sending letters to the FSF for copies of GPL these days....)

Thank you!

strptime.c
tsteven4: strptime reverted, but not moved.  future updates are already difficult because our reformatting with astyle was applied to strptime.

height.cc
tsteven4: Its a PITA to fix this due to multiple variations in the formatting of the text.

PR #363, opened 2019-05-23
Track filter fix

### LINKED ISSUE #357: Track Move Filter doesn't work from GUI
In 1.6.0 the track move filter doesn't work.  It passes the move value as \*d\*h\*m\*s which is no longer recognized by the command line tool.   This is a new bug in 1.6.0

### LINKED ISSUE #358: GUI Track Move Filter limits move time to 2000 days.
The GUI track move filter unnecessarily limits the number of days to move to +/-2000.  7168 days = 1024 weeks is of particular usefulness for the WNRO issue.

tsteven4:
Fix compatibility for track filter GUI<->CLI.

Add a week unit for the track filter move option,
in addition to the existing day, hour, minute and second units.

Allow combination of units for track filter move option, e.g.
1w2d3h4m-5s = 1 week + 2 days + 3 hours + 4 minutes - 5 seconds.
This matches historical use, and is expected by the GUI.

Add the week unit to the GUI track filter ui.

Change the limits on the track filter move specs:
week: +/-10000 (unit didn't exit before)
day,hour,minute,second: +/- 100000 (+/-2000 for day before, +/-100 for others before).

This resolves #357.
This resolves #358.

### DIFF
--- a/trackfilter.cc
+++ b/trackfilter.cc
@@ -78,34 +78,42 @@ int TrackFilter::trackfilter_opt_count()
 
 qint64 TrackFilter::trackfilter_parse_time_opt(const char* arg)
 {
-  qint64 result;
+  qint64 result = 0;
 
-  QRegularExpression re("^([+-]?\\d+)([dhms])$", QRegularExpression::CaseInsensitiveOption);
+  QRegularExpression re("^([+-]?\\d+)([wdhms])(?:([+-]?\\d+)([wdhms]))?(?:([+-]?\\d+)([wdhms]))?(?:([+-]?\\d+)([wdhms]))?(?:([+-]?\\d+)([wdhms]))?$", QRegularExpression::CaseInsensitiveOption);
   assert(re.isValid());
   QRegularExpressionMatch match = re.match(arg);
   if (match.hasMatch()) {
-    bool ok;
-    result = match.captured(1).toLong(&ok);
-    if (!ok) {
-      fatal(MYNAME "-time: invalid quantity in move option \"%s\"!\n", qPrintable(match.captured(1)));
-    }
+    int lastindex = match.lastCapturedIndex();
+    for (int idx = 1; idx < lastindex; idx += 2) {
+      bool ok;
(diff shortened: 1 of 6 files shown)


trackfilter.cc
```diff
@@ -78,34 +78,42 @@ int TrackFilter::trackfilter_opt_count()
 
 qint64 TrackFilter::trackfilter_parse_time_opt(const char* arg)
 {
-  qint64 result;
+  qint64 result = 0;
 
-  QRegularExpression re("^([+-]?\\d+)([dhms])$", QRegularExpression::CaseInsensitiveOption);
+  QRegularExpression re("^([+-]?\\d+)([wdhms])(?:([+-]?\\d+)([wdhms]))?(?:([+-]?\\d+)([wdhms]))?(?:([+-]?\\d+)([wdhms]))?(?:([+-]?\\d+)([wdhms]))?$", QRegularExpression::CaseInsensitiveOption);
```
GPSBabelDeveloper, line 83: I can't read this regex now, but can it ever match a unit into match[0] but not have a value so that the offset below will walk off the end of match? Maybe track,move=+1024   (no units)

Otherwise, lgtm. Thanx

GPSBabelDeveloper (approved):

trackfilter.cc
tsteven4: I don't think so.
1. match.captured(0) is the whole match.  we don't use 0.
2. we use pairs of match with indexes 1 & 2, 3 & 4, ...
3. since the first index in a pair is < the lastCapturedIndex we are guaranteed that
the second index in a pair is <= lastCapturedIndex.  So even if the re can capture matches not in pairs (ignoring 0) we won't index off the end.
4. the re is anchored by ^ and $ so we shouldn't get partial matches (with a few caveats).
5. the re is build of units of ([+-]?\\d+)([wdhms]).  The first group has to find one or more digits due to the + quantifier, and the second group has to match one of the supplied characters due to the implicit quantifier of 1.   So we either get 0 matches or 2 from each of these units.

PR #364, opened 2019-05-23
Gdb string read

tsteven4:
harden gdb reader to long string input.

fix bug in the gdb reader that resulted in file position being lost
if a string was truncated when reading due to the finite buffer size.

fix bug in the gdb reader that could result in data being written outside
the provided buffer.

enhance gdb readers to remove length limit on reading strings.

### DIFF
--- a/gbfile.cc
+++ b/gbfile.cc
@@ -20,13 +20,22 @@
 
  */
 
+#include <QtCore/QByteArray>   // for QByteArray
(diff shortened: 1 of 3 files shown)


tsteven4:
This is motivated by the file supplied by jobloman which was sent to gpsbabel-misc on May 5, 2019.

https://sourceforge.net/p/gpsbabel/mailman/message/36657932/

PR #365, opened 2019-05-27
fix potential memory leaks.

tsteven4:


GPSBabelDeveloper (approved):

tsteven4:
A method to find these is to add -DQT_ASCII_CAST_WARNINGS and
then look for xstrdup in the warnings.

PR #367, opened 2019-06-09
remove dependency on glob.

tsteven4:
and have magproto invoke the rd_deinit function when it uses the
gpx reader (uggh).


GPSBabelDeveloper (approved):
Yay . Nice cleanup. Thanks

PR #370, opened 2019-06-18
clion spell fixes, and a few cet retirements.

tsteven4:
cet retirements were unused.


GPSBabelDeveloper:
Wow. Some of those have probably been there for approximately forever. Thanx.
LGTM

PR #373, opened 2019-07-07
modify position filter algorithm.

tsteven4:
For routes and tracks, the algorithm is modified to
only remove consecutive points until the distance limit is
exceeded.  This prevents simulataneous removal of points on
an outgoing and return leg, which can lead to gross corruption
of the route/track shape.

A test case that demonstrates the problem with the previous
algorithm is added.


robertlipe:
Very nice. Thank you

PR #374, opened 2019-07-11
Fix g++-8 warnings in jeeps.

tsteven4:
Fix warnings:
jeeps/gpsread.cc:87:15: warning: unused variable ‘n’ [-Wunused-variable]
jeeps/gpssend.cc:137:71: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
jeeps/gpssend.cc:148:74: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]
jeeps/gpssend.cc:167:70: warning: comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare]

Note that test coverage of serial interface is minimal to non-existent, so manual review of this change is encouraged.

### DIFF
--- a/jeeps/gpsread.cc
+++ b/jeeps/gpsread.cc
@@ -84,7 +84,7 @@ int32 GPS_Serial_Packet_Read(gpsdevh* fd, GPS_PPacket* packet)
   start = GPS_Time_Now();
   GPS_Diag("Rx Data:");
   while (GPS_Time_Now() < start + GPS_TIME_OUT) {
-    if (int32 n = GPS_Serial_Chars_Ready(fd)) {
+    if (GPS_Serial_Chars_Ready(fd)) {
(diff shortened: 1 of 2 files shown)


robertlipe:
There was a time I sat out to provide enough of a mock/emulator to get some
automated coverage in Garmin-land, but it basically turned into building a
software-only Garmin and then mocking even the bugs in the Garmins of the
time.

I think we pencil-whip what we can safely judge, tread-lightly on busywork
cleanups for hyperactive warnings, and fix on fail.

The trap in this whole class of code is signed values used as errors and
then used (when not in error) as unsigned values. It's really easy to "fix"
warnings in that and then break the error paths. :-/

Approved.

PR #377, opened 2019-07-26
Don't pass MAKEFLAGS to $(MAKE)

jspricke:
make will pass them automatically. Also they are not meant to be
consumed by make as they are stripped:

Thus, if you do ‘make -ks’ then MAKEFLAGS gets the value ‘ks’.

https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -390,17 +390,17 @@ rpm: changes.html
 	tools/mkrpm $(WEB) $(VERSD) $(RELEASE)
 
 gui: gpsbabel$(EXEEXT)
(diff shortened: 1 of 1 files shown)


GPSBabelDeveloper:
Seems legit. Honestly, I rarely use the top-level Makefile for building the GUI (the only "real" child...which should probably be a sibling anyway) so this probably isn't well exercised. 

We'll be turning down 'configure; make' as the recommended path for most users anyway. I've restarted the build as it looks like it was sent to a stuck build agent. Once that succeeds, I'll fold it in.

Thanx.

PR #380, opened 2019-07-31
restore option of defining xcsv field separator as empty string,

r00t-:
see discussion in mailing list post:
https://sourceforge.net/p/gpsbabel/mailman/message/36729720/


r00t-:
closing, should use `tokenstr` instead of `tokens[0]`.
https://sourceforge.net/p/gpsbabel/mailman/message/36730259/
thanks to @tsteven4

PR #382, opened 2019-08-03
Remove unused minizip includes and targets

cjmayo:
HAVE_LIBMINIZIP was removed in:
ef579978 ("pull minizip from configure flow. (#339)", 2019-04-18)

---

Not sure if there is something else that should be changed to generate the Makefile.in changes, the section is headed "Machine generated from here down."

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -1047,11 +1047,6 @@ yahoo.o: yahoo.cc defs.h config.h zlib/zlib.h zlib/zconf.h cet.h \
 zlib/adler32.o: zlib/adler32.c zlib/zutil.h zlib/zlib.h zlib/zconf.h \
   config.h
 zlib/compress.o: zlib/compress.c zlib/zlib.h zlib/zconf.h config.h
-zlib/contrib/minizip/ioapi.o: zlib/contrib/minizip/ioapi.c \
-  zlib/contrib/minizip/ioapi.h zlib/zlib.h zlib/zconf.h config.h
-zlib/contrib/minizip/zip.o: zlib/contrib/minizip/zip.c zlib/zlib.h \
-  zlib/zconf.h config.h zlib/contrib/minizip/zip.h \
-  zlib/contrib/minizip/ioapi.h zlib/contrib/minizip/crypt.h
 zlib/crc32.o: zlib/crc32.c zlib/zutil.h zlib/zlib.h zlib/zconf.h config.h \
   zlib/crc32.h
 zlib/deflate.o: zlib/deflate.c zlib/deflate.h zlib/zutil.h zlib/zlib.h \
--- a/configure
+++ b/configure
@@ -6698,7 +6698,6 @@ done
 as_dir=jeeps; as_fn_mkdir_p
 as_dir=shapelib; as_fn_mkdir_p
 as_dir=src/core; as_fn_mkdir_p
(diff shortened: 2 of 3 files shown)


tsteven4:
Is this causing a problem?

ziparchive.cc and ziparchive.h are not compiled.  they were a work in progress that hasn't yet come to fruition, but they did get parked int the source tree, perhaps inappropriately.

I don't have a problem with the other changes, buts lets leave ziparchive.* alone.

cjmayo:
Well, it's causing work for no benefit (OK not much). Personally I wouldn't want to leave the code and the configuration out of step.

Gentoo is already carrying a patch to use system zlib:
https://github.com/gentoo/gentoo/blob/c01c2d2e17fabb61fdc0595996f08043aab2003e/sci-geosciences/gpsbabel/files/gpsbabel-1.5.4-use_system_zlib.patch

which needs updating for 1.6.0. I guess it is not a big deal to add removing the `#include "zlib/contrib/minizip/zip.h"`s in that too if they have to stay.

tsteven4:
For 1.6.0, instead of modifying gpsbabel-1.5.4-use_system_zlib.patch can you delete it and "configure --with-zlib=system"?

tsteven4:
Also for 1.6.0 I think you can drop 
gpsbabel-1.4.3-fix_binary_name.patch,  gpsbabel-1.5.4-qt-5.9-1.patch and gpsbabel-1.5.4-qt-5.9-2.patch.

Our CI is running with Qt 5.9 and Qt 5.12, no patches for Qt should be required.

cjmayo:
Dropped the ziparchive.* changes and repushed. I misunderstood and didn't realise they weren't being used at all.

Dropping the use_system_zlib.patch works, but only with the existing sed:
```sh
sed -i -e "s: zlib\/[a-z]*\.h::g" Makefile.in || die
```

tsteven4:
I see the issue is because you remove the local zlib and shapelib while 
many things that are still compiled depend on zlib.h, zconf.h.

I intended to add --with-shapelib to configure in the same fashion as 
--with-zlib.  I guess you will have the same issue with the shapefil.h 
dependency then, and would need to add another sed command.

Dropping the zlib patch seems like its an improvement, but it's too bad 
you still need sed.  We made an attempt to autogenerate the dependencies 
but we ran an issue that didn't seem worth fighting.  BTW "make dep" is 
how we generate the dependencies.  We usually do it on macos which 
results in slightly different results.

robertlipe:
There are two formats on our horizon that need minizip: ggz and kmz.

Neither of them are in active development to my knowledge, so if there's
some value in nuking them, I'm OK but there's some chance they'll be back
some day.

I'd committed ggz once, broke the world with it, then wasn't able to defend
it when it needed to be rolled back. It's just never been rolled forward
since. Honestly, demand for ggz fell, though those models continue to be
popular.

I won't fight to keep minizip; just explaining why we've been carrying it
in our source tree.

RJL

cjmayo:
I'm not too worried about this one now, it is only saving creating a temporary directory during build. I thought it was more important because Gentoo patches out the header includes in 1.5.4, but this is not needed for 1.6.0. 

If starting afresh I don't know if QuaZIP would make sense.

tsteven4:
All these autotools related files are obsolete and gone.  minizip related things are not part of the replacement system using qmake.

PR #384, opened 2019-08-04
Allow gpsbabelfe to be run from a Linux directory structure

cjmayo:
Enable data files to be stored separately from the executable.
Create a qmake variable PREFIX which if set is used to construct a
directory path PREFIX/share/gpsbabel that the translations subdirectory
and gmapbase.html are expected to be found in.

---

Would typically install the executable as `/usr/bin/gpsbabelfe` but with the data files in `/usr/share/gpsbabel/`.

A couple of additional tidy ups, not required for this but in the same area of code. Kept separate for clarity but happy to squash in if preferred.

### DIFF
--- a/gui/app.pro
+++ b/gui/app.pro
@@ -33,6 +33,10 @@ unix {
         DEFINES += HAVE_UDEV
         PKGCONFIG += libudev
     }
+    !isEmpty(PREFIX) {
+        PKGDATADIR = $$PREFIX/share/gpsbabel
+        DEFINES += PKGDATADIR=\\\"$$PKGDATADIR\\\"
+    }
 }
 
 UI_DIR = tmp
--- a/gui/mainwindow.cc
+++ b/gui/mainwindow.cc
@@ -222,7 +222,11 @@ MainWindow::MainWindow(QWidget* parent): QMainWindow(parent)
 
   ui_.outputWindow->setReadOnly(true);
 
+#ifdef PKGDATADIR
+  langPath_ = QLatin1String(PKGDATADIR);
+#else
   langPath_ = QApplication::applicationDirPath();
+#endif
   langPath_.append("/translations/");
 
   // Start up in the current system language.
--- a/gui/mainwindow.cc
+++ b/gui/mainwindow.cc
@@ -227,7 +227,7 @@ MainWindow::MainWindow(QWidget* parent): QMainWindow(parent)
 #else
   langPath_ = QApplication::applicationDirPath();
 #endif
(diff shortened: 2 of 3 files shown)


tsteven4:
Thanks.  This has been a sore spot with linux package maintainers.

Let's not assume gmapbase.html and the translation files are in the same directory.  For example, on fedora, 
/usr/share/gpsbabel/gmapbase.html
/usr/share/qt5/translations/gpsbabel_de.qm, ...
see https://koji.fedoraproject.org/koji/rpminfo?rpmID=18287661
They have a patch https://src.fedoraproject.org/rpms/gpsbabel/blob/master/f/0003-gpsbabel-1.4.3-gmapbase.patch that uses PKGDATADIR, and a patch https://src.fedoraproject.org/rpms/gpsbabel/blob/master/f/0008-Pickup-translations-from-usr-share-qt5-translations.patch that uses QTTRANSLATIONDIR.

Also, if you are a package maintainer I would be willing to review your patches and installation to avoid common miscommunications between gpsbabel and package maintainers.

One recent issue is that the gui uses both gpsbabel_\*.qm and gpsbabelfe_\*.qm files, as well as standard qt translation files.  Despite the naming of our qm files they are both used by the gpsbabelfe executable, and not by the gpsbabel executable.

If you make the package target it should place all the bits you need from us in one directory (GPSBabelFE on linux) where you can grab them from.  You should also be able to run gpsbabelfe from that directory without installation.

cjmayo:
My starting point was Gentoo's 1.5.4-r1, which does install the translations into /usr/share/gpsbabel:

https://github.com/gentoo/gentoo/blob/b761cec14174af726ba7429ff06595cae3e77d79/sci-geosciences/gpsbabel/gpsbabel-1.5.4-r1.ebuild#L110

Clearly this doesn't work, the code isn't modified to find them.

It's not just Gentoo, Arch has just packaged both into /usr/share/gpsbabel:
https://www.archlinux.org/packages/community/x86_64/gpsbabel/
(they seem to have missed those translations from coretools and patched out the map code)

And for a different example from Debian,  Scribus, also installing translations in a package directory:
https://packages.debian.org/buster/all/scribus-data/filelist


This patch doesn't make things any worse. The GPSBabelFE package would still be built as now, even Fedora saves one patch, albeit the other would need modifying.

Having said all that I can see how to make an executable that would work in the GPSBabelFE package and with /usr/share/qt5/translations/. See #385.

cjmayo:
Superseded by #386.

PR #385, opened 2019-08-05
Support storing GUI translations in a subdirectory or system location

cjmayo:
Find the translations using QLibraryInfo::TranslationsPath which will
point to the system location unless overriden by qt.conf.

The system location can be found using:
qmake -query QT_INSTALL_TRANSLATIONS

---

I've only tested the system location, not package_app.

### DIFF
--- a/gui/mainwindow.cc
+++ b/gui/mainwindow.cc
@@ -26,6 +26,7 @@
 #include <QtCore/QEvent>               // for QEvent (& QEvent::LanguageChange, QEvent::LocaleChange)
 #include <QtCore/QFile>                // for QFile
 #include <QtCore/QFileInfo>            // for QFileInfo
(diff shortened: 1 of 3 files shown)


tsteven4:
This is problematic if you don't want to have a qt.conf file next to the binary on linux, e.g. in /usr/bin, and you don't want to install the translations in the system location.  e.g. I believe gentoo installs the translations at /usr/share/gpsbabel/translations instead of the qt system location for translations.

cjmayo:
qt.conf can be included as a Qt resource. But I didn't expect it to be used when installing in /usr/bin. Rather people would switch to the qt location or provide their own patch.
And thanks for the explanatory notes in app.pro. Now I understand the purpose of concatenating the qm files. I'll certainly be proposing Gentoo use the qt location. Some Gentoo packages do already use that directory.

Superseded by #386.

PR #386, opened 2019-08-06
provide method to override file locations.

tsteven4:
linux installs typically install the translation files
and gmapbase.html out of the tree.  These operations are now
supported, see. gui/app.pro and look for PKGDATADIR, QTTRANSLATIONDIR.

### DIFF
--- a/gui/app.pro
+++ b/gui/app.pro
@@ -33,6 +33,30 @@ unix {
         DEFINES += HAVE_UDEV
         PKGCONFIG += libudev
     }
+
+    # To override the installed location of gmapbase.html set PKGDATADIR.
+    # e.g. qmake PKGDATADIR=/usr/share/gpsbabel
+    !isEmpty(PKGDATADIR):DEFINES += PKGDATADIR=\\\"$$PKGDATADIR\\\"
+
+    # To override the installed location of the translation files (*.qm)
+    # set QTTRANSLATIONDIR.
+    # Common use case 1:
+    # If the translations are installed local to the package then 
+    # gpsbabel_*.qm, gpsbabelfe_*.qm and the concatenated qt_*.qm must all
+    # be in QTTRANSLATIONDIR.
+    # .e.g. qmake QTTRANSLATIONDIR=/usr/share/gpsbabel/translations
+    # Note that the package_app target will create the concatenated qt_*.qm files.
+    # The concatenated qt_*.qm files are DIFFERENT from the
+    # the Qt provided meta catalog file of the same name, the concatenated
+    # qt_*.qm files include all the necessary Qt provided module qm files.
+    # Common use case 2:
+    # If the translations are installed in the directory that contains all the
+    # original Qt provided translations, then only gpsbabel_*.qm and
+    # gpsbabelfe_*.qm need be installed alongside the original Qt provided
+    # translations (which include the meta catalogs as well as the module
+    # qm files.)
+    # .e.g. qmake QTTRANSLATIONDIR=/usr/share/qt5/translations
+    !isEmpty(QTTRANSLATIONDIR):DEFINES += QTTRANSLATIONDIR=\\\"$$QTTRANSLATIONDIR\\\"
 }
 
 UI_DIR = tmp
--- a/gui/app.pro
+++ b/gui/app.pro
@@ -34,29 +34,57 @@ unix {
         PKGCONFIG += libudev
     }
 
-    # To override the installed location of gmapbase.html set PKGDATADIR.
-    # e.g. qmake PKGDATADIR=/usr/share/gpsbabel
-    !isEmpty(PKGDATADIR):DEFINES += PKGDATADIR=\\\"$$PKGDATADIR\\\"
+    # Setting PKGDATADIR will:
+    # 1. Look for gmapbase.html in PKGDATADIR instead
+    # of in QApplication::applicationDirPath().
+    # E.g. qmake PKGDATADIR=/usr/share/gpsbabel
+    # will use /usr/share/gpsbabel/gmapbase.html instead of
+    # QApplication::applicationDirPath() + "/gmapbase.html".
+    # 2. Substitude PKGDATADIR/translations in the search path for
+    # translation files instead of
+    # QApplication::applicationDirPath() + "/translations".
+    # E.g. qmake PKGDATADIR=/usr/share/gpsbabel
+    # will look in /usr/share/gpsbabel/translations instead of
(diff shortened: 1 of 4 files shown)


tsteven4:
I would prefer this to #385, #384 .  I have incorporated some of the ideas from these PRs.  @cjmayo will this work for you?

Note that in 1.6.0 to use use Qt translation meta catalogs as in use case 2 in app.pro you still need #360.

cjmayo:
I can work with that. Do take whatever ideas work. However, it does fail to compile for me with GCC 8.3.0:
```
map.cc:94:34: error: invalid operands of types ‘const char [20]’ and ‘const char [15]’ to binary ‘operator+’
   QString baseFile =  PKGDATADIR + "/gmapbase.html";
```

I replaced `PKGDATADIR` with `QStringLiteral(PKGDATADIR)` and it compiles (I think = is overloaded for QLatin1String but + is not, needs at least one QString)

Thanks for the tip on #360, I was hitting that issue but hadn't noticed.

tsteven4:
Which version of Qt gave you the compilation error?

cjmayo:
I'm using Qt 5.12.3.

cjmayo:
Looks like the only options are:


const QString | operator+(const QString &s1, const QString &s2)
-- | --
const QString | operator+(const QString &s1, const char *s2)
const QString | operator+(const char *s1, const QString &s2)
const QString | operator+(char ch, const QString &s)
const QString | operator+(const QString &s, char ch)

https://doc.qt.io/qt-5/qstring.html

tsteven4:
thanks.  I can reproduce the issue and I will fix it.  Nice to have near 
real time feedback!

cjmayo:
Looking at the code the same path is used to load the meta catalog and the application files:

https://github.com/gpsbabel/gpsbabel/blob/bfccb5759170180afd5c1b67b3e59048e646398f/gui/mainwindow.cc#L318

The approach shown in the Qt Internationalization documentation is to separate the two:
https://doc.qt.io/qt-5/internationalization.html#enabling-translation

I understand supporting the packaged form of the application but for the installed form I'd prefer if PKGDATADIR is passed to qmake:
* read the application files from PKGDATADIR
* use QLibraryInfo::location(QLibraryInfo::TranslationsPath) for the meta catalog

I can code that if there is any interest.

tsteven4:
This seems to give a bit different advice:

https://doc.qt.io/qt-5/linguist-programmers.html#deploying-translations

In any event, we could probably search both places for the qt_*.qm 
files.  Another solution is to have yet another variable that breaks out 
the location of the qt_*.qm files.  Then we have use case 1 and 2 
covered, and your proposed use case (3), app specific translation files 
here and qt translation files there.

A real problem with any of this is that it isn't easy for the packages 
to verify that all the "required" translation files are available.  The 
symptom is various strings show up in English instead of the local language.

tsteven4:
@cjmayo If I understand your desire it is to use the qt meta catalogs and qt module translation files in the location installed with your qttranslation package, and the application specific translations files in /usr/share/gpsbabel/translations.  Is this correct?

Note that Qt supports reassignment of the paths with qt.conf, but expects all the translations to be in the same place.  Linux packages have avoided using qt.conf, perhaps because they couldn't place a qt.conf next to the binary in /usr/bin.  But QLibraryInfo will first look for qt.conf in the resource file that is part of the executable.  So, if all the translation files, both from the application and from Qt were in one place, the qt.conf mechanism could be used to tell us that place.

cjmayo:
> This seems to give a bit different advice: https://doc.qt.io/qt-5/linguist-programmers.html#deploying-translations

I guess it is a matter of interpretation as ever. The second sentence there is:

> Usually, there are .qm files for the application, and, _if a version of Qt is used that is not installed on the system, Qt's .qm files need to be deployed as well._

I read that as if the Qt libraries are bundled with the application then the translation files need to be included too. So for macdeployqt, absolutely this needs to be the case.

I think code speaks much better than me, so I've has another go: #391.

cjmayo:
Haven't had a chance to run c7c1f21 yet. Looks like it covers all possibilities - I'm still not really convinced about the need for all of them. In an unscientific survey I looked at QGIS and Scribus and they seem to just use QLibraryInfo::TranslationsPath for the system translations.

But, ultimately I'm happy if /usr/share/gpsbabel/translations + /usr/share/qt5/translations works. Plus it's not like this is changing all the time so I don't see the time to search mattering.

tsteven4:
@cjmayo , I see #391.  I think your desire to have some of the translations in one place and some in another place is unusual.  QLibraryInfo assumes there is one path for translations and provides a mechanism to alter these (qt.conf, which can be a file or a resource bundled inside the executable).  Given that, every Qt application that wants to have multiple paths has to code that themselves, so I don't think you will find wide spread support for multiple paths.  But as in this PR and #391 it can be done one application at a time.

I feel like having a search path is a bit safer.  Not finding some of the translation files is quite a common bug in linux packages, so trying a bit harder seems a reasonable precaution to me.  Qt tries quite hard itself to get the filename correct making a whole series of transformations, but doesn't try to guess directory variations.

cjmayo:
I've tested c7c1f21 by rebasing #391 on top of this PR, after deleting the code changes and leaving the app.pro changes. That works fine for me.

I don't know what Linux bug reports you have had but I suggest providing a `make install` would reduce them because if it is provided people are more likely to use it instead of creating their own installation method (and they get a working Google map and desktop icon).

GPSBabelDeveloper:
Wow. You really have patience for this explosion of combinations.  I think this is the way to go (even if it's a bear to test in automation) and my understanding of all this - which is admittedly probably the early days of Qt 4 - matches that giant comment block. I DO remember back then that Qt on Mac had a nearly totally different handling of resources than on the others and that's where some of our "try ALL THE PLACES" code came from. Our other dent at the time was that Windows wouldn't let you change the language once the OS was installed - that's why we have FAKE_LANGUAGE_MENU. I had fits with it. 

Is that #else block for PKGDATADIR worthwhile? Should we just let PKGDATADIR be treated like being first in a $PATH with a sequential fallback to applicationDirPath() and then to TranslationsPath() ?

Earlier in this discussion, it was proposed to put something in qt5/Something. I know that's because Linux distros still have Qt4 and even Qt3 installed, but we should probably not encourage that if we can avoid it. We should see Qt6 by the end of the year. Once we're comfortable with that, we can shore up C++17 (move to system std::optional). So let's not put things in qt5/ now if we can avoid it.

tsteven4:
Closed in favor of #622

PR #387, opened 2019-08-06
add support for use of libshp to configure.

tsteven4:
added a new configure option --with-shapelib=(included)|system|no

the included shapelib bits are used by default.
a value of system searches for libshp
a value of no disables the shape format that depends on shapelib.

the configure option --disable-shapefile is removed, equivalent functionality can be had with --with-shapelib=no.

@cjmayo please see if this let's you remove your shapelib patch, although you will need to add sed editing of the dependencies like you do with zlib.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,11 +56,6 @@ set(ALL_FMTS
   energympro.cc mynav.cc ggv_bin.cc globalsat_sport.cc geojson.cc
 )
 
-#DEPRECATED_FMTS=cetus.cc copilot.cc gpspilot.cc magnav.cc psp.cc gcdb.cc quovadis.cc gpilots.cc geoniche.cc palmdoc.cc hsa_ndv.cc coastexp.cc pathaway.cc coto.cc msroute.cc mag_pdb.cc axim_gpb.cc delbin.cc google.cc
-
-#DEPRECATED_HEADERS=geo.h quovadis.h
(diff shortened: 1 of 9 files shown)


tsteven4:
@cjmayo I think your non-standard location for shapefil.h is going to prevent this from working without changing the #include in shape.cc.  Following our zlib example I would add
#if HAVE_LIBSHP
#include <shapefil.h>
#else
#include "shapefil.h"
#endif
but I don't think your install location for shapefil.h is in the system include path.

robertlipe:
I'm a bit chilly to this. configure/make isn't our future. adding
complexity for versions of these libs that aren't in our tree - and that
may give compile errors or worse, incompatible changes, seems a bit a bust
to me.

I'd almost rather the configure/make path be more thoroughly deprecated.

cjmayo:
Always the same tension between Linux distributions and application developers I guess.

Anway, I did try this but it also didn't work at the linking stage. To get it to work I used pkg-config and the PKG_CHECK_MODULES autoconf macro. Additional commit on top of this PR:

https://github.com/cjmayo/gpsbabel/commit/81a218818fb375c562a578cea251fb864b53403a

tsteven4:
This is being closed with the deprecation of autotools/configure.  It is supported with qmake.  See https://github.com/gpsbabel/gpsbabel/blob/master/INSTALL.  @cjmayo thanks for the pkg-config idea, it is part of the qmake implementation.

Issue #389, opened 2019-08-07
gpsbabelfe "Translation options" lights not shown


cjmayo:
If I run `gpsbabelfe` (with my default LANG environment variable, set to en_GB.utf8) I see the light symbols against the Waypoints, Routes and Tracks check boxes.

But if I run `LANG=de_DE gpsbabelfe` I just see "hyphens" in line with the bottom of the check boxes. Changing the Input Format using the combo box causes the lights to be displayed.

Starting values

Input Format: GPX XML
Output Format: Alan Map500 Waypoints and Routes (.wpr)

This is on GPSBabel 1.6.0 plus the patches for translations.
I don't see it on GPSBabel 1.5.4 - the Gentoo installation with translations in /usr/share/gpsbabel and without the code modified to load them.

tsteven4:
Thanks.  I can reproduce this.  Please open an issue on github.

tsteven4:
That was a great find, thanks @cjmayo for noticing and reporting it.

tsteven4:
This was fixed by #390

PR #390, opened 2019-08-08
fix interaction between translation and lights.

tsteven4:
When using a language other than English the initial painting of
the gui mainwindow showed the placeholder text for the labels next
to the waypoints, routes and tracks check boxes instead of the
lights.  This was issue #389.


tsteven4:
The QLabels that dynamically held QPixmaps containing the lights were setup in the UI to contain text.  This was probably done because the images are dynamically scaled and this option isn't available when creating the UI.  Apparently the translator messes with the text and that ends up being displayed initially instead of the dynamically set QPixmaps.

To fix this I changed the UI to contain a placeholder pixmaps instead of text, thus taking the translator out of the picture.  I temporarily used QPixmap::save to save a scaled image in mainwindow.cc.  Then I recolored this scaled image "convert 11scaled.png -colorspace Gray lights.png" and set this in the three QLabels using Qt Designer 5.12.4 on windows.  The rest of the UI changes were made automatically by Qt Designer.

tsteven4:
This fixes #389.

robertlipe:
Very interesting. I was about to ask for help, but that second message made
it all make sense. What a find! Thanx.

RJL

tsteven4:
This fixed #389

PR #391, opened 2019-08-08
[WIP] Support make install for GUI on Linux systems

cjmayo:
Worth some review but this is my ideal, requiring:
```
qmake PREFIX=/usr
make
make install
```
and you have a working GUI with translations, while hopefully not changing anything for package_app (untested).

I can get make to run lrelease by adding `CONFIG += lrelease` but can't see how to get the qm files created in .qm subdirectory installed.
https://bugreports.qt.io/browse/QTBUG-77398

### DIFF
--- a/gui/app.pro
+++ b/gui/app.pro
@@ -33,58 +33,6 @@ unix {
         DEFINES += HAVE_UDEV
         PKGCONFIG += libudev
     }
-
-    # Setting PKGDATADIR will:
-    # 1. Look for gmapbase.html in PKGDATADIR instead
-    # of in QApplication::applicationDirPath().
-    # E.g. qmake PKGDATADIR=/usr/share/gpsbabel
-    # will use /usr/share/gpsbabel/gmapbase.html instead of
-    # QApplication::applicationDirPath() + "/gmapbase.html".
-    # 2. Substitude PKGDATADIR/translations in the search path for
-    # translation files instead of
-    # QApplication::applicationDirPath() + "/translations".
-    # E.g. qmake PKGDATADIR=/usr/share/gpsbabel
-    # will look in /usr/share/gpsbabel/translations instead of
-    # QApplication::applicationDirPath() + "/translations".
-
-    # We search for translation files (*.qm) in this order:
-    # 1a. Relative to the executable, specifically at
-    # QApplication::applicationDirPath() + "/translations".
-    # This works when we package the app on windows and macos.
-    # It also works we create a bundled app on linux,
-    # for example with the package target.  However, linux packagers typically
-    # install the translation files in another location.
-    #  OR
-    # 1b. In PKGDATADIR/translations.  This works for linux packages that
-    # place the translations in PKGDATADIR/translations, e.g.
-    # qmake PKGDATADIR=/usr/share/gpsbabel with the translations in
-    # /usr/share/gpsbabel/translations.
-    # 2. In the Qt TranslationsPath.  If this is not overridden in qt.conf it
-    # will point to the hard-coded paths that are compiled into the Qt library.
-    # This hard-coded path can be found with "qmake -query QT_INSTALL_TRANSLATIONS".
-    # This works for linux packages that have some or all translation files
-    # installed in the original location they used when compiling Qt.
-
-    # There are three sets of translation files that should be available
-    # when running gpsbabelfe:
-    # 1. gpsbabelfe_*.qm
-    # 2. gpsbabel_*.qm
-    # 3a. The Qt supplied meta catalogs (qt_*.qm) and the module files they
-    #     refer to (qt*_*.qm).
-    #     Note if you are counting on finding the Qt supplied translations,
-    #     then the package that provides the Qt translations
-    #     needs to be a prerequisite to the gpsbabel gui package.
-    #  OR
-    # 3b. The concatentated translation files for the modules that gpsbabelfe
-    #     uses.  These files are created by building the package target.
-    #     These are named identically to the Qt translation meta catalogs, i.e.
-    #     qt_*.qm, but contain the necessary translation data for the modules
-    #     gpsbabelfe uses.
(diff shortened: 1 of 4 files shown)


GPSBabelDeveloper:
I'll defer to Steven for reviewing/integrating this since the two of you seem to have quite the groove going on in this area, but I can provide some historic perspective.

We used to support a 'make install' for Linux, but it was too messy because each distro wanted their Qt stuff in a different place. We were encouraged to not even try because everybody installs the version packaged by the OS anyway.

That may or may not be correct now (or ever) but that's how we got here.

cjmayo:
I don't want to cause more bug reports! I was thinking more about maintainers than users, hence it only being created if you set PREFIX (i.e. no default). Personally I prefer to have a make install, even if it puts things in the "wrong" place at least you can see what upstream intends to be installed (easier than going through the code).

But the main thing is to get the GUI working with translations and map when installed in a Linux system as per the other PRs. At least I found a Qt bug with this one.

tsteven4:
CONFIG+= lrelease was added in Qt 5.12 and the bug above was fixed in Qt 5.14.2.

cjmayo:
I've got Qt 5.14.2 installed now but hadn't realised that it contained the fix (thanks for pointing that out), so I will have a go at updating this PR.

tsteven4:
@cjmayo 
I think there are a couple of problems here:
1. I think the install file set varies with distribution.
2. I think some distributions create or alter files.
3. I think the directory structure is not common between distributions, e.g. some distributions put our translation files in /usr/share/qt5/translations, not /usr/share/gpsbabel/translations.
4. We have a floor of Qt 5.9 for gpsbabel.
5. some distributions may have their own methods of installing files?

So I am not confident that we can come up with an install target that works for all the distributions.  However, I still would like to make it easier so patching wasn't required to force the file locations.  I would also like to make it easier to make sure the critical files were distributed in some location.

I have been creating a file trying to capture what and where some distributions install things that I would like to share so you could add a column.  google docs?

tsteven4:
embedding gpsbabelfe_*.qm, qpsbabel_*.qm as resources is another option.  It would increase the size of the executable about 200kB.  lrelease.prf has an option to do this.  Then all the distributions have to do is have the Qt supplied qm files available and in a place we can find them.

tsteven4:
I'm leaning towards offering an option to embed the translation resources inside the executable.  I like the stuff @cjmayo found with CONFIG+=lrelease.  It has an option CONFIG+=embed_resources.  Unfortunately this was recently developed and requires a newer version of Qt than our floor.  However, I have implemented something like it directly in app.pro that should work without depending on the latest Qt enhancements/bug fixes.
 
So, @cjmayo , @jspricke would you be happy to have the executable increase 200kB and have the gspbabel translations compiled into the executable?  You would still need the qt supplied translations, but I think that just amounts to you adding a package dependency.

GPSBabelDeveloper:
5.9 is now an unsupported LTS release. Support ended May 2020, so we should
probably get away from that. 5.12LTS was new in Jan 2019 with 5.15 only in
this year and expected to be the last of Qt 5. After the Clank/Webkit split
several years ago, we really haven't had to have a very "wide" swath that
we've supported.

Consider a move to raise the floor to 5.12 pre-approved. 5.15 might be a
little more contentious but really things have been pretty quiet for us in
LTS land now for a couple of years. If Debian can solve some problem
relying on 5.15, that's OK with us. CentOS has always been the one that's
held us back, but I really don't think they're typical of our desktoppy
users that are following our development. Our CI pipeline now more closely
resembles real users, I think.

Distributions make all kinds of choices about icon/resource locations,
whether to minimize disk space, how to handle multiple installed versions
of us and everything we depend on, etc. so I'm really not sure how far down
the hole we can go with a universal 'make install' here that does anything
beyond the traditional "add it in the $PATH" version. We used to actively
encourage OSVs to put their packaging in our trees so we could help
centralize it, but that didn't go very far.

Actually, can you OSVs offer any help with the translations? When I added
them, I thought that would be key for world-wide dominance with non-coders
around the world helping out, but that also didn't play out.

jspricke:
>So, @cjmayo , @jspricke would you be happy to have the executable increase 200kB and have the gspbabel translations compiled into the executable?  You would still need the qt supplied translations, but I think that just amounts to you adding a package dependency.

Debian ships the GUI in a separate package so 200kB more should not 
matter.

>If Debian can solve some problem
>relying on 5.15, that's OK with us. 

The Debian packaging is already quiet clean (apart from the 
translations) so I don't really care and we currently ship Qt 5.14.2.

>Distributions make all kinds of choices about icon/resource locations,
>whether to minimize disk space, how to handle multiple installed versions
>of us and everything we depend on, etc. so I'm really not sure how far down
>the hole we can go with a universal 'make install' here that does anything
>beyond the traditional "add it in the $PATH" version.

You should define a default install layout and distributions have tools 
to adopt them to their requirements.

>We used to actively
>encourage OSVs to put their packaging in our trees so we could help
>centralize it, but that didn't go very far.

That's discouraged by the distributions cause you could only change the 
packaging with new upstream versions.

>Actually, can you OSVs offer any help with the translations? When I added
>them, I thought that would be key for world-wide dominance with non-coders
>around the world helping out, but that also didn't play out.

Ubuntu has a project and supports Qt translations:

https://wiki.ubuntu.com/Translations

Also Weblate works well for other projects I'm involved:

https://hosted.weblate.org/

cjmayo:
I'm fine with embedding resources and let's complete the job by including gmapbase.html.

My main concern with installation was for paths that had to be included in the code one way or another.  Files that are managed by the desktop environment can be resolved by the downstream packagers. Although if their installation is implemented in a way in which paths are configurable (which I think the patch to gui/app.pro includes) it shouldn't be too much of a burden beyond the initial setup and it also helps give them a pointer to what they should be installing.

tsteven4:
Closed in favor of #622

PR #394, opened 2019-08-10
store gmapbase.html as an internal resource.

tsteven4:
This removes all requirements about gmapbase.html from
any packaging efforts, it is no longer necessary to include it
in the package on all systems including Windows, Macos, and Linux.

### DIFF
--- a/gui/app.qrc
+++ b/gui/app.qrc
@@ -15,6 +15,7 @@
      <file>images/cancel.png</file>
      <file>images/reload.png</file>
      <file>images/appicon.png</file>
+     <file>gmapbase.html</file>
  </qresource>
  </RCC>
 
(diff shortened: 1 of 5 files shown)


tsteven4:
@cjmayo , I liked your QUrl::fromLocalFile, but I think this is much better.  Essentially gmapbase.html is compiled in now, so you don't have to do anything regarding gmapbase.html (set any options during qmake or make, patch source, or distribute the file, etc.).

Still working on translation files.

cjmayo:
I did try this and it worked fine for me. I just wondered if it would make development more hassle if you wanted to work on gmapbase.html so didn't suggest.

GPSBabelDeveloper:
I like it.
We really should look at usage metrics and see if the map is worth keeping. Requiring Webkit/Blink has a lot of overhead for us.

tsteven4:
Closed in favor of #622

PR #395, opened 2019-08-12
Implement FIT writing support

gromit1811:
Generate FIT courses from tracks+waypoints. Waypoints get mapped to course
points (generic by default, left/right turns if waypoint names contain the
string "left"/"right" or their equivalents in German, French, Spanish,
Italian).

If the original track contains neither time stamps nor speed values,
a fixed speed of 10 km/h gets assigned to the whole course.

Resulting FIT files have been tested and are accepted by a Garmin Edge 130
unit, but are assumed to be usable with any GPS unit supporting FIT.

Track segement markers are ignored at the moment.


garmin_fit.cc
```diff
@@ -769,6 +880,478 @@ fit_read()
...
+}
+
+static void
+fit_collect_trackpt(const Waypoint* waypointp)
+{
+  course.push_back(FitCourseRecordPoint(*waypointp, false));
+}
+
+static void
+fit_collect_track_tlr(const route_head *rte)
```
GPSBabelDeveloper, line 1206: Should we have a generic version of this in, say, route.cc? Recomputing this seems like something we have in a few formats.

GPSBabelDeveloper (changes requested):
Sorry this fell off my radar. This looks gnarly to have worked through, but the implementation seems straight-forward.

Can you please flush this with a test case in testo.d/garmin_fit that verifies as much of the new code as possible? Then we can pull it into the project and not be afraid of breaking it in the future.

Thanx

garmin_fit.cc
gromit1811: ACK. This is pretty similar to what we're doing for Garmin units supported by jeeps. I just don't have the slightest idea about how an API for this should look like. Do you have a function in mind which may be called by formats operating on courses instead of tracks and which returns a course for each track (or calls a callback for each course)?

gromit1811:
> Sorry this fell off my radar. This looks gnarly to have worked through, but the implementation seems straight-forward.

Yes, once you understand the concepts behind this strange format called FIT, it's actually not that difficult. Then the remaining problem is just to find out which of the optional data fields (and in FIT, essentially everything is optional) you need to send to make the device happy...

> Can you please flush this with a test case in testo.d/garmin_fit that verifies as much of the new code as possible? Then we can pull it into the project and not be afraid of breaking it in the future.

ACK, will try (I'm on holiday the next two weeks, so this will take some time). A bit of documentation probably wouldn't hurt either.

BTW, another thing on my roadmap is to try to automatically create course points for turns on the track. My idea is to check the derivative of the heading between track points (possibly interpolated) and to create a course point if it exceeds a certain threshold. However, this is not FIT-specific, so I guess this would better be placed in a separate filter. What do you think?

garmin_fit.cc
```diff
@@ -769,6 +880,478 @@ fit_read()
...
+  // it.
+  u8buf.append(len - u8buf.size(), '\0');
+  gbfwrite(u8buf.data(), len, 1, fout);
+}
+
+
+static void
+fit_write_position(double pos) {
+  if (pos >= -180 && pos < 180) {
+    gbfputint32((pos / 180.0) * 0x7fffffff, fout);
```
tsteven4: Please use GPS_Math_Deg_To_Semi instead.

gromit1811: Thanks for your feedback, I changed this as requested in the updated pull request.

gromit1811:
Updated (and rebased to current master). This adds a test case as requested plus a bit of documentation.

gromit1811:
Just noticed the failing test case, most likely caused by me rebasing the pull request without updating the reference file used in the test case. I'll take care of this and probably also improve the test case to check more FIT write features. Please stand by (may take another few days).

tsteven4:
Possible fixes for windows test failure and compile warning (https://ci.appveyor.com/project/gpsbabel/gpsbabel/builds/28246156) are in the following diff file:

[garmin_fit.txt](https://github.com/gpsbabel/gpsbabel/files/3751710/garmin_fit.txt)

gromit1811:
> Possible fixes for windows test failure and compile warning (https://ci.appveyor.com/project/gpsbabel/gpsbabel/builds/28246156) are in the following diff file:
> 
> [garmin_fit.txt](https://github.com/gpsbabel/gpsbabel/files/3751710/garmin_fit.txt)

Thanks, applied (even though I don't fully understand why EOF is considered to be an error on Windows)!

The updated pull request also improves the FIT write test case by also checking course point handling and speed calculation. Plus the odometer field of course points gets assigned a proper value. Finally, I removed some trailing whitespace.

Ready for merge from my side, assuming the CI tests succeed.

tsteven4:
I didn't fully understand the windows failure either. 

I think it may have to do with restrictions when switching from reading to writing or writing to reading (see fopen):

> In update mode ('+'), both input and output may be performed, but output cannot be followed by input without an intervening call to fflush, fseek, fsetpos or rewind, and input cannot be followed by output without an intervening call to fseek, fsetpos or rewind, unless the input operation encountered end of file. In update mode, implementations are permitted to use binary mode even when text mode is specified. 

The restrictions are honored in the "Update file header CRC" code.
These restrictions are NOT honored in the "Write file CRC" code.

So even though the code now works on windows I think we are not strictly meeting the usage restrictions of update mode.

tsteven4:
Indeed the windows failure was related to switching from writing to reading.  A patch to honor these restrictions, and make it obvious that we are honoring the reading to writing restrictions by reverting a few lines of the last patch is:

[garminfit2.txt](https://github.com/gpsbabel/gpsbabel/files/3753275/garminfit2.txt)

gromit1811:
> Indeed the windows failure was related to switching from writing to reading. A patch to honor these restrictions, and make it obvious that we are honoring the reading to writing restrictions by reverting a few lines of the last patch is:

Good catch, thanks! I updated the pull request with your patch. So now it finally should be ready for merge.

GPSBabelDeveloper (approved):
Thanx for the additional work. I'm cool with this. This is a lovely addition.

PR #400, opened 2019-09-06
Fix manpage xml

jspricke:


GPSBabelDeveloper (approved):
That whole doc probably needs eyes on it if we're going to keep it. I didn't think it was used anywhere.

PR #401, opened 2019-09-14
don't dump garmin_fit from lezyne devices just because occasion endian field is not 0 or 1.

spacelama:
The lezyne "super gps", every 10th ride or so, outputs a .fit file that is entirely valid other than a weird non 0/1 endian field.  Don't dump the data just because of that.

I've added a test file that errored out, but I have no idea whether I've formatted the test correctly.  Maybe you'd want a flag that specified that the checking was to be less strict.  Or maybe you'd just get rid of the warning message altogether since the file appears to be otherwise valid.


robertlipe:
Hi, and thanx.

It seems likely this flag just isn't endianness.  Big Endian seems to be
pretty dead in any computing environment (networking aside) so I'm OK with
this being a warning or demoted to even a debug message.

The tests are failing because there's an extra waypoint in your data that's
not accounted for. Can you please analyze, regenerating the reference files
if necessary, and re-upload?
You should be able to see this if you run a "./testo garmin_fit", but the
relevant part of https://travis-ci.org/gpsbabel/gpsbabel/jobs/584946886 is:

+  <wpt lat="-37.720647510" lon="144.964655098">
3994+    <name>LAP001</name>
3995+    <cmt>LAP001</cmt>
3996+    <desc>LAP001</desc>
3997+  </wpt>

spacelama:
OK, new reference file committed - the failed CI check above appears to be unrelated?

GPSBabelDeveloper:
Thank you for humoring me. Yes, the failure was spurious, so I retriggered the flake and we're good.

Thanx for the hand.

tsteven4:
In addition to the unusual endian fields, the reference file provided, lezyne_super_gps-garmin_fit-sample-bad-endian.fit, has a bad file CRC.  Setting the unusual endian fields to 1 does not result in a good file CRC.  The header CRC is present, non-zero, and correct.  It would appear that something besides the two unusual endian fields is corrupted in this file.

tsteven4:
lezyne_super_gps-garmin_fit-sample-bad-endian.gpx is quite corrupt.  The file is the expected length indicated in the header, but the last ~half of the file makes no sense.  This is in the area where the unusual endian fields are found.  This point is after all the recoverable data is recovered, i.e. even if we ~~fatal~~ abort the read on endian>1 the gpx output produced is identical.  Using -D8 shows we are quite lost after about half the file is read.

I question the desirability of this PR.

spacelama:
Correct, the second half of the file is corrupt, but the first half successfully contains the entire trip - the device in question then goes crazy perhaps as it loses GPS signal in the last few seconds before being told to shut down, but in the meantime writes half a megabyte of garbage.

But remember the original code bails at this point, not writing anything out.  That good data is not recoverable without some sort of patch saying "hey, ignore the CRC for now - I'll fix it in post-processing"

Also, even though current gpsbabel bails and doesn't write out any content, it isn't sufficient simply to write out the translated content so far, and then bail.  Quite a few files produced by this device have the corrupted content at the start of the file, and the second half contains good data.  

Better to leave the fixing up to post processing - just log a warning instead.

tsteven4:
@spacelama, thanks for writing back.

My goals, in priority order, are

- always produce valid output given valid input.  If you can prove that we produce invalid output given valid input that's our bug.
- process the portion of valid input that we understand.  We might not understand some feature of the input, but if the input is valid we strive to get by, perhaps skipping the parts of the input we don't understand.  The fit format has lots of data we don't understand that we should be able to step around, e.g. unknown message types.
   ...
- somewhere at the bottom of the list: tolerate invalid input when implementation is low risk and low complexity.
 

These goals apply to all formats.  Validity might be determined by XML schema validation (offline) against an XSD.  Invalid input/output might be indicated by a CRC failure.

All gpsbabel garmin fit code so far doesn't compute read CRCs.  ~~On my system the sample provided actually produces the same output if we fatal at the "Unusual endian ..." detection.  I can imagine this might depend on write buffering and might fail on another system as you seem to indicate.~~

I am about to propose a change that calculates the file CRC immediately before any other reading, and fatals if it is wrong.  It will also fatal on the header CRC if it is present and incorrect.  However, I will add an option to ignore the CRC errors and throw caution to the wind.  From a support perspective I want to know that we are processing a corrupt file so I won't waste time chasing down the exact circumstances of the failure.  In the case of your sample, after floundering around for a while we end up attempting to read past the end of file 4350 times because of other permissive behavior:

> // fail gracefully for GARMIN Edge 800 with newest firmware, seems to write a wrong record length
    // for the last record.

By the way, your sample file has exactly the length indicated in the header.  So even though about half the file has corrupted data the byte count is correct.  That may be a hint as to why the file is compromised.  The root of the problem you reported is upstream, that is where you should be seeking relief.

My leaning is to check and fatal immediately on any CRC error.  But if the ignorecrc option is provided then to process the file as master branch currently does.  Will this work for you?

spacelama:
On Sun, 26 Apr 2020, tsteven4 wrote:

> By the way, your sample file has exactly the length indicated in the header.  So even though about half the file has corrupted data the byte count is correct.  That may be a hint as to why the file is compromised.  The root of the problem you reported is upstream, that is where you should be seeking relief.

There's definitely an upstream bug - I have reported it, but the company
are pretty uninterested.

> My leaning is to check and fatal immediately on any CRC error.  But if the ignorecrc option is provided then to process the file as master branch currently does.  Will this work for you?

Sounds good to me.



-- 
Tim Connors

robertlipe:
Right on, tsteven4. That's a well stated list of our goals (that we should
probably actually formally state somewhere...) for all the formats for all
the many years we've been at this.

Adding a flag that is --i-know-my-gps-is-broken-and-dont-bother-us seems a
very reasonable compromise.

I'm sad that we've earned the reputation of repairpeople for broken
GPSes just because we're the only one that answers.  Just like we don't go
out of our way to squawk about unknown tax in GPX, we also don't crawl
through glass to "fix" files that are clearly dumb. I can think of one
exception, but that got fixed upstream at least partially because our
workaround generated a message to contact support at the company that wrote
it. :-) We no longer have that code.

Tim, my cycling days are over so I'll admit I hadn't even heard of
Lezyene units until this discussion. Can you please provide us with a
paragraph of models that are affected or do we have to assume that all of
their units (apps?) are similarly broken? Any words you can provide o this
topic that would help affected users can appear in our next doc build.

The "New Edge" firmware is from 2012. I have no idea if that 2.7 was fixed
or is still in circulation.  We tend to support things way longer than the
manufacturers do. (See above on reputation...) Unfortunately, people don't
contact us when kludges are no longer necessary.

spacelama:
> Tim, my cycling days are over so I'll admit I hadn't even heard of
Lezyene units until this discussion. Can you please provide us with a
paragraph of models that are affected or do we have to assume that all of
their units (apps?) are similarly broken? Any words you can provide o this
topic that would help affected users can appear in our next doc build.

At least the "Lezyne Enhanced Super GPS" is affected, but it could be the full range of devices using this firmware ("Year 10 and Newer Devices (2017-2019 Model Years)"): https://www.lezyne.com/gpsroot/gps_updategps.php

Perhaps the models affected are these: https://ride.lezyne.com/pages/user-manuals
2017: Mini, Macro, Micro, Super, Micro C
2019: Mega XL/Mega C

I last updated the firmware in about Jan 2020 (13.41), and had correspondence with them about the problem in Jun 2019.  There is a newer minor release out (13.88), but nothing relevant in the changelogs - there's been 4 major releases since the problem being apparent (in 9.64) and now.

tsteven4:
In general we get changes like this because somebody handy with a broken gps/serial interface/... got a fatal error trying to read.  They commented out the error or made it a warning and something useful got written.  It is sort of a miracle, because by commenting out the error the reader wanders around in never never land for quite a while before somehow completing and letting the writer write anything the reader read before it walked into the weeds.

IMO what we need is for the guts of the reader to throw an exception once it detects it is in the weeds.  Then at a high level in the reader, e.g. read(), the exception is caught and, depending on an option either fatals or issues a warning and returns from read().  This minimizes the wandering around in the weeds hoping to find our way out someday part of the story, and conditionally allows the writer to try to output any data that was recovered.

I will proceed in this direction for garmin_fit.

tsteven4:
@spacelama, I believe your sample file has three issues.  They are consistent with a lezyne software problem.

1. I believe the correct file should be the first 22183 bytes of your sample, with two changes in the header:
2. The length of the data section needs to be corrected in bytes 4-7. (22167)
3. This requires a new header CRC in bytes 12-13 (0x6c3f).

I note that the correct file CRC is already at bytes 22182-22183 of your original file.

It appears to me that your device failed to update the header after updating the data section and the file CRC.  Additionally the file length corresponded to the incorrect header.

PR #396, opened 2019-08-22
Replace libusb 0.1 usage with libusb 1.0

tsteven4:
Replace our modified libusb 0.1.12 with libusb 1.0.22 on macos.  On macos we statically link to our inclcuded libusb.
Use libusb 1.0 instead of libusb 0.1 on linux.  On linux we dynamically link to a distribution provided libusb, or optionally build without libusb support.

This fixes #381 
This resolves #25 as requested (albeit 3.5 years later)

PR #409, opened 2019-10-27
Remove remnants of NEW_STRINGS

robertlipe:
NEW_STRINGS haven't been new for many years at this point. I'm tired of looking at that code. Most of what's left is in paths never executed or tested anyway. :-/


tsteven4 (changes requested):
The file notes should not be part of this PR.
mac/libusb/Makefile should not be part of this PR.
mac/libusb/libusb-1.0a should not be part of this PR.

The xmldoc/chapters/build.xml change probably belongs in PR and conflicts with #410 .
The changes in xmldoc/format/options/miniHomer-gps-utc-offset.xml, xmldoc/formats/options/miniHomer-gps-week-rollover.xml would be more at home in #410 .

tsteven4 (changes requested):
xmldoc/chapters/build.xml changes are already merged into the HEAD, this shouldn't be part of this PR.

mac/libusb/libusb-1.0.a shouldn't be checked in, thus it shouldn't be part of this PR.  I did add it to the .gitignore file in #396

PR #419, opened 2019-11-06
make improvements for macos

tsteven4:
Correct macos bundle identifier using recommended reverse DNS format.

Build a bundle with the cmake flow on macos.

Delete gpsbabelfe script that messed with LD_LIBRARY_PATH and
then executed gpsbabelfe-bin. This file was obsolete, and
it's existence upset Xcode (entitlement file handling).


robertlipe:
Nice! I've always hated.that messy "fe" stuff.

PR #424, opened 2019-11-21
use Qt containers for format/style/filter vecs.

tsteven4:


tsteven4:
This can be slightly improved if we drop member parent from struct vecs_t.  It isn't used by the GUI, and I can't find any usage in the tools directory.

The output of "gpsbabel -^3" and "gpsbabel -%1" is identical to what it was before this PR.

tsteven4:
sort_and_unify_vecs still leaks memory, although not quite as much.  This could be fixed but it would required tracking which elements in the unified list were synthesized and which were not.  Currently all usages of sort_and_unify_vecs are shortly before gpsbabel exits.

robertlipe:
I've started this code a couple of times, so I welcome it.

inlining alpha() as a lambda is something that one of us should do when we're here next time. Maybe.

I don't much care that sort&unify leaks. We're about to exit and it's always in a one-shot wrapper for a GUI-like invocation. The long-term solution is to not have two separate, half-connected arrays for the xcsv derived and "real" formats. Some day, each file has a FooFormat :: Format that registers itself into the vector list and offers specializations for ReadInit, WriteInit(), Options(), and such. This is a good stepping stone to that.

PR #426, opened 2019-11-28
clang-tidy modernize-use-using

tsteven4:
replace typedef syntax.
this was done by hand.

there is also some cleanup of icon mapping tables wrt const.

### DIFF
--- a/an1.cc
+++ b/an1.cc
@@ -48,10 +48,10 @@ static double radius = 0.0;
 static long serial=10000;
 static long rtserial=1;
 
-typedef struct roadchange {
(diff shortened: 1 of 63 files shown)


tsteven4:
While this PR centers on items caught by modernize-use-using and involves removing typedef syntax, most of the changes do not introduce using syntax, rather they are:

- typedef struct [tag] {body} ref -> struct ref {body} and
- typedef enum [tag] {body} ref -> enum ref {body}.

Only the function pointer definitions involve deleting tyedef syntax and adding using syntax.

PR #433, opened 2019-12-09
Cleanups: make a gbfreadbuf, move units from defs and into dedicated file, reduce xstrdup,xmalloc use, reduce CString/QString roundtrips, more.

robertlipe:
I'd meant to have this in multiple submits, but here we are. Most of this is fairly mechanical. The KML change is worth a callout.

Our KML coordinate reader has been around the block. At various times,
it's been SAX, DOM, raw expat, expat front-ended for sane buffer management,
and is now a weird combination of C-level scanf and QString. I would have
guessed it to be faster than making two additional copies, the second of
which is parsed again after the first parse knocks down whitespace to make
the second level easy to parse for commas.

I'd have been very wrong.

With the existing code, parsing an arbitrary file of about 38,000
points takes about 9.8 seconds as measured by
$ time ./gpsbabel -t -i kml -f test.kml

With the new code, it's about .45 seconds.

It still passes validation. A 21x speedup that's also way simpler to read
isn't bad.

### DIFF
--- a/kml.cc
+++ b/kml.cc
@@ -417,11 +417,13 @@ void trk_coord(xg_string args, const QXmlStreamAttributes*)
     QStringList coords = vec.split(',');
     auto csize = coords.size();
     auto trkpt = new Waypoint;
-    if (csize >= 2) {
+
+    if (csize == 3) {
+      trkpt->altitude = coords[2].toDouble();
+    }
+    if (csize == 2 || csize == 3) {
       trkpt->latitude = coords[1].toDouble();
       trkpt->longitude = coords[0].toDouble();
-    } else if (csize >= 3) {
-      trkpt->altitude = coords[2].toDouble();
     } else {
       Warning() << MYNAME << ": malformed coordinates " << vec;
     }
--- a/kml.cc
+++ b/kml.cc
@@ -53,6 +53,7 @@
 #include "src/core/xmlstreamwriter.h"   // for XmlStreamWriter
 #include "src/core/xmltag.h"            // for xml_findfirst, xml_tag, fs_xml, xml_attribute, xml_findnext
 #include "xmlgeneric.h"                 // for cb_cdata, cb_end, cb_start, xg_callback, xg_string, xg_cb_type, xml_deinit, xml_ignore_tags, xml_init, xml_read, xg_tag_mapping
+#include "units.h"
 
 // options
 static char* opt_deficon = nullptr;
--- a/kml.cc
+++ b/kml.cc
@@ -413,9 +413,9 @@ void trk_coord(xg_string args, const QXmlStreamAttributes*)
   }
   track_add_head(trk_head);
 
-  auto vecs = args.simplified().split(' ');
+  const auto vecs = args.simplified().split(' ');
   for(const auto& vec : vecs) {
-    QStringList coords = vec.split(',');
+    const QStringList coords = vec.split(',');
     auto csize = coords.size();
     auto trkpt = new Waypoint;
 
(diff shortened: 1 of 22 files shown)


tsteven4:


```diff

Issues
======
+ Solved 5
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/gpsbabel/gpsbabel/pullRequest?prid=4637915&bid=15540345)

tsteven4:


```diff

Issues
======
+ Solved 5
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/gpsbabel/gpsbabel/pullRequest?prid=4637915&bid=15540345)

kml.cc
```diff
@@ -406,34 +406,28 @@ void wpt_icon(xg_string args, const QXmlStreamAttributes*)
...
-    // Line malformed or two-arg format without alt .  Rescan.
-    if (2 == n) {
-      sscanf(CSTRc(iargs), "%lf,%lf%n", &lon, &lat, &consumed);
-    }
 
-    if (3 == n) {
-      trkpt->altitude = alt;
-    }
+  auto vecs = args.simplified().split(' ');
+  for(const auto& vec : vecs) {
```
tsteven4: I recommend running clazy when you touch any Qt container, especially in a range based for loop:
qmake QMAKE_CXX=/path/to/clazy; make
You may want to export a modified CLAZY_CHECKS variable, but clazy-range-loop is probably enabled by default as it is included in the level 1 checks.

> kml.cc:416:3: warning: c++11 range-loop might detach Qt container (QStringList) [-Wclazy-range-loop]
  for(const auto& vec : vecs) {
  ^
                        qAsConst( )

The clazy suggested fix will work, but an alternative I prefer when possible is to make the container const:
`const auto vecs = args.simplified.split(' ');`

kml.cc
```diff
@@ -406,34 +406,28 @@ void wpt_icon(xg_string args, const QXmlStreamAttributes*)
...
+    QStringList coords = vec.split(',');
```
tsteven4: I recommend making Qt containers const when possible as it makes it clear you don't want to modify the container and  it prevents issues with detach.

kml.cc
```diff
@@ -406,34 +406,28 @@ void wpt_icon(xg_string args, const QXmlStreamAttributes*)
...
+    auto csize = coords.size();
+    auto trkpt = new Waypoint;
 
+    if (csize == 3) {
+      trkpt->altitude = coords[2].toDouble();
+    }
+    if (csize == 2 || csize == 3) {
+      trkpt->latitude = coords[1].toDouble();
+      trkpt->longitude = coords[0].toDouble();
```
tsteven4, line 427: For read access I prefer at() over operator[] because it will never cause a detach, .i.e.
`coords.at(n)`
etc.
See the discussion at https://doc.qt.io/qt-5/qlist.html#details
However if the list is const then they are equivalent.

kml.cc
robertlipe: Integrated. Thanx. It drives me crazy that 'auto' doesn't always return the "best" type.

Should we work on a clazy pass as a presubmit test?

kml.cc
robertlipe: Fixed as above.

kml.cc
robertlipe: They're now const.

tsteven4:


```diff

Issues
======
+ Solved 12
           
```


See the complete overview on [Codacy](https://app.codacy.com/app/gpsbabel/gpsbabel/pullRequest?prid=4637915&bid=15540345)

PR #434, opened 2019-12-09
tweaks for format vectors.

tsteven4:
flush out hiketech capability array.

use copy constructor of ff_vecs_t when creating unified vector list.


robertlipe (approved):
All this tooling that insists that these perfectly defined members have to have initializers really isn't much helping the maintainabiilty of that aspect of our code. Sigh. That SO needs designated initializers.
Good catch in getting a new ff_vecs_t instead of the initialized and dirty one.

PR #435, opened 2019-12-11
Delete file erroneously added in #409.

tsteven4:
This object library should not be under version control.  It was erroneously added in #409 .


robertlipe:
Rats. Thanx. That tree was a disaster where I accidentally merged the WIP
libusb work on top of my working tree and never really got them separated.

Still, NEW_STRINGS are finally no longer considered "new" ... after
something like six years.

PR #437, opened 2019-12-13
fix NMEA date fiddling.

tsteven4:
Three errors were corrected:
1) SECONDS_PER_DAY is assumed to be an integer type.  It was recently
errorously changed to to double.
2) when backfilling dates we need to subtract a day at rolloever,
not addit.
3) If we have a date, then it came with a time.  Don't set the backfill
time to 23:59:59.

Add test coverage for the above.


GPSBabelDeveloper:
Good catch. I'm surprised our operator overload for += didn't catch this,
but it's probably too specific on the time_t and missing the promotion of
the int (or whatever it it) to the time_t.

Now that creation_time is solidly a real time class, I wonder if it's time
to pull out that overload and do "real" time handling in anyplace that
still uses it.

PR #438, opened 2019-12-14
use QDebug in logging.

tsteven4:
This gives us full access to all the methods, stream operators and
manipulators that QDebug has access to.  There is a demonstration using nospace() and hex in nmea.cc.

It seems like we should be able to do this more directly with one of the other QDebug constructors, but I haven't found a way to flush when using those constructors.  Without the flush we lose output directed to Fatal().

I don't really like invoking exit in the destructor, but we don't want to throw an exception from a destructor.

### DIFF
--- a/nmea.cc
+++ b/nmea.cc
@@ -547,7 +547,7 @@ gpgga_parse(char* ibuf)
       waypt->fix = fix_pps;
       break;
     default:
-      Warning() << MYNAME << ": unknown vix value" << fix;
+      Warning() << MYNAME ": unknown fix value" << fix;
   }
 
   nmea_release_wpt(curr_waypt);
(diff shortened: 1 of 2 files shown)


nmea.cc
```diff
@@ -547,7 +547,7 @@ gpgga_parse(char* ibuf)
       waypt->fix = fix_pps;
       break;
     default:
-      Warning() << MYNAME << ": unknown vix value" << fix;
+      Warning() << MYNAME ": unknown fix value" << fix;
```
robertlipe: As discussed, a blank default is fine here.

src/core/logging.h
```diff
@@ -19,49 +19,34 @@
...
-#include <QtCore/QTextStream>
-#include <QtCore/QFile>
-#include <cstdlib> //exit()
+#include <QtCore/QDebug>       // for QDebug
+#include <QtCore/QFile>        // for QFile
+#include <QtCore/QIODevice>    // for QIODevice, QIODevice::WriteOnly
+#include <QtCore/QString>      // for QString
+#include <QtCore/QTextStream>  // for QTextStream
+#include <cstdio>              // for stderr
+#include <cstdlib>             // for exit
```
robertlipe, line 31: What do you use for generating these clean, sorted #include trees?

robertlipe (approved):
Nice harmony with qDebug.

src/core/logging.h
tsteven4: include-what-you-use.  But I almost always have to hand massage them.  The ordering conventions are different, but more importantly include-what-you-use often blows throw things like <ctime> or <cstdint> and tells you to use some underlying system dependent include file.
I also had to build custom mapping files for Qt, the defaults had a few issues.  I see they have updated their Qt imp file now, so maybe that isn't necessary.

PR #439, opened 2019-12-15
Datetime optimization

tsteven4:
Prefer Qt::UTC TimeSpec for storing creation_time.
This avoids very significant performance hits from converting
to/from Qt::LocalTime, which involve DST calculations as well
as the usual offsets.
Test suite with this commit runs at 63% of wall-clock time of the
baseline (time ./testo)!    lowranceusr test runs at 37% of wall-clock time compared to the baseline.

Delete obsolete operators and methods of DateTime.


tsteven4:
For the record I found "valgrind -callgrind" with kcachegrind for examining the results to much more enlightening than gprof.

robertlipe (approved):
I like this a lot. Go for it, please.

PR #449, opened 2019-12-29
Introduce Format class

tsteven4:
Also, introduce Vecs and FilterVecs classes.  These are implemented as Meyers singeltons.  This avoids some issues with initialization order.

All formats except gpx are used as LegacyFormats, which is deprecated.  The gpx format has been converted to demonstrate the preferred usage.


format.h
```diff
@@ -0,0 +1,221 @@
...
+  void set_name(QString nm)
+  {
+    name = nm;
+  }
+
+private:
+  QString name;
+
+protected:
+  template <class MyForamt>
```
robertlipe: s/MyForamt/MyFormat/g

main.cc
```diff
@@ -404,13 +399,13 @@ run(const char* prog_name)
       break;
     case 'x':
       optarg = FETCH_OPTARG;
-      filter = find_filter_vec(optarg);
+      filter = FilterVecs::Instance().find_filter_vec(optarg);
```
robertlipe, line 402: Could filter just have a scope within this case and the 'free_filter_vec' be the destructor?

vecs.h
```diff
@@ -0,0 +1,1293 @@
...
+   * the static vec that is used to construct a legacy format
+   * instance is guaranteed to have already constructed when an instance
+   * of this class is constructed.
+   */
+#if CSVFMTS_ENABLED
+  LegacyFormat xcsv_fmt {xcsv_vecs};
+#endif // CSVFMTS_ENABLED
+  LegacyFormat geo_fmt {geo_vecs};
+  GpxFormat gpx_fmt;
+  LegacyFormat mag_sfmt {mag_svecs};
```
robertlipe, line 245: Can the constructor of, say, geo_fmt be in geo.cc instead of here? This looks like a nearly circular include where everything indirectly includes vecs.h and any change to any format results in a recompilation of everything.
Even if each format has a global LegacyFormat geo_fmt{geo_vecs} (globals are bad. i know...) that registered the blahFormat, that would be one less file to modify/recompile when things are changed, right?

robertlipe (approved):
Since I've coached and worked with this change quite a bit, it's no surprise that I have nearly no input at this point.
I like the shape of this a lot and it really is how I imagined Formats working even before we were in C++. In my mind, a format has always been an object that could even be found and initialized at runtime via the -i/-o 
I have a work tree where I had a couple of FooFormats already moved from LegacyFormat, so please land this and I'll come back to tiger, unicsv, and v900. (I started at the end in CLion...)

format.h
tsteven4: thanks

main.cc
tsteven4: Perhaps, but I think that would work better if find_filter_vec was the filter constructor.  At this point our formats and filters rely on state being preserved between uses (e.g. stack filter) so we aren't ready for dynamic filters/formats.

vecs.h
tsteven4: I think the constructors could be local to each format, but that would require editing every format.  In either case there is a global, e.g. either geo_fmt or geo_vecs.

As it stands a change to a format only requires recompilation of that format.  A change to vecs.h or legacyformat.h requires a few files to be recompiled.  With your suggestion a change to legacyformat.h would require every format to be recompiled.

If we every go to dynamic formats then I don't think the constructor belongs in the format.

PR #450, opened 2020-01-01
Use https for www.gpsbabel.org

cjmayo:
This is a global grep and sed. I can't spot any issues, let's see how CI reacts.
Maybe it should exclude reference/ and just change user-visibile instances.


tsteven4:
Did this get flagged by something in your process?

Let's leave the translation files (*.ts) alone, they are tracking new and old source text to be translated as well as the translated text.  Of all these there is only one translation that needs an update, but that process is more complicated and can be handled separately.

tsteven4:
I manually updated the relevant translation in #451.

cjmayo:
> Did this get flagged by something in your process?

Nothing clever, just spotted one of the links and followed it and was automatically redirected to the https version.

Removed the translations.

tsteven4:
The changed urls are:
https://www.gpsbabel.org
https://www.gpsbabel.org/capabilities.html
https://www.gpsbabel.org/changes.html
https://www.gpsbabel.org/download.html
https://www.gpsbabel.org/formats/delorme-new-interstate/
https://www.gpsbabel.org/htmldoc-1.6.0
https://www.gpsbabel.org/htmldoc-development/GarminIcons.html
https://www.gpsbabel.org/htmldoc-development/Styles.html
https://www.gpsbabel.org/os/Linux_Hotplug.html
https://www.gpsbabel.org/osnotes.html
https://www.gpsbabel.org/style3.css
https://www.gpsbabel.org/upgrade_check.html

If http is used all these are redirected to https except upgrade_check.html which is doing a post.  It seems like either http or https works with these requests from the client side, but @gpsbabel should verify from the server side.  Note that http must work without redirect to support existing code.

http://www.gpsbabel.org/htmldoc-1.6.0 is redirected twice, first to https://www.gpsbabel.org/htmldoc-1.6.0, then to https://www.gpsbabel.org/htmldoc-1.6.0/.  But this test is misleading, the url is appended to before being used.  In at least one case the appended information is incorrect.  This is corrected in #452 

Subject to approval from @gpsbabel wrt the upgrade_check this PR looks good to me, it should avoid a bunch of 302 responses.

Thanks

tsteven4:
The upgrade check does seem to work with https when using testing=true in upgrade.cc.

robertlipe (approved):
LGTM++

GPSBabelDeveloper:
Of those, I think upgrade_check is the only one that's not basically handed
off to the internal browser.

It's also the one that caused me the most grief. "Change everything to
https", Google Webmaster Tools demanded. A few redirects later, done. It
was probably a full release later that I noticed that the upgrade check was
failing for everyone because that path didn't handle a redirect. Because
there is this long tail of people that use us infrequently but do still
update when offered, that URI gets special cased on the server to handle
both.

Meanwhile, today alone, we have several users (mostly in non-English
locales) that are running 1.4.0 and have pressed that decline button
hundreds of times. To that person that started 1.3.7-beta20100116 today
(!?!) - well, good luck.

At least we have the support load from people running older versions mostly
quietened...


OK, let's go with it. Thank you, Steven for the (always excellent) deep
dive.

tsteven4:
thanks Chris.

PR #453, opened 2020-01-06
Prevent negative len in ggv_bin_read_bytes.

ra1fh:
Both QByteArray::resize and QIODevice::read handle negative
len values gracefully, but there is no error status returned and
parsing fails later at and unrelated location. So better fail
early instead.

### DIFF
--- a/ggv_bin.cc
+++ b/ggv_bin.cc
@@ -39,6 +39,8 @@ static QString read_fname;
 static void
 ggv_bin_read_bytes(QDataStream& stream, QByteArray& buf, int len, const char* descr = nullptr)
 {
+  if (len < 0)
+    fatal(MYNAME ": Read error, negative len (%s)\n", descr ? descr : "");
   buf.resize(len);
   if (stream.readRawData(buf.data(), len) != len || stream.status() != QDataStream::Ok)
     fatal(MYNAME ": Read error (%s)\n", descr ? descr : "");
@@ -86,6 +88,12 @@ static void
 ggv_bin_read_text32(QDataStream& stream, QByteArray& buf, const char* descr = nullptr)
 {
   quint32 len = ggv_bin_read32(stream, descr);
+  // The following check prevents passing an unsigned int with a
+  // value greater than INT32_MAX to a signed int parameter in
+  // ggv_bin_read_bytes later on. If this happens, the file is
+  // almost certainly corrupted.
+  if (len > INT32_MAX)
+    fatal(MYNAME ": Read error, max len exceeded (%s)\n", descr ? descr : "");
   ggv_bin_read_bytes(stream, buf, len, descr);
   buf[len] = 0;
   if (global_opts.debug_level > 1)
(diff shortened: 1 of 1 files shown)


ggv_bin.cc
```diff
@@ -86,6 +88,8 @@ static void
 ggv_bin_read_text32(QDataStream& stream, QByteArray& buf, const char* descr = nullptr)
 {
   quint32 len = ggv_bin_read32(stream, descr);
+  if (len > INT_MAX)
```
robertlipe: Should we push this test into read32() itself? 
This probably fails in ILP64, but I don't know if any such architectures remain...

Are there any real-world cases where these conditions can happen and we could do better than a fatal() or is this basically damaged files that we'd be reading?

robertlipe (changes requested):

ggv_bin.cc
ra1fh: I wouldn't push the test into read32() as there are other places that use this function for values where the meaning and value range is not understood (and can be ignore for the purpose of extracting the geo data). Hence we might reject legitimate files if we push the check into read32().

You are right about the ILP64, this should rather be INT32_MAX. I will update the diff.

This check is added to catch damaged files. There are no real world uses cases. None of my 150 test samples triggers this case. So I have created a test case manually to check what would happen and noticed that there's only a warning:

QIODevice::read (QFile, "overflow.ovl"): Called with maxSize < 0

But no error status is returned, parsing continues at a wrong offset and fails shortly after one of the next values is rejected because it is out of range.

robertlipe:
>
> I wouldn't push the test into read32() as there are other places that use
> this function for values where the meaning and value range is not
> understood (and can be ignore for the purpose of extracting the geo data).
> Hence we might reject legitimate files if we push the
>
That seems a bit dangerous, but OK.

This check is added to catch damaged files. There are no real world uses
> cases. None of my 150 test samples triggers this case. So I


Please note that this is for damaged files and just something we don't
handle.



> But no error status is returned, parsing continues at a wrong offset and
> fails shortly after one of the next values is rejected because it is out of
> range.
>

Awesome. We sure don't have that number of files to test against.  Do you
have anything that increases code coverage in that source file enough to
make it worth integrating into our testo.d/ggv.test?  We don't hear much
from users of this format.

Minimally, add the comet that these are for damaged files and we'll be done.

Thanx
RJL

tsteven4:
codacy seems to have forgotten how to sort through our files to display the coverage report for ggv_bin.cc.  running "qmake; make coverage" locally shows no coverage for entry type 2 (lines 150-163) and entry type 9 (lines 202-209).

ra1fh:
> > I wouldn't push the test into read32() as there are other places that use this function

> That seems a bit dangerous, but OK

Yes I think that's the best thing to do here since only the caller really knows how to interpret the result. There are actually two callers that are potentially fine with the full 32bit value (not that I think any such ovl files exist with that large amount of entries).

Theres is another place where the result of read32() is passed to an int, I will add a check and comment there as well with the next revision of the change.

> > None of my 150 test samples triggers this case

> Awesome. We sure don't have that number of files to test against.
> Do you have anything that increases code coverage in that source file enough to
> make it worth integrating into our testo.d/ggv.test?

The test samples I have are the ones I found on the Internet when I developed the ggv_bin module. Some of those can still be found by putting the magic bytes into various search engines. They don't have any copyright attached, so I don't feel confident to add them to gpsbabel testo.d. The ones that are included are generated myself and one version 2 file contributed by a user that triggered a bug. With my copy of GeoGrid Viewer I can only generate version 3 files.

I'm currently working on the ggv_bin writer part. With that I will be able to generate some files that may increase coverage for both version 2 and version 3 files. I can verify correctness with GeoGrid Viewer and may add them to the test cases later on.

> We don't hear much from users of this format.

That either means the code is perfect or nobody besides me and the one user who reported a bug uses it. I guess the latter :-)

The map software that bundled GeoGrid Viewer was only used in Germany and Austria and has been discontinued some time ago. That reduces the user base quite a bit. There are still a few newer OVL files appearing in some places, but it's getting less since the software was discontinued.

GPSBabelDeveloper (approved):

PR #456, opened 2020-01-11
fix potential memory leaks in ggv_ovl.

tsteven4:


robertlipe (approved):
OK
Project for budding babel-head: invert these tests to be if (foo.isNull()) ...

tsteven4:
ok, you shamed me into a less myopic viewpoint.

PR #457, opened 2020-01-11
eliminate usage of obsolete QString::sprintf().

tsteven4:
in favor of the merely not recommended QString::asprintf().

### DIFF
--- a/an1.cc
+++ b/an1.cc
@@ -828,7 +828,7 @@ static void Read_AN1_Lines(gbfile* f)
       Waypoint* wpt_tmp = new Waypoint;
       wpt_tmp->latitude = DecodeOrd(vert->lat);
       wpt_tmp->longitude = -DecodeOrd(vert->lon);
-      wpt_tmp->shortname = QString().sprintf("\\%5.5lx", rtserial++);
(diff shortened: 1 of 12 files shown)


tsteven4:
This is a blanket replacement.  A recommended approach could be used for some of these, but that would have to be done by hand.

```
#!/bin/bash
for file in *.cc
do
 echo $file
 ed -s $file <<"EOJ"
1,$s/QString().sprintf(/QString::asprintf(/
w
q
EOJ
done
```

GPSBabelDeveloper:
Thanx. Some of those could be .arg() without a lot of pain.

We're probably in a race on whether Qt removes this at all/first, we
implement our own subclass of QString that has a method that calls
vasprintf, we just call vasprintf ourselves (ick), we embrace the streaming
techniques (double ick) or the "new" C++ {fmt} library comes to us either
via a third_party reference or via inclusion in the libraries,  If we have
a reason to dislike {fmt}, maybe we go Google's String::Substitute. Given
the noise of the new stuff in this area, wrapping vsprintf may well be a
permanent-ish move

The ugly cases  in the long term for replacement are in our style formats
where we expose sprintf syntax directly to users. A break in those formats
isn't absolutely out of the questions.

Such little changes make such big ripples...

RJL

PR #459, opened 2020-01-16
retire cet_convert_strings

tsteven4:
The only use of cet_convert_strings was for garmin specific data(gmsd).
gmsd is migrated from char strings to QStrings,
and any necessary conversions are handled within the formats using
gmsd.
This takes main out of the string conversion business except for init and deinit.

cet is still used by some formats, entitize in util, mkshort, and gbfile.


tsteven4:


```diff

Issues
======
+ Solved 18
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gpsbabel/gpsbabel/pullRequest?prid=4842697&bid=16077189)

GPSBabelDeveloper (approved):
This makes me happy to such an icky and historically problematic chunk of code go away. Go for it.

Some day we should rething that get/set FOO macro madness, but this is still soooo much better that I'll take it. Thank you

tsteven4:
I agree on the macro madness.  I think it is less mad than before, all the macros do now is consistently create the getters and setters which are real member functions.  Also, I reduced the variety of setters.
We could paste in the preprocessor output into garmin_fs.h to eliminate the macros, but that makes it hard to maintain the getters and setters.

robertlipe:
I'm rarely about letting perfect be the enemy of good. You consistently
make good calls.

Closer to home, "less pain" and "cured" aren't the same thing.

Those probably could have been templates, but that's just a different form
of madness. Are we papering over this code's C origins? GMSD is really just
a key/value store.. Maybe this needs to be a QMap<QString, QVariant>. We
never iterate over them. Set becomes insert(). Has becomes contains(). Get
becomes value(). Let me think more about it because this just feels like
we're using a hammer to tighten a nut here.

Don't let this get in the way of this commit. Go for it.

PR #460, opened 2020-01-17
clean up make icon doc.

tsteven4:
and fix an old capitalization error in the document.


GPSBabelDeveloper (commented):
The rules of this have changed of the years, IIRC. Is a large table in a header that may be included in multiple units cool or uncool these days? Would we ever get multiple copies in the same executable or compilers squawking about violations of the One Definition Rule?

This sets off alarms for me, though I don't recall the details right now. If you give it a second thought and declare it OK, go for it.

tsteven4:
The large table isn't included in multiple execution units of the same executable.
$ grep garmin_icon_tables *cc *h
garmin_tables.cc:#include "garmin_icon_tables.h"  // for garmin_icon_table, garmin_smart_icon_table
mkicondoc.cc:#include "garmin_icon_tables.h"  // for garmin_icon_table, garmin_smart_icon_table
garmin_tables.o is part of the gpsbabel executable, mkicondoc.o is part of the rarely if ever run mkicondoc executable.

robertlipe:
Compelling. Sold.

This really is a big mess from the days when people cared about Garmin
icons and Garmin was changing their numbering scheme every couple of
quarters. It's been quiet for a long time.

PR #462, opened 2020-01-18
Convert ggv_bin format to sub-class of Format

ra1fh:


tsteven4:
@ra1fh I am very pleased you discovered the new Format class and choose to convert ggv_bin. :clap:

I think all the static functions in ggv_bin.cc should become private member functions of GgvBinFormat.  There are a few non-static non-member functions in gpx.cc that are shared (and don't really belong in gpx.cc), but the functions that are just pertinent to GpxFormat are private member functions instead of static functions.

1. This requires changing them in ggv_bin.cc from "static T name" to "T GgvBinFormat::name".  This will produce a bunch of compile errors.  If there were other non-static, non-shared functions they should probably have been static and should be converted as well.
2. With g++ you can get the basic declarations needed from the error messages with
`make ggv_bin 2>&1 | grep "member function declared" | sed 's/’.*/;/' | sed 's/.*‘//' | sed 's/GgvBinFormat:://'`
Paste that into ggv_bin.h in a place such that they are private.
3. Now this should compile, but clang-tidy will whine about readability-named-parameter.  Assuming you created your Makefile with qmake:
`make compile_commands.json`
`clang-tidy --header-filter=.* --checks=readability-named-parameter ggv_bin.cc --fix`
Make sure clang-tidy didn't fix any other headers, revert any extraneous changes (git checkout -- leavemealone.h).
This results in the parameter names being present as comments in the declaration.
4. Our historical style is to have the parameter names un-commented in the declarations. For consistency I removed the "/\*" and "\*/" comment delimiters produced by clang-tidy by hand.
5. It is possible that some of the newly created member functions could be static.  You can find any of these with the clang-tidy check readability-convert-member-functions-to-static.

I won't insist on step 4, but I would like steps 1-2 to make the static functions private member functions, and step 3 to appease clang-tidy.  I think the readability-named-parameter check is good one, and assuming the parameters have reasonable names, adds understanding.

I actually did these changes to generate the above recommendations.  Here is a patch:

[memberfunc.txt](https://github.com/gpsbabel/gpsbabel/files/4081786/memberfunc.txt)

In the future can you please avoid the forced pushes?  It makes it harder to see the evolution of the pull request on github.  You can just push additional commits to your branch (ra1fh:ggv-bin-convert-to-format) and they will be added to the PR.  If I merge the PR I will very likely squash it so the detailed commit history of the branch doesn't get carried into the history of gpsbabel:master.  Perhaps this was an necessity of the way you did the merge with sha sums?   I follow [https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork](url).  After those steps you can sync your branch by`cd yourbranchname; git merge master`.  This will generate a more understandable merge commit message like "Merge branch 'master' into yourbranchname" instead of a message that just references the shas.  Or was there an issue you were trying to overcome with the forced push that I haven't comprehended?

tsteven4 (changes requested):
See my comment above.

ra1fh:
Hi, thanks for the review.

Initially I did add all the static helper functions in ggv_bin.cc as static class members, but then decided to remove them again because they don't need to be part of the class interface and they do not require access to any class member variables. This has the advantage that compilation units that include the header do not need to be recompiled whenever helper functions are added/removed or functions signatures change. But I'm happy to change that again.

This particular force push was a rebase onto master that I did to get another CI run with latest changes from master. I agree that's kind of hard to see when clicking the force-push link, since it contains all the tree differences from the rebase, plus my own changes (although my patch did not change between the two trees). I could have done a merge as well in this case and will do so in the future. Would simple force pushes without rebase still be ok? For those, the changeset evolution is nicely visible by clicking on the "force-push" link in the PR.

ggv_bin.h
```diff
@@ -0,0 +1,79 @@
...
+
+private:
+  static void ggv_bin_read_bytes(QDataStream& stream, QByteArray& buf, int len, const char* descr = nullptr);
+  static quint16 ggv_bin_read16(QDataStream& stream, const char* descr = nullptr);
+  static quint32 ggv_bin_read32(QDataStream& stream, const char* descr = nullptr);
+  static void ggv_bin_read_text16(QDataStream& stream, QByteArray& buf, const char* descr = nullptr);
+  static void ggv_bin_read_text32(QDataStream& stream, QByteArray& buf, const char* descr = nullptr);
+  static double ggv_bin_read_double(QDataStream& stream, const char* descr = nullptr);
+  static void ggv_bin_read_v2(QDataStream& stream);
+  static void ggv_bin_read_v34_header(QDataStream& stream, quint32& number_labels, quint32 &number_records);
```
tsteven4: Our style has been to ram the ampersand next to the type, i.e "quint32 &number_records" -> "quint32& number_records"

tsteven4 (commented):
Thanks.   I just saw one little nit.

I realize there is a choice with static helper functions vs. static class members.  We've been considering a style guide like https://google.github.io/styleguide/cppguide.html#Nonmember,_Static_Member,_and_Global_Functions but haven't got there yet.  In any event I like the current implementation.

I didn't even notice that forced-push was a link.  You are correct, in this case it is not pleasant to look at.  I am not sure about simple force pushes without rebase. I think your the first one to have a forced push in a PR.

Again, thanks for taking this on and supporting this format.  It's also nice to have somebody with expertise in OS variants we are unfamiliar with.

ggv_bin.h
ra1fh: Just found astylerc and applied it to both .h and .cc. That's much easier than sensing the preferred coding style from surrounding code.

tsteven4 (approved):

PR #464, opened 2020-01-19
Add OpenBSD support to qmake project

ra1fh:
OpenBSD provides libusb-1.0 and can mostly be treated like Linux. With exception of LINUX_HID, which is not available on OpenBSD.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,8 +219,11 @@ if(MSVC)
   add_compile_options(/MP -wd4100)
 endif()
 
-if (UNIX AND NOT APPLE)
+if (CMAKE_SYSTEM_NAME MATCHES "Linux")
   add_definitions(-DHAVE_LINUX_HID)
+endif()
+
+if (UNIX AND NOT APPLE)
   set(LIBS ${LIBS} usb-1.0)
 endif()
 
--- a/GPSBabel.pro
+++ b/GPSBabel.pro
@@ -172,7 +172,7 @@ load(configure)
 
 CONFIG(release, debug|release): DEFINES *= NDEBUG
 
-macx|linux {
+macx|linux|openbsd {
   qtCompileTest(unistd) {
     # this is used by zlib
     DEFINES += HAVE_UNISTD_H
@@ -207,6 +207,9 @@ win32-msvc* {
 
 linux {
   DEFINES += HAVE_LINUX_HID
+}
+
+linux|openbsd {
   LIBS += "-lusb-1.0"
 }
 
@@ -249,14 +252,14 @@ DEFINES += CSVFMTS_ENABLED
 QMAKE_CFLAGS_WARN_ON -= -W
 QMAKE_CXXFLAGS_WARN_ON -= -W
 
-macx|linux{
+macx|linux|openbsd{
   check.commands = PNAME=./$(TARGET) ./testo
   check.depends = $(TARGET)
   QMAKE_EXTRA_TARGETS += check
 }
 
 # build the compilation data base used by clang tools including clang-tidy.
-macx|linux{
+macx|linux|openbsd{
   compile_command_database.target = compile_commands.json
   compile_command_database.commands = make clean; bear make
   QMAKE_EXTRA_TARGETS += compile_command_database


tsteven4 (commented):
Thanks.  It's nice to have participation from somebody running OpenBSD which isn't represented in our continuous integration process.

From your PR I gather  the linux scope doesn't include openbsd.  If so we need to include openbsd in GPSBabel.pro (as you have done) and gui/app.pro.  CMakelists.txt probably needs a similar change.  Perhaps instead of openbsd we should use the more generic bsd scope?  We probably wouldn't want that to trigger on any darwin/macos/macosx systems.  It isn't clear to me what the tree of Qt scopes looks like.

However, it looks like we don't use HAVE_LINUX_HID anywhere except:
-  ./deprecated/delbin.cc
-  ./CMakeLists.txt
-  ./GPSBabel.pro
-  ./configure
-  ./configure.ac
-  ./Makefile
-  ./config.h.in

It appears to me we should have ripped out HAVE_LINUX_HID from our build process when we moved to libusb-1.0, leaving delbin.cc, which is deprecated, alone.

We do check in the generated file configure.  I would like to use GNU Autoconf 2.69 to regenerate it.  This matches the version on systems that we usually use to generate it.

We could rip out  HAVE_LINUX_HID out in a subsequent PR, or you can add that to this PR.   Let me know which way you would prefer to go.

ra1fh:
> From your PR I gather the linux scope doesn't include openbsd.

That's correct.

> If so we need to include openbsd in GPSBabel.pro (as you have done) and gui/app.pro. 

The official OpenBSD port does use the unmodified gui/app.pro to build the GUI. From that I'd derive that no modifications are needed to build the GUI at least. Btw, they use configure for the non-UI parts.

> CMakelists.txt probably needs a similar change.

I will give it a try.

> Perhaps instead of openbsd we should use the more generic bsd scope? We probably wouldn't want that to trigger on any darwin/macos/macosx systems. It isn't clear to me what the tree of Qt scopes looks like.

A brief look at mkspecs suggests darwin and macos are not included in the bsd scope.
Anyway I would rely on the porters of the other BSDs to submit patches. Chances are they are not using qmake.

> We could rip out HAVE_LINUX_HID out in a subsequent PR, or you can add that to this PR. Let me know which way you would prefer to go.

I will take a look at CMakeLists.txt and add the changes to this PR if necessary. I'd prefer to keep the LINUX_HID cleanup separate.

robertlipe:
Ralf, do you know of an easy way for us to run OpenBSD with CI? We use
Appveyor for MacOS and Linux and it's worked well for us. Is there a
similar resource you can help us set up that we should be watching?

If you have any pull on using configure, please vote for NOT using
configure. It's easy to imagine a world for us where configure falls into
ill repair or removed completely. If I watch it check for <stdio.h> one
more time or have to walk into that 7800 line shell script again, it could
all go up in a burst of flames. I think that qmake is about the right
balance for us.

tsteven4:
A docker image, with all the prerequites we need, that we could run on travis would be great.  I build a ubuntu image, see tools/make_docker_image and tools/Dockerfile, tools/Dockerfile_qtio.  Travis runs ubuntu anyway but we started down this path before they had bionic.  It also allows us to test with newer versions of Qt.

ra1fh:
> Ralf, do you know of an easy way for us to run OpenBSD with CI?

I'm not aware of an easy way to run OpenBSD (or any of the BSD) with CI. I only found this one here:

https://erouault.blogspot.com/2016/09/running-freebsd-in-travis-ci.html

This  uses qemu to run FreeBSD on Travis-CI. This is not really easy and low maintenance, as the qemu image has to be created and maintained. Travis has the ability to run full virtualization, but according to their ticket system they reserve that for MacOS.

> If you have any pull on using configure, please vote for NOT using
configure

I can certainly submit a patch to their mailing list once the next gpsbabel release is out to switch to qmake. If you want to be sure that configure is not used anymore, you'll have to remove it ;-)

I was wondering already why you care to maintain three build system. Since gpsbabel wouldn't build without Qt, the target platforms should always have qmake, no?

> A docker image, with all the prerequites we need, that we could run on travis

As the standard docker runtime runs on the host kernel using Linux namespaces and cgroups, a simple docker container is not possible for any of the BSDs due to different system call ABI. The qemu in docker approach could maybe work to package the approach described in the above blog post. But it's another layer of complexity with little benefit.

PR #469, opened 2020-01-21
Delete unused cet bits.

tsteven4:
Try to consistently use Latin1 for garmin format.

At one time the garmin format tried to figure out the character set based on the model.  Subsequently there was an attempt to let the user force the character set.  At this point the character set is hardcoded, hopefully it consistently uses Latin1 now.  We could add support for the other historical character set, windows-1252, but it's not clear if there is any interest.


tsteven4:


```diff

Clones removed
==============
+ cet.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gpsbabel/gpsbabel/pullRequest?prid=4862174&bid=16127670)

tsteven4:


```diff

Clones removed
==============
+ cet.cc  -2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gpsbabel/gpsbabel/pullRequest?prid=4862174&bid=16127670)

GPSBabelDeveloper (approved):
This makes me happy. I know how hard it is to delete code with this amout of tentacles.

It's worth mentioning that we're about to unmask a possible set of bugs.

In the beginning, the format writers were passed a const waypoint. There were some formats (not sure if any remain) that cast away the const-ness and modified the waypoint when writing them. Olaf laughed this way as OK because we were making copies of everything before passing them to the writer.

There is a possibility that by nuking CET, we'll open ourselves to pipeline conditions where
gpsbabel ... -o naughtyformat -F file1 -o gpx -F file2 will have different contents in 'file2' depending on what damage was done by naughtyformat.

I'm unsympathetic to that being a reason to keep this. We should remain vigilant on writers not doing dumb things and simply fixing cases like this if we see them.

The turndown of CET (patches before this one) is a happy day for memory use: we'll make one fewer copy of MOST strings while writing. Hooray!

tsteven4:
Your correct about introducing a vulnerability to the abuse of casting away const, at least in the case where the encoding isn't utf8 (we didn't do the backup/restore then).  gdb looks like it is up to no good.  I think I changed all the casting away of const to const_casts, so they are easy to find.  I agree, this isn't a good reason to delete the backup/restore, but we should look at the const_casts in the formats.

tsteven4:
Another thing to watch for is formats that access the global_waypoint_list.  netstumbler is one, and it seems to modify waypoints it finds via the list.  However, it appears to do this on read, i.e. to waypoints it just created.

PR #475, opened 2020-01-25
gpsbabel::textstream now accepts open mode flags

tsteven4:
Previously only an open mode flag was accepted, now a combination
of flags will be accepted.

### DIFF
--- a/src/core/textstream.h
+++ b/src/core/textstream.h
@@ -16,9 +16,10 @@
     Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
  */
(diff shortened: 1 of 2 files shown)


src/core/textstream.h
```diff
@@ -32,7 +33,7 @@ namespace gpsbabel
 class TextStream : public QTextStream
 {
 public:
-  void open(const QString& fname, QIODevice::OpenModeFlag mode, const char* module, const char* codec = "UTF-8");
+  void open(const QString& fname, QIODevice::OpenMode mode, const char* module, const char* codec = "UTF-8");
```
GPSBabelDeveloper, line 36: +1 Good idea. That is kind of a wart in the Qt API.

PR #486, opened 2020-01-30
relocate no output format display routine.

tsteven4:
waypt_disp() is used by main when no output format is specified.
This moves that routine from waypt.cc to main.cc.


tsteven4:
- waypt_display_all() often requires pollution of the global namespace to share information with the callback routine.  In this case the information to be shared is captured by reference in a lambda expression instead.  It is important to avoid creating dangling references when doing so.
- The mkshort handle is created and destroyed by the constructor and destructor of FallbackOutput.

tsteven4:
This PR also cleans up a memory leak.  Previously the global_waypt_list was empty at exit, but it can still consume memory.

PR #487, opened 2020-01-31
Fix memory leak in lowranceusr version 4 writer.

tsteven4:
memory was still reachable, due to global pointer it wasn't lost.

### DIFF
--- a/lowranceusr.cc
+++ b/lowranceusr.cc
@@ -85,28 +85,32 @@
 
 */
 
-#include <cmath>                 // for M_PI, atan, exp, log, tan
-#include <cstdio>                // for printf, snprintf, sprintf, SEEK_CUR
-#include <cstdlib>               // for atoi, abs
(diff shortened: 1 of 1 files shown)


lowranceusr.cc
```diff
@@ -337,9 +341,7 @@ static char*          opt_content_descr;
 static char*          opt_serialnum;
 static int            opt_serialnum_i;
 
-static const Waypoint**     waypt_table;
-static int            waypt_table_sz;
-static int            waypt_table_ct;
+static QList<const Waypoint*>* waypt_table{nullptr};
```
GPSBabelDeveloper, line 345: If this was a std::shared_ptr that was either allocated in the ctor or reset() when needed, just nulling it or letting it fall out of scope (when formats.are.lazy initted) would let it be freed.

That's also a good project for future days.. eliminations Manual allocation will be a good starter project for someone. Either way is fine here, but any time we type delete, we should think about RAII.

PR #488, opened 2020-01-31
re-implement gpx version handling.

tsteven4:
Clean up data members related to version information.
Avoid casting string literals to (non-const) char*s.

### DIFF
--- a/gpx.cc
+++ b/gpx.cc
@@ -35,6 +35,7 @@
 #include <QtCore/QStringRef>                       // for QStringRef
 #include <QtCore/QTime>                            // for QTime
 #include <QtCore/QVector>                          // for QVector
+#include <QtCore/QVersionNumber>                   // for QVersionNumber
 #include <QtCore/QXmlStreamAttribute>              // for QXmlStreamAttribute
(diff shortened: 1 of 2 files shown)


GPSBabelDeveloper (approved):
Much easier on the eyes. Thanks!
I know that crossing input and output traits is funky, but should we warn if we read gpx 1.0 and write >=1.1? We can fix link but third.party schemas will become dog meat.

tsteven4:
This shouldn't change any behavior.  The ways to read 1.0 and write 1.1 are

1. -i gpx -f ver1.0.gpx ... -o gpx,gpxver=1.1
2. -i gpx -f ver1.0.gpx ... -o gpx,humminbirdextensions
3. -i gpx -f ver1.0.gpx ... -o gpx,garminextensions
4. -i gpx -f ver1.0.gpx -f ver1.1.gpx -o gpx ...

which is the same as it's always been.   In case 1 the user explicitly requested 1.1, In cases 2, 3 they requested an extension that requires 1.1, and in case 4 we had mixed input.  I don't see the need for a warning, especially as this doesn't seem to have been an issue and it has worked this way for many years.

I could see the default being 1.1 instead of 1.0, 

> The GPX 1.1 schema was released on August 9, 2004.

PR #492, opened 2020-02-04
fix removal of tomtom test temporary ov2 file

cjmayo:
Typo TMDIR.


tsteven4:
Nice find @cjmayo. thanks.

PR #491, opened 2020-02-04
Destructor of class Waypoint should be virtual

ra1fh:
In nmea.cc, waypoints are allocated as NmeaWaypoint, which
is a subclass of Waypoint. Later on those waypoints are
deleted through a pointer to the base class in WaypointList::flush().

Found by gcc 9.2.1 with address sanitizer on Fedora 31.

(./gpsbabel -i nmea -f ./reference/track/nmea -o gpx -F /tmp/gpsbabel.69273/nmea.gpx)
=================================================================
==70649==ERROR: AddressSanitizer: new-delete-type-mismatch on 0x610000000340 in thread T0:
  object passed to delete has wrong type:
  size of the allocated type:   192 bytes;
  size of the deallocated type: 184 bytes.
    #0 0x7f210e8a70b5 in operator delete(void*, unsigned long) (/lib64/libasan.so.5+0x1110b5)
    #1 0x6edd68 in WaypointList::flush() /home/gpsbabel/gpsbabel.git/waypt.cc:742
    #2 0x6e4ea3 in route_head::~route_head() /home/gpsbabel/gpsbabel.git/route.cc:391
    #3 0x6e5929 in RouteList::flush() /home/gpsbabel/gpsbabel.git/route.cc:473
    #4 0x6e29e8 in route_flush_all_tracks() /home/gpsbabel/gpsbabel.git/route.cc:180
    #5 0x6e29fe in route_deinit() /home/gpsbabel/gpsbabel.git/route.cc:187
    #6 0x74648f in main /home/gpsbabel/gpsbabel.git/main.cc:669
    #7 0x7f210dd3c1a2 in __libc_start_main (/lib64/libc.so.6+0x271a2)
    #8 0x4104bd in _start (/home/gpsbabel/gpsbabel.git/gpsbabel+0x4104bd)

0x610000000340 is located 0 bytes inside of 192-byte region [0x610000000340,0x610000000400)
allocated by thread T0 here:
    #0 0x7f210e8a59d7 in operator new(unsigned long) (/lib64/libasan.so.5+0x10f9d7)
    #1 0x484492 in gpgga_parse /home/gpsbabel/gpsbabel.git/nmea.cc:514
    #2 0x489418 in nmea_parse_one_line /home/gpsbabel/gpsbabel.git/nmea.cc:996
    #3 0x489bd3 in nmea_read /home/gpsbabel/gpsbabel.git/nmea.cc:1092
[...]

PR #497, opened 2020-02-06
add address sanitizer, undefined behavior sanitizer, clazy to CI

tsteven4:
All these run on a Docker image of Ubuntu Eoan Ermin, 19.10.  We will need to migrate them to Focal Fossa 20.04 LTS when it is available (scheduled release April 2020).  These images are close to the bleeding edge.

The sanitizers are run via gcc, currently on Eoan this is version 9.2.1.  This version did expose recent issues, #491, #494, commit 5366625ad308898dc620d37ce2eb7cd9bfa4bf14.

clazy on Eoan is version 1.5.  It often catches errors like
c++11 range-loop might detach Qt container (QByteArray) [-Wclazy-range-loop]
which can be fixed by making the container const or using qAsConst() in the for clause.

### DIFF
--- a/mmo.cc
+++ b/mmo.cc
@@ -706,7 +706,7 @@ mmo_read_CObjTrack(mmo_data_t* data)
       // XXX ARB was u8 = gbfgetc(fin); but actually a string
       // Don't construct a QString we aren't going to use.
       // avoid clazy-unused-non-trivial-variable
-#if 1
+#ifdef MMO_DBG
       QString text =
 #else
       (void)
--- a/mmo.cc
+++ b/mmo.cc
@@ -704,7 +704,14 @@ mmo_read_CObjTrack(mmo_data_t* data)
 
     if (mmo_version >= 0x16) {
       // XXX ARB was u8 = gbfgetc(fin); but actually a string
-      QString text = mmo_readstr();
+      // Don't construct a QString we aren't going to use.
+      // avoid clazy-unused-non-trivial-variable
+#if 1
+      QString text =
+#else
+      (void)
+#endif
+        mmo_readstr();
       DBG((sobj, "text = \"%s\"\n", qPrintable(text)));
       uint16_t u16 = gbfgetuint16(fin);
       DBG((sobj, "unknown value = 0x%04X (since 0x16)\n", u16));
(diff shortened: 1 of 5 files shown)


mmo.cc
```diff
@@ -704,7 +704,14 @@ mmo_read_CObjTrack(mmo_data_t* data)
 
     if (mmo_version >= 0x16) {
       // XXX ARB was u8 = gbfgetc(fin); but actually a string
-      QString text = mmo_readstr();
+      // Don't construct a QString we aren't going to use.
+      // avoid clazy-unused-non-trivial-variable
+#ifdef MMO_DBG
+      QString text =
```
GPSBabelDeveloper, line 710: Can these be
Qstring junk = ...
Q_UNUSED(foo)
?

tools/build_extra_tests
```diff
@@ -0,0 +1,47 @@
...
+# echo some system info to log
+uname -a
+if [ -e /etc/system-release ]; then
+	cat /etc/system-release
+fi
+if [ -e /etc/os-release ]; then
+	cat /etc/os-release
+fi
+git --no-pager log -n 1
+# the timestamps from a svn co are unpredicatble.
```
GPSBabelDeveloper, line 16: Svn to git (or more abstract)

GPSBabelDeveloper (approved):
Awesome. The state of tooling these days is great for us. Thanks!

tsteven4:


```diff

Issues
======
+ Solved 2
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gpsbabel/gpsbabel/pullRequest?prid=4953929&bid=16372394)

mmo.cc
tsteven4: This is a common pattern where the only usage is in a debug statement which imbeds another conditional test.    I don't particularly like cutting up the code with the preprocessor directives but I don't see a better alternative.  I think somewhere else we use preprocessor directives to split the non-debug and debug cases where the call to the read function appears in both, i.e.
```
#ifdef DEBUG
  QString junk = read()
  std::cout << junk << std:endl;
#else 
  (void) read();
#endif 
```

1. QString junk =, where junk is unused, will be flagged by clazy as [unused-non-trivial-variable](https://github.com/KDE/clazy/blob/master/docs/checks/README-unused-non-trivial-variable.md).
2. Q_UNUSED(foo) is an alternative to (void) foo, but only for the non-debug case.  Casting to void is a common solution independent of Qt.  Have a glance at [Q_UNUSED](https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/global/qglobal.h?h=5.12.7#n117)

GPSBabelDeveloper:
I think the next time I'm bored, I'll make MMO_DBG go away and make all
that into a runtime test.(I'm not a fan inside gdb or gpi, either. All have
the same original author...)  Having multiple compilation paths is just
hard to maintain.  Code like mmo_get_object rewinding the file offset and
then -rereading two bytes is just weird.

I knew that Q_UNUSED was just eye candy for a void cast. The war between
optimizers finding dead loads and stores and programmers leaving
conditional code in the build rages on.


We have more interesting fish to fry for now.

That sounds pretty yummy, actually...


ASAN (and its siblings) found lots of good things on our code at a previous
employer, so I'm looking forward to it helping catch weird stuff in code so
we humans don't have to. Thanx for setting this up.

PR #499, opened 2020-02-06
Some fixes for issues found with afl and address sanitizer

ra1fh:


GPSBabelDeveloper (approved):
Thanx!

ra1fh:
Btw do you have any insight in which formats are used most (besides the obvious ones like gpx and kml)? Would make most sense to let afl run on those. The qmake project distinguishes between MINIMAL_FMTS, ALL_FMTS and DEPRECATED_FMTS. For now I have primarily looked at the ones from MINIMAL_FMTS (plus igc which had another address sanitizer finding recently and has fairly suspicious code).

tsteven4:
LGTM as well. Thanks.

PR #501, opened 2020-02-09
Implement GCID parsing in unicsv and extend ID to 64bit

ra1fh:
This reimplements the parser and adds some test cases. The test
cases have been validated with several online converters, so this
hopefully works better than the previous implementation.

This also fixes an endless loop that afl managed to trigger.


unicsv.cc
```diff
@@ -366,7 +366,7 @@ unicsv_parse_gc_id(const QString& str)
   }
 
   long long res = 0;
-  for (auto c : s) {
+  for (auto c : qAsConst(s)) {
```
tsteven4, line 369: CI rocks!

robertlipe:
That's some of the oldest code in the tree, originally from the 3 digit
GCID days. It's been dragged into 4 digit hex and later "extended 4" and
more and later into Qt, so it's had a tough life.

Thanx for the update. LGTM!

unicsv.cc
```diff
@@ -1010,7 +1048,7 @@ unicsv_parse_one_line(const QString& ibuf)
       switch (unicsv_fields_tab[column]) {
 
       case fld_gc_id:
-        gc_data->id = value.toInt();
+        gc_data->id = value.toLongLong();
```
tsteven4: Would it be more clear if  we used the ok parameter of toLongLong instead of relying on the result being zero when conversion fails (which is guaranteed by toLongLong).
Also a comment here saying if conversion succeeded then we assume the input is a geocache ID, and otherwise we assume it is a geocache reference code.

unicsv.cc
```diff
@@ -323,23 +323,61 @@ static QVector<arglist_t> unicsv_args = {
 
 /* helpers */
 
-// There is no test coverage of this and it's been wrong for years and
-// nobody has noticed...
-static int
+// Convert GCID / geo cache reference code into int64
+// (see also https://api.groundspeak.com/documentation#referencecodes)
```
tsteven4, line 327: Thanks for the citation.

tsteven4 (commented):
It disturbs me when we do an echo test that doesn't echo what we read, i.e.
"gpsbabel -i format -f input -o format -F output" results in input != output.
Sometimes there are explanations, e.g. ignored input fields in xcsv files.  The document defines

> gcid =     Geocache cache id

In your new test we are reading reference codes, not geocache ids.  On output we are writing geocache ids.  Are we smart enough on input that we can tell the difference and we do the correct thing in either case?  It seems we apparently are.  Should we have a different keyword for geocache reference codes?  Should we output both on write?  Should we clarify the documentation?

unicsv.cc
ra1fh: Yes, that's nice indeed. Especially since I do not have the latest version of all the checker tools everywhere.

unicsv.cc
ra1fh: @tsteven4  Might also make sense to explicitly add a base, otherwise this would accept hex and octals with 0x and 0 prefix.

> It disturbs me when we do an echo test that doesn't echo what we read

Yes indeed, but there doesn't seem to be the corresponding output code (besides a partial implementation in deprecated/geoniche.cc).

> In your new test we are reading reference codes, not geocache ids

The terminology was not entirely clear to me while implementing this, that might show in some places. An unrepresentative search engine query suggest that GC-Code and GC-ID are mostly used. The term "reference code" may be a generalization that also includes trackable codes starting with "TB". But I'm not sure.

tsteven4:
thanks again.

PR #503, opened 2020-02-09
A few build changes for Open/FreeBSD

ra1fh:
This was tested on Linux, OpenBSD and  FreeBSD 12.1. See individual commits for more details.

NetBSD is kind of special, they doesn't seem to define QMAKE_PLATFORM in their Qt5 package (although the mkspecs for NetBSD on other platforms do so). So it's somewhat hard to make that work with qmake. I did try at least.


ra1fh:
Maybe this was too much for a single PR, will resubmit individually.

PR #507, opened 2020-02-15
retire route_head_alloc().

tsteven4:
in favor of "new route_head", which is all route_head_alloc did.


GPSBabelDeveloper (commented):
So much hard work went into making these look so obvious. 😀 Thanks

PR #513, opened 2020-02-25
Change qmake target to lowercase gpsbabel on BSD

ra1fh:
The BSDs all package lowercase "gpsbabel", so switch the qmake
TARGET to lowercase as well. See also:

https://github.com/freebsd/freebsd-ports/blob/master/astro/gpsbabel/pkg-plist
https://github.com/openbsd/ports/blob/master/geo/gpsbabel/pkg/PLIST-main
http://cvsweb.netbsd.org/bsdweb.cgi/~checkout~/pkgsrc/geography/gpsbabel/PLIST

### DIFF
--- a/GPSBabel.pro
+++ b/GPSBabel.pro
@@ -15,7 +15,7 @@ if(equals(QT_MAJOR_VERSION, $$MIN_QT_VERSION_MAJOR):equals(QT_MINOR_VERSION, $$M
 
 QT -= gui
 
-linux: {
+linux|bsd: {
   TARGET = gpsbabel
 } else {
   TARGET = GPSBabel


tsteven4:
The various capitalizations of the command line executable is a PITA.  It dribbles into test, coverage (which makes a symbolic link in the pro file), etc..  It is in various places in the gui code.  The case sensitivity of the file system can matter.  I suspect the current GUI doesn't work on a case sensitive file system if the command line executable isn't lower case.

I think your change is a step in the correct direction.  An alternative might be

> unix:!mac {
>   TARGET = gpsbabel
> } else {
>   TARGET = GPSBabel
> }

or

> win32|mac {
>   TARGET = GPSBabel
> } else {
>   TARGET = gpsbabel
> }


You can see in GPSBabel.pro, gui/app.pro that we are not very consistent in our use of scopes.  I think this is the result of an accident as opposed to a plan.

I'd go with what you have, or one of the above alternatives.  @robertlipe ?

GPSBabelDeveloper:
I'll confirm that most of this is an accident.

We have two different things that we like to call "gpsbabel" (in some
casing...) - the GUI and the CLI. On some combination of OSes and history,
life was very bad with two of the same name, so we had the "GPSBabelFE"
nonsense but that was never very awesome.

We're probably never installed into an Applications on Mac that's
case-sensitive because half the apps break. People needed to develop in
worlds where case matters on Macs do it on an external drive or a DMG
that's created with a case-sensitive filesystem.

Of the three options, I think I like the last the best. Windows and Mac are
the weird ones - let's let them be the special case. Let's try to never
make *bsd the special case just because they're in a minority of test suite
and mind share (sorry, Ralf) for us.

HOWEVER, can we rule this one with an iron fist and just say the command
line version is lower case, period? I'm preparing a pull request right now
to see if that passes integration.

P.S. Sorry for the delays. I haven't been well.

RJL

tsteven4:
Between 23379e7 and #520 the command line executable is now always lower case gpsbabel.  @ra1fh thanks for your work.  Motivated by your efforts we have solved this particular problem in another way.  Please continue to let us know about other issues you encounter.

PR #518, opened 2020-03-13
Let KML <TimeSpan> interpolation not count the ending fencepost

GPSBabelDeveloper:
Formerly, four points would be counted as four points for interpolation, but it's actually three.


Stadly:
When can I expect this to become part of a release?

Issue #517, opened 2020-03-12
dg-100 regression: download fails with checksum mismatch


akhuettel:
I recently re-activated my old and trusty DG-100 logger, but now I can't download recorded data anymore. My issue (on Gentoo linux) seems to be identical with Debian bug 804476, 
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=804476

If that is the case, then the bug was introduced somewhere between 1.5.0 and 1.5.2.

huettel@noumea ~ $ gpsbabel -t -w -D 1 -i dg-100 -f /dev/ttyUSB0 -o gpx -F /home/huettel/ki-test.gpx
GPSBabel Version: 1.5.4 
options: module/option=value: dg-100/erase="0" (=default)
options: module/option=value: dg-100/erase_only="0" (=default)
TX: Frame Start a0 a2 Payload_Len 0003 Cmd: getfileheader
RX: Start a0a2 Len 0016 Cmd: getfileheader
found 1 headers, nextheader=1
Header #0: Seq: 0 Time: Thu Mar 12 14:47:30 2020
TX: Frame Start a0 a2 Payload_Len 0003 Cmd: getfileheader
RX: Start a0a2 Len 000a Cmd: getfileheader
found 0 headers, nextheader=0
no further headers, aborting the loop
TX: Frame Start a0 a2 Payload_Len 0003 Cmd: getfile
RX: Start a0a2 Len 0401 Cmd: getfile
RX: Start a0a2 Len 0401 Cmd: getfile
TX: Frame Start a0 a2 Payload_Len 0001 Cmd: getconfig
RX: Start a0a2 Len 0039 Cmd: getconfig
checksum mismatch: data sum is 0x043c, checksum received is 0x0004
huettel@noumea ~ $

akhuettel:
Reverting dg-100.cc to the 1.5.0 state (and re-applying some type fixes so things build again) makes the command work:

huettel@noumea ~ $ gpsbabel -t -w -D 1 -i dg-100 -f /dev/ttyUSB0 -o gpx -F /home/huettel/ki-test.gpx
GPSBabel Version: 1.5.4 
options: module/option=value: dg-100/erase="0" (=default)
options: module/option=value: dg-100/erase_only="0" (=default)
TX: Frame Start a0 a2 Payload_Len 0003 Cmd: getfileheader
RX: Start a0a2 Len 0016 Cmd: getfileheader
found 1 headers, nextheader=1
[...]

PR #519, opened 2020-03-14
work around DG-100 getconfig issues

tsteven4:
Don't issue the getconfig command on the DG-100.  Different DG-100
devices seem to return different numbers of bytes, which leads
to checksum errors, which we treat as fatal.  Before release
1.5.1 we didn't issue this command.  It was apparently added because
it changed the status light to green on the DG-200.

### DIFF
--- a/dg-100.cc
+++ b/dg-100.cc
@@ -1,4 +1,4 @@
-/*
+  /*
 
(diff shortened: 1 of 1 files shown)


tsteven4:
This fixes #517

PR #525, opened 2020-03-27
update Dockerfiles to use webengine instead of webkit.

tsteven4:


GPSBabelDeveloper:
Cool. There's probably some #ifdef baggage that can go away now if Linux is
on a supportable web interface.

tsteven4:
support for qtwebengine5 started on ubuntu with bionic.  not sure about other distros.  I don't see a need to rip out webkit support at this time.

PR #530, opened 2020-04-03
Optimize QString usage in gpx format. 

tsteven4:
Don't convert QStringRef to QString unnecessarily.
Use QLatin1String, QStringLiteral.

An 18% performance improvement has been measured with
gpxbabel -f lowrance-v4.gpx -o lowranceusr,wversion=4 -F lowrance-v4.usr

valgrind --tool=callgrind was useful.  Further improvements in the
reader are likely to be found in xml_parse_time.


GPSBabelDeveloper:
Wow, 18% boost is nice! 
Someday, all those inline QString foos may need to get moved out into class-scoped members, but determining their life cycle and interactions between read and write could get ugly.
The current_tag was a cute trick in C, but we're "just" building something that (probably) has a good hash value. These files have a efw stretch marks on them, being some of our oldest pillars.

Thanx for keeping them spiffed!

Issue #534, opened 2020-04-10
subRip subtitle datetime options aren't optional


Matthias84:
Hi, I encounter a lot of troubles, while getting a working setup to georeference a video sequence with a GPS track.
Long story short: The gps_date setting isn't optional, [like the docs say](https://www.gpsbabel.org/htmldoc-development/fmt_subrip.html).
Currently if you run it only with gps_time (because you expect it be optional, as it is just one day), it will fall back to start always at the very first GPS node, ignoring your gps_time
`gpsbabel -i gpx -f tracklog.gpx -o subrip,video_time=000142,gps_time=192521 -F subtitles.srt`

IMHO this is unexpected and pretty confusing. I recommend to print at least a warning, that we fallback to this default behaviour. Otherwise indicating both parameters as non optional, will avoid this misunderstanding?
Another idea is to use just the very first day of a GPS track as default, so in most cases it doesn't need the extra gps_date setting?

P.S: At the GUI (subrip options dialog) the gps_date label is wrong. The format string is 'yyyymmdd'

tsteven4:
Thanks for your report, also the postscript.  I am sorry for your difficulties.

I note there is a comment in the code https://github.com/gpsbabel/gpsbabel/blob/a66e6468bd7f04eb4c3984ce4b404259d5329d9b/subrip.cc#L264

It seems this was done:
In the document https://www.gpsbabel.org/htmldoc-development/fmt_subrip.html under gps_time is states 

> This option must be used together with gps_date; if one or both are missing, the timestamp of the first GPS trackpoint is used. 

Under gps_date it states 

> This option must be used together with gps_time; if one or both are missing, the timestamp of the first GPS trackpoint is used.

What part of the document led you to believe you could use one without the other?

Printing a warning is reasonable.  I don't understand what you mean by "Otherwise indicating both parameters as non optional".   Using the first day of the track as a default could be confusing as the date could easily roll over, especially since the times are in UTC.

Matthias84:
Hi, thanks for the quick response :+1: 

if you just fire up the GUI, you are allowed to use one field or the other :wink: So I just expected, that you get an appropriate date otherwise. As the processing finishes without troubles (and in some constellations with a good result), it was hard to see what is the problem.

A warning could be a hint, just mentioning the default fallback, or a error message, showing that you will need both params to get a useful. Last one was the "inidicating non optional parameters".

[...]

PR #536, opened 2020-04-13
rework subrip format synchronization.

tsteven4:
support subsecond synchronization.
validate option input.
add debug messages for synchronization.
correct option text for GUI.


tsteven4:
This fixes #534

PR #537, opened 2020-04-14
Fix access to QByteArray/QString outside the valid range.

tsteven4:
With Qt 5.15, and likely 5.14, the following warnings were generated during
testo:
Using QByteRef with an index pointing outside the valid range of a QByteArray. The corresponding behavior is deprecated, and will be changed in a future version of Qt.
Using QCharRef with an index pointing outside the valid range of a QString. The corresponding behavior is deprecated, and will be changed in a future version of Qt.
Note this requires Qt to be compiled for debug.

These warnings can be debugged by running testo with
"export QT_FATAL_WARNINGS=1" to generate core dumps.

I suspect the appending of null terminators in ggv_bin is unnecessary,
i.e. I believe QByteArray::resize() in ggv_bin_read_bytes will take
care of this.  Never the less I slavishly kept adding them just to make
certain.

### DIFF
--- a/ggv_bin.cc
+++ b/ggv_bin.cc
@@ -84,7 +84,7 @@ GgvBinFormat::ggv_bin_read_text16(QDataStream& stream, QByteArray& buf, const ch
 {
   quint16 len = ggv_bin_read16(stream, descr);
   ggv_bin_read_bytes(stream, buf, len, descr);
-  buf[len] = 0;
+  buf.append('\0');
   if (global_opts.debug_level > 1) {
     qDebug() << "ovl: text =" << QString::fromLatin1(buf.constData()).simplified();
   }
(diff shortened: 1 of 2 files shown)


robertlipe (approved):

ra1fh:
@tsteven4 As long as the QByteArrays are used via constData(), the null-termination is required. There is no guarantee that the data fields in the file are null-terminated, although they typically are.

The alternative of constructing QString from QByteArray directly instead from constData() is not ideal either, as this may include all the binary data in output and track name and not stop at the first null terminator.

So your change looks good :-)

Thanks,
Ralf

tsteven4:
@ra1fh thanks for reviewing the change.

The reason I thought appending the terminator may not be required is that QByteArray makes a promise "Behind the scenes, it always ensures that the data is followed by a '\0' terminator".  It appears QByteArray::resize() keeps this promise: https://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/text/qbytearray.cpp#n1743.  But, the documented promise seems a bit vague, and the Qt source code isn't a promise of future operation.  So the safest thing to do is explicitly add the terminator as was done before and after this PR.

ra1fh:
@tsteven4 Ah I see. I missed that part in the docs. So it would be safe to omit the additional null-terminator.

PR #538, opened 2020-04-14
Fallthrough corrections

tsteven4:
Fix an actual bug that resulted in a fallthrough warning!

Silence other fallthrough warnings with comments recognized by gcc.

The real fix is the to add the attribute [[fallthrough]];
which was added in c++17.
gcc accepts this, but until c++17 there wasn't
a requirement to ignore unrecognized attributes, so we can
imagine a pre c++17 compiler might choke on it.

### DIFF
--- a/dmtlog.cc
+++ b/dmtlog.cc
@@ -586,7 +586,8 @@ inflate_buff(const char* buff, const size_t size, char** out_buff)
 
     switch (res) {
     case Z_NEED_DICT:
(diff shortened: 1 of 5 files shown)


robertlipe:
I wondered about using both for a while and then deleting the commented
version later but the GCC on MacOS, which is actually a front end to an
old-ish version of clang (11) actually doesn't accept that even in
-xclang=17 or 20 modes.

Carry on.

RJL

PR #540, opened 2020-04-15
A variety of loose ends

tsteven4:
fix parameter passed by value that can be passed by const reference (clang-tidy).

Add widgets to QT variable in gui/app.pro.  Somehow this was required with Qt 5.15 in a hacked version that excluded QtWebEngine.  The document clearly states it should be added when one used QDialog.

Add qmake target to generate gpsbabel.pdf.

minimize low value output when generated documents:
- drop --load-trace xsltproc options when building document.
- set logging level for fop to WARN.
This shows numerous instances where the pdf document has text off the right side of the page.

Set doc version to development in configure.ac.  It was set in Makefile.in, but that misses tools/mkcapabilities which caused extraneous output when the document was generated.


robertlipe (commented):
Nice cleanups. Thanx.


The case of $ac_file$ac_mode in configure is weird, but it's generated so I won't puzzle over it much.

tsteven4:
> The case of $ac_file$ac_mode in configure is weird, but it's generated so I won't puzzle over it much.

Straight out of the second example in https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Configuration-Files.html

PR #543, opened 2020-04-19
improve MSVC 2015 workaround.

tsteven4:
use improved workaround for MSVC 2015.  Supplying a parametrized constructor and the default constructor allows the intitialization lists to work properly in MSVC2015.  This same workaround has been used in several other places.

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -972,19 +972,28 @@ void setshort_defname(short_handle, const char* s);
 #define ARG_NOMINMAX nullptr, nullptr
 
 struct arglist_t {
-  const char* argstring;
-  char** argval;
-  const char* helpstring;
-  const char* defaultvalue;
-#if !defined(_MSC_VER) || (_MSC_VER >= 1910) /* !MSVC or MSVC 2017 or newer */
-  const uint32_t argtype{ARGTYPE_UNKNOWN};
-#else
-  /* MSVC 2015 generates C2440, C2664 errors with the above. */
-  uint32_t argtype;
+  /* MSVC 2015 generates C2440, C2664 errors without some help. */
+#if defined(_MSC_VER) && (_MSC_VER < 1910) /* MSVC 2015 or earlier */
+  arglist_t() = default;
+  arglist_t(const char* astr, char** aval, const char* hstr, const char* dval,
+            const uint32_t atyp, const char* minv, const char* maxv, char* avp) :
+            argstring(astr),
+            argval(aval),
+            helpstring(hstr),
+            defaultvalue(dval),
+            argtype(atyp),
+            minvalue(minv),
+            maxvalue(maxv),
+            argvalptr(avp) {}
 #endif
-  const char* minvalue;		/* minimum value for numeric options */
-  const char* maxvalue;		/* maximum value for numeric options */
-  char* argvalptr;	/* !!! internal helper. Not used in definitions !!! */
+  const char* argstring{nullptr};
+  char** argval{nullptr};
+  const char* helpstring{nullptr};
+  const char* defaultvalue{nullptr};
+  const uint32_t argtype{ARGTYPE_UNKNOWN};
+  const char* minvalue{nullptr};    /* minimum value for numeric options */
+  const char* maxvalue{nullptr};    /* maximum value for numeric options */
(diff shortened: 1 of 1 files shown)


robertlipe:
Can we bump the floor to 2017 and avoid ifdefs?

Should we consider dropping it in favor of Visual Studio, which eliminates
price concerns and is therefore easier to update?

We don't have a lot of windows devs that are visible here.

tsteven4:
I have been considering adding MSVC 2019 and dropping MSVC 2015.  Appveyor seems to update their MSVC 2019 environment much more often that their MSVC 2017 or MSVC 2015 environment.

I have also been thinking about c++17.  I would like static constexpr containers for things like garmin_fit, but I don't think c++17 gets us those.  It would avoid more conditionals, see comment in our code "Until c++17 we have to define odr-used constexpr static data members at namespace scope."  It would also let us retire our version of optional.

The savings in lines of code is minor and not very important.  We could do better by deprecating one extinct dinosaur format.  More important to me is the time to work around MSVC 2015 issues.

The MSVC tools are used by Visual Studio.  There is no cost for Visual Studio Community edition or the MSVC tools.  Updating the MSVC tools used by Visual Studio happens when you update Visual Studio.  VS let's you know when an update is available.

I think our most important windows build environment by far is appveyor.

robertlipe:
On Mon, Apr 20, 2020 at 1:05 PM tsteven4 <notifications@github.com> wrote:

> I have been considering adding MSVC 2019 and dropping MSVC 2015.
>
Approved.

I have also been thinking about c++17.s
>
Approved.


> I would like static constexpr containers for things like garmin_fit, but I
> don't think c++17 gets us those.
>
I think it does. C++17, being a relatively small bump, seems quite common
so that's a reasonable floor, IMO. Class or even struct-local inline consts
are now a thing. One less reason for tables of #defines or global const
kFoo.

if (auto data = ReadValue; data != failure)
   use(data)
else
  log_failure(data)
data's dtor runs here as it goes out of scope.

No more
  auto data:
  // data has life span before the if/else.
We do this all over the place.

string_view was superhandy but QString takes some of its thunder and I
think it even inherited a non-owning string view.

Of recent interest, the attributes for [[fallthrough]] and
[[maybe_unuesed]] have (ugly) Qt equivalents, but they're speced now. No
more (void) foo and magic comments.

__has_include erases a bit of our preprocessor magic.

I'd replaced a raft of Qt filesystem with C++17 a while ago. I don't feel
that strongly about it here. Needing boost for filesystem because the
language doesn't have a way to walk a directory has been dumb for a long
time. (Yes, I know about POSIX...)

The other one that comes to mind as hot (though I've not yet actually used
9)  was automatic parallelization of container walking. The compiler was
free to automatically vectorize or create/destroy threads automatically.
Something like:

std::foreach(std::execution:par, std::begin(a), std::end(a), [&](int) {
  ++x;
}

might result in vector ops or threads. I'm sure I've forgotten the details,
but I've long wanted fuss-free threading in our filters. Parallelism in
general is now an official thing!


I was bummed that {{fmt}} didn't make it into C++17. That would make some
of our printf-hoops so much easier to jump through. (I've flirted with
bringing in that implementation...)

The savings in lines of code is minor and not very important. We could do
> better by deprecating one extinct dinosaur format. More important to me is
> the time to work around MSVC 2015 issues.
>
Absolutely. Time and mental bandwidth is our most precious resource to
manage. Every failed presumbit because of some weird rule that isn't
present on our respective systems is just senseless soul spending.



> The MSVC tools are used by Visual Studio.
>
Is that equally true for Visual Studio Code, the cross-platform IDE? Wait,
VSC is "just" an IDE wrapping GCC, Clang, or Microsoft's own pro language.

In any combination we can care about, the cost to not upgrade to a current
version doesn't matter.

> I think our most important windows build environment by far is appveyor.
>

Agreed. Our long cow-towing to a mythical MS developer that's invariably
trapped on an un-upgradeable environment is silly. We just don't see any
activity there.

I don't see the need to paint the code with C++17 immediately, but I'd
rather vaporize the MSVC15 that requires the #ifdefs that insprired this
discussion and move to a newer base, even if we don't turn on C++17 mode.

Objections?

PR #545, opened 2020-04-21
Use strncmp to compare array to string

jspricke:


robertlipe:
Can we instead just dump these into a Qstring where we can test == (or
.equals) for readability? Feel free to respell surrounding code the same
way.

Thanx!

jspricke:
we would need to copy the array into a Qstring, I guess. But I'm not really into QT so feel free to adopt if you have a better way.

tsteven4:
@jspricke can you answer a few questions about this PR?

1. Do you use this format?
2. What was your motivation for this PR?  For example, did some tool make a suggestion about these lines of code?  If so, please identify the tool and suggestion.  Did gpsbabel fail to read a v900 file correctly?  If so, can you submit a test case.  If you had other motivation please explain it.

Thanks

jspricke:
* tsteven4 <notifications@github.com> [2020-05-25 05:41]:
>1. Do you use this format?

No.

>2. What was your motivation for this PR?  For example, did some tool make a suggestion about these lines of code?  If so, please identify the tool and suggestion.  Did gpsbabel fail to read a v900 file correctly?  If so, can you submit a test case.  If you had other motivation please explain it.

It failed to compile with gcc-10:

https://bugs.debian.org/957304

tsteven4:
Thanks for the details.

This is a regression test failure.
It is reproducible on focal fossa with g++ 10.0.1 20200411 (experimental)

Something strange is going on, an alternative "fix" just adds a qDebug statement:
```# git diff v900.cc
diff --git a/v900.cc b/v900.cc
index 5766030c..7fad54fe 100644
--- a/v900.cc
+++ b/v900.cc
@@ -70,7 +70,7 @@ INDEX,TAG,DATE,TIME,LATITUDE N/S,LONGITUDE E/W,HEIGHT,SPEED,HEADING,FIX MODE,VAL
 for a little more info, see structures:
        one_line_advanced_mode, one_line_basic_mode, one_line_common_start.
 ******************************************************************************/
-
+#include <QtCore/QDebug>
 #include "defs.h"
 #include <cassert>
 #include <cstdio>
@@ -328,6 +328,7 @@ v900_read()
       wpt->pdop = atof(line.adv.pdop);
 
       /* handle fix mode (2d, 3d, etc.) */
+      qDebug() << line.adv.fixmode;
       if (!strcmp(line.adv.valid,"DGPS")) {
         wpt->fix = fix_dgps;
       } else if (!strcmp(line.adv.fixmode,"3D")) {
```

tsteven4:
It also fails with g++ 10.1.0 from
```https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/ppa
# apt show g++-10
Package: g++-10
Version: 10.1.0-2ubuntu1~20.04
Priority: optional
Section: devel
Source: gcc-10
Maintainer: Ubuntu Core developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: Debian GCC Maintainers <debian-gcc@lists.debian.org>
Installed-Size: 31.1 MB
Provides: c++-compiler, c++abi2-dev
Depends: gcc-10-base (= 10.1.0-2ubuntu1~20.04), gcc-10 (= 10.1.0-2ubuntu1~20.04), libstdc++-10-dev (= 10.1.0-2ubuntu1~20.04), libc6 (>= 2.14), libgmp10 (>= 2:5.0.1~), libisl22 (>= 0.15), libmpc3, libmpfr6 (>= 3.1.3), libzstd1 (>= 1.3.2), zlib1g (>= 1:1.1.4)
Suggests: g++-10-multilib, gcc-10-doc (>= 10)
Download-Size: 9045 kB
APT-Manual-Installed: yes
APT-Sources: http://ppa.launchpad.net/ubuntu-toolchain-r/ppa/ubuntu focal/main amd64 Packages
Description: GNU C++ compiler
 This is the GNU C++ compiler, a fairly portable optimizing compiler for C++.
```

tsteven4:
Failure is related to optimization.  -O1 works, -O2, which we use, fails.

jspricke:
Note that `strcmp` is the wrong API to use, you want to check the full string. The optimization just shows that gcc deduces that `strcmp` is useless and removes the call.

tsteven4:
If the compiler is smart enough to realize line.adv.fixmode and line.adv.valid aren't NULL terminated strings of the necessary length despite that they are each followed with a structure member that is a NULL char, then although your fix eliminates this concern for those two fields I remain concerned that the same "trick" which is used to separate every other field may not work as well.  Looks like a little array decay can be your friend sometimes.

tsteven4:
Your fix is clearly an improvement even if it doesn't make us squeaky clean on the other fields.  Accepted.

Thanks for opening the PR, and maintaining the gpsbabel package on debian.

PR #547, opened 2020-04-25
introduce xasprintf that accepts a QScopedPointer&

tsteven4:
and use it to fix gcc 9 warnings in magproto, as well as other
possible string truncations/overflows.

This is a minimal risk stop gap measure.  It requires much less effort than conversion of all the legacy character strings, and avoids the back and forth between QStrings and character strings that sometimes occurs.  The use of a smart pointer to wrap the dynamically allocated string from xasprintf minimizes the risk of memory leaks.

QScopedPointerPodDeleter uses free as opposed to our xfree (which uses free).  For a purer implementation we could write our own deleter or get rid of our own memory allocation/deallocation routines, but this slight mismatch seems acceptable to me.

This could all be implemented with std::unique_ptr instead of QScopedPointer, but we have already introduced QScopedPointer to our code base while std::unique_ptr has not been used yet.

The passing of a smart pointer by non const reference to a function that will reset it is suggested by
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r33-take-a-unique_ptrwidget-parameter-to-express-that-a-function-reseats-thewidget

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -19,38 +19,39 @@
 #ifndef DEFS_H_INCLUDED_
 #define DEFS_H_INCLUDED_
 
-#include <algorithm>            // for sort, stable_sort
-#include <cmath>                // for M_PI
-#include <cstdarg>              // for va_list
-#include <cstddef>              // for NULL, nullptr_t, size_t
-#include <cstdint>              // for int32_t, uint32_t
-#include <cstdio>               // for NULL, fprintf, FILE, stdout
-#include <ctime>                // for time_t
-#include <utility>              // for move
+#include <algorithm>              // for sort, stable_sort
+#include <cmath>                  // for M_PI
+#include <cstdarg>                // for va_list
+#include <cstddef>                // for NULL, nullptr_t, size_t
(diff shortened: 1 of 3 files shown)


GPSBabelDeveloper:
That's very clever. I like it.

I consider every use of xmalloc/xfree a bug at this point. We've been whittling it down for years, but mostly it's low-value busy work to make them go away. Any time I'm working on something that uses it, I'll replace it with "real" C++. (Likewise time_t)

Some of the places we're using snprintf we can easily make QStrings and some have just incompresnsible caller context with a mixture of Qt and C/C++ that would be nightmares to replace with ostringstream or even {{fmt}}. This is a nice, low-friction support for those really old modules like magproto and nmea.

I think we favored QScopedPtr in the era when Qt had C++ and whatever floor we used of C++ didn't.  unique_ptr is fair game now, IMO, now that auto_ptr has died. 

Part of the allocation issues can go away when we dynamically create and destroy modules. Certainly having them function-scope like this is very nice.

Nice solution!

PR #549, opened 2020-04-28
add recoverymode and CRC checking to FIT reader.

tsteven4:
If present, the header CRC is checked.
The file CRC and length is checked.

A recoverymode option is added.
In the default mode we will fatal with:
a bad CRC,
a bad endian field,
an attempt to read when the data section doesn't have sufficient data,
an unexepected EOF.
In recovery mode when we encounter one of these errors we will abort
read processing and continue.  This allows a more immediate cleaner
exit from the reader while still allowing any writer to use data that
was recovered previous to the read abort.

### DIFF
--- a/garmin_fit.cc
+++ b/garmin_fit.cc
@@ -22,22 +22,26 @@
 
  */
 
-#include <cstdint>            // for uint8_t, uint16_t, uint32_t, int32_t, int8_t, uint64_t
-#include <cstdio>             // for EOF, SEEK_SET, snprintf
-#include <deque>              // for deque, _Deque_iterator, operator!=
-#include <memory>             // for allocator_traits<>::value_type
-#include <utility>            // for pair
-#include <vector>             // for vector
+#include <cstdint>             // for uint8_t, uint16_t, uint32_t, int32_t, int8_t, uint64_t
+#include <cstdio>              // for EOF, SEEK_SET, snprintf
+#include <deque>               // for deque, _Deque_iterator, operator!=
+#include <memory>              // for allocator_traits<>::value_type
+#include <string>              // for operator+, to_string, char_traits
+#include <utility>             // for pair
+#include <vector>              // for vector
 
-#include <QtCore/QByteArray>  // for QByteArray
-#include <QtCore/QDateTime>   // for QDateTime
-#include <QtCore/QString>     // for QString
-#include <QtCore/Qt>          // for CaseInsensitive
+#include <QtCore/QByteArray>   // for QByteArray
+#include <QtCore/QDateTime>    // for QDateTime
+#include <QtCore/QFileInfo>    // for QFileInfo
+#include <QtCore/QString>      // for QString
+#include <QtCore/Qt>           // for CaseInsensitive
+#include <QtCore/QtGlobal>     // for qint64
 
 #include "defs.h"
 #include "garmin_fit.h"
-#include "gbfile.h"           // for gbfputc, gbfputuint16, gbfputuint32, gbfgetc, gbfread, gbfseek, gbfclose, gbfgetuint16, gbfopen_le, gbfputint32, gbfflush, gbfgetuint32, gbfputs, gbftell, gbfwrite, gbfile, gbsize_t
-#include "jeeps/gpsmath.h"    // for GPS_Math_Semi_To_Deg, GPS_Math_Gtime_To_Utime, GPS_Math_Deg_To_Semi, GPS_Math_Utime_To_Gtime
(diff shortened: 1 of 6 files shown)


tsteven4:
This is related to #401 @spacelama,  This works with your sample file, but you may want to build it and check other files you have.

This is also related to 78559f889ee7c1fa9ac48512f83155edf3f5474b, from May 2012.  I couldn't find a sample file from Nils, but from the comment he left I believe this fix would work for his deficient Garmin Edge 800 files.

robertlipe:
Re: ...9f8, the files from Nils were provided in the first post of "[Gpsbabel-misc] Garmin Edge 800 FIT format won't convert after firmware upgrade". Sorry I broke the audit chain.

$ shasum */*800*fit  ~/Downloads/2012-04-12-14-53-58.fit
6f8a484973ba382d936f7df6bb26940f1654587b  track/garmin-edge-800.fit
6f8a484973ba382d936f7df6bb26940f1654587b  ~/Downloads/2012-04-12-14-53-58.fit

tsteven4:
That file is in our reference files.  It doesn't seem to require the 
change Nils made, we don't detect anything odd about it.

GPSBabelDeveloper:
Overall, I really like this. It's a nice, disciplined use of throw/catch.

Is this missing the git add to reference/xmldoc that provides the hew help
prose?
Stylistically, why do you mix stringizing and .nospace in warnings?

Perhaps we've made other changes in robustness handling that have made that
change unnecessary. Our understanding of  the structure of .fit has
certainly evolved over time. For some time the "new edge 800 firmware
issue" was one of our leading support problems, so I don't think that Nil's
device was uniquely blessed.

Thanx!

tsteven4:
nospace was used to to avoid a space between the file name and the final period, and independent of the concatenation of adjacent string constants.

Maybe our length was off by one, but I can't see how that would be specific to garmin edge 800 files.  With this PR I would throw the exception if we tried to read beyond the data section length, and that isn't happening with garmin-edge-800.fit.

I expanded on the helpstring in the document with https://github.com/gpsbabel/gpsbabel/pull/549/commits/dd93feb0d95b6e30270ace92f75891276b60d1a2

robertlipe:
Awesome. Thanx. That explanation makes sense.

I don't remember the details - and you've already looked up the code - but
we had happy Edge 800 users that updated and then they turned into sad
users. Then we released an update and it's been pretty quiet since. I
dunno. Something was unique to that file, but it's probably not worth it to
time travel back 8 years to see.

tsteven4:
Thanks for chasing the edge 800 reference file.  That let me reconstruct the history.

I went back and rebuilt 78559f88.  I put the fatal error in fit_getuint8 back in.  This caused both the garmin-edge-200 and garmin_edge-800 test cases to fail.  The root cause of this failure was because the header length was increased from 12 to 14 bytes to allow for a header CRC.  We failed to locate to the beginning of the data section after fit_parse_header.  When we read those two bytes we interpreted them as data messages with 0 fields.  This caused us to incorrectly decrement fit_data.len by 2, which brought us up short in the end. 

So the comment
`    // fail gracefully for GARMIN Edge 800 with newest firmware, seems to write a wrong record length
    // for the last record.
`
was misleading.  The GARMIN wrote a valid fit file.  The header length stated it was 14 instead of 12.  But we didn't use the header length information we recovered to position to the start of the data section.

So, we don't need the workaround in fit_getuint8 from 78559f88  that I already took out, and those cases shouldn't and don't generate anything that requires recoverymode.

This means all the throws in fit_getuint* are probably unnecessary, they could be fatals in all cases we know about.

GPSBabelDeveloper:
What an analysis! Thanks.

Since that is/was a reverse engineered format, we just didn't know that at
the time.

tsteven4:
The analysis of the garmin 800 failure discussed above shows we tried to interpret data messages that hadn't been defined.  d7a403a closes this hole, it will throw an exception if we try to interpret an undefined message.  

> GPSBabel Version: 1.4.3
fit: header len=14
fit: protocol version=16
fit: profile version=140
fit: fit_data.len=29564
fit: starting to read data with fit_data.len=29564
.....fit: got data message at fit_data.len=29563 ...local message type 0x0
......fit: parsing fit data ID 0 with num_fields=0
......fit: storing fit data with num_fields=0
.....fit: got data message at fit_data.len=29562 ...local message type 0x0
......fit: parsing fit data ID 0 with num_fields=0
......fit: storing fit data with num_fields=0
.....fit: got definition message at fit_data.len=29561 ...local message type 0x0

PR #562, opened 2020-05-19
let clang-tidy provide default member initializers

tsteven4:
This is the check cppcoreguidelines-pro-type-member-init, but
it was only applied to the class declarations for our filters,
formats and vectors.
This is currently irrelevant as all these are constructed globally,
but if one constructs dynamic instances errors can be seen in gpx, kml, nmea
and lowranceusr formats.

### DIFF
--- a/arcdist.h
+++ b/arcdist.h
@@ -40,7 +40,7 @@ public:
   void process() override;
 
 private:
-  double pos_dist;
+  double pos_dist{};
   char* distopt = nullptr;
   char* arcfileopt = nullptr;
   char* rteopt = nullptr;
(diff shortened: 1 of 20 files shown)


GPSBabelDeveloper:
Another option that's commonly used is to initialize everything in class construction or initialization. I've always found that a bit error prone because it's easy to forget to init a member when adding one and member order initialization is a pain.

Go for it, please. Some day, I'd like all the formats and filters to be constructed dynamically at runtime so we don't have a zillion globals to init on startup.

tsteven4:
Providing a brace or equal initializer when declaring the variable 
protects against constructors that fail to initialize that variable.

See 
https://arne-mertz.de/2015/08/new-c-features-default-initializers-for-member-variables/

PR #564, opened 2020-05-20
Fail configuration if libusb-1.0 is not found

### LINKED ISSUE #563: undefined reference to `gusb_init(char const*, gpsdevh**)
/usr/bin/ld: jeeps/gpsdevice_usb.o: in function `gdu_on(char const*, gpsdevh**)':
/home/rick/proj/gpsbabel/jeeps/gpsdevice_usb.cc:37: undefined reference to `gusb_init(char const*, gpsdevh**)'
collect2: error: ld returned 1 exit status
make: *** [Makefile:156: gpsbabel] Error 1

tsteven4:
The configuration flow configuration will now fail if libusb-1.0 isn't found and libusb usage hasn't been disabled with configure option --without-libusb.  This prevents subsequent link failures where the cause is not readily apparent.  This is not relevant for macos builds as libusb is included in gpsbabel on that platform.

With the qmake flow libusb-1.0 is assumed to be available, and compilation will fail if the libusb-1.0 headers are not found.  This behavior is unchanged.

This resolves #563 .

### DIFF
--- a/configure
+++ b/configure
@@ -6328,7 +6328,7 @@ as_fn_error $? "libusb-1.0 not found (--without-libusb to disable)
 See \`config.log' for more details" "$LINENO" 5; }
 fi
 
-			OSJEEPS=jeeps/gpslibusb.o
+		OSJEEPS=jeeps/gpslibusb.o
 
 fi
  ;;
--- a/configure
+++ b/configure
@@ -728,6 +728,7 @@ enable_csv
 enable_most
 enable_filters
 with_zlib
+with_libusb
 with_doc
 '
       ac_precious_vars='build_alias
@@ -1377,6 +1378,7 @@ Optional Packages:
   --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
   --with-zlib=(included)|system|no
 
+  --without-libusb        disable support for libusb
   --with-doc=DIR          Path where the documentation will be stored.
 
 Some influential environment variables:
@@ -6261,17 +6263,21 @@ fi
   *) :
 
 	GBSER=gbser_posix.o
-	{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for libusb-1.0" >&5
-$as_echo_n "checking for libusb-1.0... " >&6; }
+
+# Check whether --with-libusb was given.
+if test "${with_libusb+set}" = set; then :
+  withval=$with_libusb;
+else
+  with_libusb=yes
+fi
+
 	if test "$with_libusb" = "no"; then :
 
-		{ $as_echo "$as_me:${as_lineno-$LINENO}: result: check not done" >&5
-$as_echo "check not done" >&6; }
 		OSJEEPS=jeeps/gpsusbstub.o
 
 else
 
-
+    # Die if libusb-1.0 not found. This prevents link failures later.
 		{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for libusb_init in -lusb-1.0" >&5
 $as_echo_n "checking for libusb_init in -lusb-1.0... " >&6; }
 if ${ac_cv_lib_usb_1_0_libusb_init+:} false; then :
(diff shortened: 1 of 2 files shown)


GPSBabelDeveloper (commented):
I'm torn. Is it better to let the build proceed without it silently and "just" encounter failures at runtime if someone tries to use it or should we be forcing the default configuration to depend on libusb?

Input from our packagers particularly welcome here. But you can approve this exactly as it stands.

Our percentage of users that need libusb over time is only going to fall. GPS vendors decided/learned that mass storage is a much better way to go.

tsteven4:
The previous behavior was to fail mysteriously at link time with "undefined reference to gusb_init".  gusb_init is usually supplied by our jeeps/gpsusbstub or jeeps/gpslibusb (or jeeps/gpsusbwin).  When libusb-1.0 wasn't found, and the user hadn't set the variable with_libusb to no, then, on linux we would compile with jeeps/gpslibusb.  We did this irregardless of weather libusb-1.0 and it's include files were found.  However, we would set HAVE_LIBUSB_1_0 only if libusb-1.0 was found.  jeeps/gpslibusb doesn't do anything if HAVE_LIBUSB_1_0 is undefined.  Hence the link failure when libusb-1.0 wasn't found.  gusb_init isn't a libusb-1.0 entry point, it is ours.  Thus it isn't obvious that the link failure is due to libusb-1.0 being missing.

Also note that --without-libusb, or one of its' synonyms, wasn't a configure option.  The only way to avoid the link failure without libusb-1.0 was to set the shell variable with_libusb=no before running configure.

With this PR on linux if libusb-1.0 isn't found and --without-libusb isn't supplied (or one of equivalent it's synonyms) to configure the packager will see the following, and error status will be set.
```checking for libusb_init in -lusb-1.0... no
configure: error: in `/home/ubuntu/gpsbabel':
configure: error: libusb-1.0 not found (--without-libusb to disable)
See `config.log' for more details
root@focal:/home/ubuntu/gpsbabel# echo $?
1
```
The error message is as suggested from the third example for AC_ARG_WITH in the autoconf manual.  It is much clearer than the link failure.  It requires action.  I think this is especially good because libusb-1.0 is a new dependency.  The intent is the packager sees this message and either supplies libusb-1.0, or disables it and accepts reduced functionality.

An alternative would be to (nearly) silently build with or without libusb-1.0 depending on if we found it.  The only indication of what we were doing would be the line in the configure standard output
```
checking for libusb-1.0... checking for libusb_init in -lusb-1.0... no
```
This alternative may be better than generating a link error as master does now.  But downstream a user would generate a run time error if they attempted to use a garmin usb serial device.  The packager may not even be aware that libusb-1.0 was a dependency that wasn't found and thus the generated package omitted this functionality.

The previous communication we have had from packagers is that they do not want to supply libusb 0.1, but that they do have libusb 1.0 available.  If one didn't want to supply libusb 1.0 then they could accept the reduced functionality by using the --without-libusb configure option.  I think this is preferable to the variety of inventive patches they will come up with to get what they want.  Those patches risk living forever, even after they may no longer be required.  Those patches also make it hard for us diagnose problems that get reported to use because we are ignorant of them.

GPSBabelDeveloper:
On Thu, May 21, 2020 at 7:01 AM tsteven4 <notifications@github.com> wrote:

> The previous behavior was to fail mysteriously at link time with
> "undefined reference to gusb_init".
>
That's clearly bad. I didn't mean "previous" like "last week", but rather
at the last release.

Also note that --without-libusb, or one of its' synonyms, wasn't a
> configure option. The only way to avoid the link failure without libusb-1.0
> was to set the shell variable with_libusb=no before running configure.
>
There was a time that we only used the HAVE_LIBUSB case if we, well, had
LIBUSB and we used the stub otherwise. I build on Linux infrequently enough
(certainly lately - and tend to have robust development systems when I do)
that I probably hadn't checked this in a while.



> An alternative would be to (nearly) silently build with or without
> libusb-1.0 depending on if we found it. The only indication of what we were
> doing would be the line in the configure standard output
>
> checking for libusb-1.0... checking for libusb_init in -lusb-1.0... no
>
> That is how we once worked. Then someone doing a weird mingw or cygwin
build didn't have to chase down the USB code if they were just converting
files.

I also, at least once, made a binary release that had LIBUSB disabled just
because I didn't have it. So I get the argument against the "silently, 99%
working" approach. It's awesome unless you're in that 1%. (Numbers
fictitious.)

> but that they do have libusb 1.0 available. If one didn't want to supply
> libusb 1.0 then they could accept the reduced functionality by using the
> --without-libusb configure option. I think this is preferable to the
> variety of inventive patches they will come up with to get what they want.
> Those patches risk living forever, even
>
Fair point. We're out of the Linux binary biz, so punt this to the
packagers and let them make the choice explicitly.

OK, Sold.
Thanx for talking it through.

(Be prepared to link to this conversation when it comes up. :-)

PR #565, opened 2020-05-21
fix a few -Wstringop-truncation warnings

tsteven4:
thanks to @rickrich for pointing out the bushnell fix.


GPSBabelDeveloper (approved):

tsteven4:
@gpsbabel, please look at garmin.cc, I believe there was a real bug there as explained in https://github.com/gpsbabel/gpsbabel/pull/565/commits/903ea4cbb772fd8c4e700f25e784c7cb759aa3ba

tsteven4:
Those memcpy operations in garmin have another unlikely bug, they can overwrite the structure member for source > 256 characters.  They were introduced in https://github.com/gpsbabel/gpsbabel/commit/467e373cda78a76a6d2e6f49bdbfe7dc02ef0a69.  This could easily be fixed but...

I am going to call it on garmin and go with it as it stands.  It's better than what we had and gcc is happier.

robertlipe:
Good call. This kind of thing leads to whack-a-mole too easily. Take the
victory and run!

PR #568, opened 2020-05-24
fix -Wstringop-truncation warning in osm.

tsteven4:
The values hash key used to be a QString formed from the concatenation of an integer and a constant length NULL filled character string.  Instead we just use a QPair consisting of an integer and a QString formed from the character string without padding.  padding and constant length are unnecessary and likely inefficient.

### DIFF
--- a/osm.cc
+++ b/osm.cc
@@ -20,11 +20,22 @@
 
 */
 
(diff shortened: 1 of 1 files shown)


GPSBabelDeveloper (approved):
Nice!

PR #576, opened 2020-05-29
make garmin_tables const

tsteven4:
and clean up mkicondoc cppcheck errors.


GPSBabelDeveloper (commented):
We have this idiom in our code a lot.  It would be a good starter project for a budding contributor to replace many of our table lookups with something like a QHash.

PR #578, opened 2020-05-29
clean up mkstyle.sh

### LINKED ISSUE #577: custom style was not excluded
One stated intent of https://github.com/gpsbabel/gpsbabel/commit/5e5ef57d1 was to "Exclude custom stye."  However, the style was never excluded because the extension was stripped before the comparison to "custom.style".

We should either drop the attempt at excluding it in mkstyle.sh or correct it.

tsteven4:
as suggested by shellcheck used $() instead of backticks,
and double quote to prevent globbing and word splitting.
mkstyle.sh now works if the directory path contains a space.

drop the failed attempt to exclude custom.style. This resolves #577
drop the exclusion of README.style, which hasn't existed for a long time.

use basename to drop the suffix instead of piping to sed.

### DIFF
--- a/mkstyle.sh
+++ b/mkstyle.sh
@@ -12,34 +12,32 @@ export LC_COLLATE
 # require gnu sed even though we aren't using gnu extensions.
 # this avoids portability issues with other seds.
 if gsed v /dev/null 1>/dev/null 2>&1; then
-    SED=gsed
+  SED=gsed
 elif sed v /dev/null 1>/dev/null 2>&1; then
-	# sed is gnu sed
-    SED=sed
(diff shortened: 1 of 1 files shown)


tsteven4:


```diff

Issues
======
+ Solved 8
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gpsbabel/gpsbabel/pullRequest?prid=5617521&bid=18402624)

GPSBabelDeveloper (approved):
Agreed. The special case for custom hasn't applied in a very long time. Our days of caring about System V /bin/sh are long over.

PR #583, opened 2020-05-31
fix Qt 5.15.0 Wdeprecated-declarations warnings for

tsteven4:
QTime::start()
QTime::elapsed()

### DIFF
--- a/gui/map.cc
+++ b/gui/map.cc
@@ -117,11 +117,11 @@ Map::~Map()
 #ifdef DEBUG_JS_GENERATION
   if (dbgout_) {
     delete dbgout_;
-    dbgout_ = NULL;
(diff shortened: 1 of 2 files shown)


GPSBabelDeveloper (commented):
What if we eliminated the life cycle issues for dbgdata_ and dugout_ by changing them from pointers allocated to just members?

Also, is the WebView code in this file still in play? Have all the Linuxen moved to WebEngine yet?
Is manager_ really unused?

tsteven4:
It is hard to know what people are using wrt webkit vs. webengine.  Ubuntu focal, which just released April 23, 2020, supports both  libqt5webkit5 and  libqt5webengine5.  webengine certainly isn't light weight.

PR #584, opened 2020-05-31
fix Qt 5.15.0 Wdeprecated-declarations warnings for

tsteven4:
QXmlDefaultHandler
QXmlInputSource
QXmlSimpleReader

as suggested use QXmlStreamReader instead.


GPSBabelDeveloper (approved):
This file predates QXmlStreamReader, I think. Good cleanup.

PR #585, opened 2020-06-01
Extract expected travel durations from Garmin BaseCamp GDB output

lintondf:
Autorouted GDB durations are reported as text fields in newly added <rtept/src> elements in gpx output.  <src> elements are only generated for <rtept> elements with non-zero durations.

### DIFF
--- a/gdb.cc
+++ b/gdb.cc
@@ -75,7 +75,7 @@
 
 #define GDB_DEBUG		(GDB_DBG_WPTe) /* | GDB_DBG_RTE) */
 #undef GDB_DEBUG
-// #define GDB_DEBUG 0xff
+#define GDB_DEBUG 0xff
 
 #define DBG(a,b)		if ((GDB_DEBUG & (a)) && (b))
 
@@ -518,7 +518,7 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
   }
   int display = FREAD_i32;
 #if GDB_DEBUG
-  DBG(GDB_DBG_WPTe, i)
+  DBG(GDB_DBG_WPTe, 1)
   printf(MYNAME "-wpt \"%s\" (%d): display = %d\n",
          qPrintable(res->shortname), wpt_class, display);
 #endif
@@ -591,18 +591,17 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
     FREAD(buf, 1);
     unsigned int duration = gbfgetuint32(fin);
 
-   	res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
-    auto wc = garmin_fs_t::get_wpt_class(gmsd, 0);
-    if (wc == gt_waypt_class_map_intersection || wc == gt_waypt_class_map_line) {
-    	garmin_fs_t::set_duration(gmsd, duration);
-    	if (res->description.length() == 0) {
-    		res->description = res->shortname;
-    	}
-    	res->notes = QString("[%1]").arg( gdb_to_ISO8601_duration( duration ) );
+    res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
+    if (wpt_class == gt_waypt_class_map_intersection || wpt_class == gt_waypt_class_map_line) {
+      garmin_fs_t::set_duration(gmsd, duration);
+      if (res->description.isEmpty()) {  //
+        res->description = res->shortname;
+      }
+      res->notes = QString("[%1]").arg( gdb_to_ISO8601_duration( duration ) );
 #if GDB_DEBUG
-        DBG(GDB_DBG_WPTe, 1)
-        printf(MYNAME "-wpt \"%s\" (%d): duration = %u\n",
-          qPrintable(res->shortname), wpt_class, duration);
+      DBG(GDB_DBG_WPTe, 1)
+      printf(MYNAME "-wpt \"%s\" (%d): duration = %u\n",
+             qPrintable(res->shortname), wpt_class, duration);
 #endif
     }
     int url_ct = FREAD_i32;
--- a/gdb.cc
+++ b/gdb.cc
@@ -563,14 +563,15 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
     garmin_fs_t::set_addr(gmsd, fread_cstr());
 
     FREAD(buf, 1);
-    res->duration = (unsigned int) FREAD_i32;
+    unsigned int duration = (unsigned int) FREAD_i32;
+    if (duration != 0) {
+    	garmin_fs_t::set_duration(gmsd, duration);
 #if GDB_DEBUG
-    if (res->duration != 0) {
-      DBG(GDB_DBG_WPTe, 1)
-      printf(MYNAME "-wpt \"%s\" (%d): duration = %u\n",
-          qPrintable(res->shortname), wpt_class, res->duration);
-    }
+        DBG(GDB_DBG_WPTe, 1)
+        printf(MYNAME "-wpt \"%s\" (%d): duration = %u\n",
+          qPrintable(res->shortname), wpt_class, duration);
 #endif
+    }
 
     res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
     int url_ct = FREAD_i32;
--- a/gdb.cc
+++ b/gdb.cc
@@ -303,6 +303,32 @@ gdb_add_route_waypt(route_head* rte, Waypoint* ref, const int wpt_class)
   return res;
 }
 
+QString gdb_to_ISO8601_duration( unsigned int seconds ) {
+	if (seconds == 0u)
+		return QString("PT0S");
+	unsigned int days = seconds / 86400u;
+	QString out = "P";
+	if (days != 0) {
+		out.append( QString("D%1").arg(days));
+		seconds -= 86400u * days;
+	}
+	out.append( QString("T") );
+	unsigned int hours = seconds / 3600u;
+	if (hours != 0) {
+		out.append(QString("%1H").arg(hours));
+		seconds -= 3600u * hours;
+	}
+	unsigned int minutes = seconds / 60u;
+	if (minutes != 0) {
+		out.append(QString("%1M").arg(minutes));
+		seconds -= 60u * minutes;
+	}
+	if (seconds != 0) {
+		out.append(QString("%1S").arg(seconds));
+	}
+	return out;
+}
+
 /*******************************************************************************/
 /* TOOLS AND MACROS FOR THE WRITER */
 /*-----------------------------------------------------------------------------*/
@@ -564,16 +590,21 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
 
     FREAD(buf, 1);
     unsigned int duration = gbfgetuint32(fin);
-    if (duration != 0) {
+
+   	res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
+    auto wc = garmin_fs_t::get_wpt_class(gmsd, 0);
+    if (wc == gt_waypt_class_map_intersection || wc == gt_waypt_class_map_line) {
     	garmin_fs_t::set_duration(gmsd, duration);
+    	if (res->description.length() == 0) {
+    		res->description = res->shortname;
+    	}
+    	res->notes = QString("[%1]").arg( gdb_to_ISO8601_duration( duration ) );
 #if GDB_DEBUG
         DBG(GDB_DBG_WPTe, 1)
         printf(MYNAME "-wpt \"%s\" (%d): duration = %u\n",
           qPrintable(res->shortname), wpt_class, duration);
 #endif
     }
-
-    res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
     int url_ct = FREAD_i32;
     for (int i = url_ct; (i); i--) {
       QString str = FREAD_CSTR_AS_QSTR;
--- a/gdb.cc
+++ b/gdb.cc
@@ -562,7 +562,16 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
 
     garmin_fs_t::set_addr(gmsd, fread_cstr());
 
-    FREAD(buf, 5);				/* instruction depended */
+    FREAD(buf, 1);
+    res->duration = (unsigned int) FREAD_i32;
+#if GDB_DEBUG
+    if (res->duration != 0) {
+      DBG(GDB_DBG_WPTe, 1)
+      printf(MYNAME "-wpt \"%s\" (%d): duration = %u\n",
+          qPrintable(res->shortname), wpt_class, res->duration);
+    }
+#endif
+
     res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
     int url_ct = FREAD_i32;
     for (int i = url_ct; (i); i--) {
--- a/gdb.cc
+++ b/gdb.cc
@@ -563,7 +563,7 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
     garmin_fs_t::set_addr(gmsd, fread_cstr());
 
     FREAD(buf, 1);
-    unsigned int duration = (unsigned int) FREAD_i32;
+    unsigned int duration = gbfgetuint32(fin);
     if (duration != 0) {
     	garmin_fs_t::set_duration(gmsd, duration);
 #if GDB_DEBUG
--- a/gdb.cc
+++ b/gdb.cc
@@ -20,7 +20,7 @@
 	along with this program; if not, write to the Free Software
 	Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
-  A format description obtained from reverse-engineering is at 
+  A format description obtained from reverse-engineering is at
   https://www.memotech.franken.de/FileFormats/Garmin_MPS_GDB_and_GFI_Format.pdf
 */
 
@@ -87,7 +87,7 @@ static char gdb_release_date[] = "$Date: 2011-04-14 01:30:01 $";
 static gbfile* fin, *fout, *ftmp;
 static int gdb_ver, gdb_category, gdb_via, gdb_roadbook;
 
-static QList<Waypoint *> wayptq_in, wayptq_out, wayptq_in_hidden;
+static QList<Waypoint*> wayptq_in, wayptq_out, wayptq_in_hidden;
 static short_handle short_h;
 
 static char* gdb_opt_category;
@@ -112,10 +112,10 @@ static int trk_ct;	/* informational: total number of tracks in/out */
 #define NOT_EMPTY(a) (a && *a)
 
 static void
-gdb_flush_waypt_queue(QList<Waypoint *>* Q)
+gdb_flush_waypt_queue(QList<Waypoint*>* Q)
 {
 
-  while(!Q->isEmpty()) {
+  while (!Q->isEmpty()) {
     const Waypoint* wpt = Q->takeFirst();
     if (wpt->extra_data) {
       // FIXME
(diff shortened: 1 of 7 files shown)


tsteven4:
@lintondf please fix the regression test failure.  For example, see https://travis-ci.com/github/gpsbabel/gpsbabel/jobs/342370570  The usual method is to comment out the trap line in testo, run "./testo gdb", and then copy the noted file from /tmp to the indicated file in reference.  This assumes that the generated file in /tmp is correct.

In the same log there is a -Wreorder warning as well.  However, I am not sure we want to add duration to Waypoint.  For data of limited use we usually add it as format specific data, keeping the Waypoint class to a set of common core data members.  @gpsbabel whats your view on an additional Waypoint data member vs. format specific data?

lintondf:
I didn’t know there was a place for format specific data.   If you could point it out I will relocate the field.  It is certainly not of general use.  Btw. I built on both Mac OS and windows and neither pulled the warning that GCC saw.

lintondf:
Following up:  I found the Garmin format specific data and will move to data field there.   Also was able to build gpsbabel using gcc under WSL Ubuntu and see the warnings which I will also dispatch.

tsteven4:
It might be easier and cleaner to create a new format specific data type.  You could add a FsType in formspec.h.  The complicated looking numbers are just 4 ASCII characters related to the enumerator.  e.g. G = 0x47, M= 0x4d, S=0x53, D=0x44 so kFsGmsd = 0x474d5344L.  The important thing is that they are unique.

lintondf:
It is your house, but the GMSD class already holds fields from exactly the same source elements including some experimental code that saves the Uber occult subclass hex strings.   It looked to me like one gets created for each of the RTEPT waypoints already.

Tell me your decision

tsteven4:
If gmsd goes in cleanly use it.  gmsd has some rough edges related to communication with garmin devices via serial connections, but it isn't obvious to me that you will be drug into that.  (garmin.cc, PWay, jeeps/*).  I was just pointing out an alternative.

tsteven4:
@lintondf regression still needs work.  for example https://travis-ci.com/github/gpsbabel/gpsbabel/jobs/343698942

When you run a test under the testo script we set a variable to indicate to run the test as if the datetime was 1970-01-01T00:00:00Z.  This allows us to compare the gpsbabel output with constant reference files regardless of the time the test was run.  You checked in a reference file that was not run with testo, so it shows a date in 2020 instead of our expected fixed date of 1970-01-01
```
GPSBABEL_FREEZE_TIME=y
export GPSBABEL_FREEZE_TIME
```

gpx.cc
```diff
@@ -1364,7 +1364,13 @@ GpxFormat::gpx_write_common_description(const Waypoint* waypointp, const QString
   } else {
     writer->writeOptionalTextElement(QStringLiteral("desc"), waypointp->description);
   }
-  /* TODO: src should go here */
+  auto* gmsd = garmin_fs_t::find(waypointp); /* gARmIN sPECIAL dATA */
+  if (isRtept && garmin_fs_t::has_duration(gmsd)) {
+	  uint seconds = garmin_fs_t::get_duration(gmsd, 0u);
+	  QString source = "AUTOROUTE:";
+	  source.append( QString("duration=%1;").arg(seconds) );
```
tsteven4: QString source = QString("AUTOROUTE:duration=%1;").arg(seconds);

gdb.cc
```diff
@@ -562,7 +562,17 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
 
     garmin_fs_t::set_addr(gmsd, fread_cstr());
 
-    FREAD(buf, 5);				/* instruction depended */
+    FREAD(buf, 1);
+    unsigned int duration = (unsigned int) FREAD_i32;
```
tsteven4: Although gdb doesn't use it gbfgetuint32 exists, so maybe just
unsigned int duration = gbfgetuint32(fin);

tsteven4 (commented):
I had a few minor suggestions, but this looks good to me.

tsteven4:
@lintondf for the record, is there a reason you picked the gpx src element for this information?  Does some other software place this information there?
thanks

lintondf:
The autoroutes rtept waypoints are generated by Basecamp rather than by the user.   So marking their source as autoroute seemed logical and adding the duration as and then the duration as an annotation followed naturally.

tsteven4:
OK.  My understanding is that we are not trying to duplicate the verbatim output of some other program with the gpx src element, we are simply trying to include the duration content in some manner.  I'm fine with your implementation, I am just trying to establish the requirements should they be needed in the future.  Thanks.

tsteven4:
@lintondf Here are some ideas for you to consider.  I am not asking for any changes, but just for you to consider if you would prefer to implement any of these compared to what you have now.

1. combine the the "instruction" which is stored in rte->description with the duration.  So the description might be something like "Turn left onto Us1 (87)</desc>
This would be picked up by all formats, not just gpx.  This might be more readily picked up by GPS devices that the src element.  It is much simpler to implement.
2. provide units for the duration, e.g. "AUTOROUTE:duration=87s;"
3. drop the semicolon in the duration.
4. adjust the display units based on the magnitude.  I can imagine "Turn left onto I80" with a duration of hours.  For a duration of 2 hours 24 minutes 5 seconds do you want to show 8645 or 8645s or 8645sec or 2:24:05?

GPSBabelDeveloper:
Hi, and Welcome. Thanx for diving in.

Like Steven, I find this a bit odd. Smuggling "real" data in a text field in GPX with the goal of interoperability is a bit strange. Is there anything else that read or writes GPX that expects to see this data there? It seems a bit unlikely that we'd be in the middle of a GDB->Garmin device transaction. If you read this file in, say, ExpertGPS, you'll now have an extra field in the pin on the map - that's good and bad.

Do you know if there are localization issues? If the current language is Farsi or whatever, is the word "autoroute" and the direction expected to change? This is one of the downfalls of smuggling data in text like this instead of in a device extension where it's meant to be read by computers instead of humans.

I'm fine with the implementation. If you've thought all this through and still think there's value in doing it this way, just answer "yes, please" and one of us will push the button to squash and merge. Things can be odd to us (neither of us use Basecamp a lot...) and still be the Right Thing To Do.

Thanx!

lintondf:
I'm  in the opposite state, I pretty much only use Basecamp and the GPX
format.  I certainly don't want to do any violence to Universal Translator
aspects of GPSBabel.

You are correct in that I am putting English text in to the SRC element
when in a perfect world I would add
<extensions><newschema::duration>15</newschema::duration></extensions>.
I'm too lazy to do the full monty.

Here's my alternative:
For RTEPT waypoints, BaseCamp only generates descriptions with no notes.
gpx.cc fills the <cmt> element from description and the <desc> with notes
or if no notes then repeats the description.
What about deleting the new <src> node generation code and putting the
duration into the waypointp->notes in as an ISO 8601 Duration:
P[n]Y[n]M[n]DT[n]H[n]M[n]S
Wikipedia example:  "P3Y6M4DT12H30M5S" represents a duration of "three
years, six months, four days, twelve hours, thirty minutes, and five
seconds".
2 hours, 3 seconds would be: "P2H3S"

Thoughts?

lintondf:
I committed the change but when I tried my full end to end use case the
correct durations are not coming thru.  Please stand by....

lintondf:
Never mind, just a bug in my Java code reading the new format.

For the record, 2 hours, 3 seconds would be PT2H3S, not

If this is acceptable please squash and merge.

gdb.cc
```diff
@@ -562,8 +588,23 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
 
     garmin_fs_t::set_addr(gmsd, fread_cstr());
 
-    FREAD(buf, 5);				/* instruction depended */
-    res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
+    FREAD(buf, 1);
+    unsigned int duration = gbfgetuint32(fin);
+
+   	res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
+    auto wc = garmin_fs_t::get_wpt_class(gmsd, 0);
```
tsteven4: No need to look this up, you already have the waypoint class in the local variable wpt_class.

gdb.cc
```diff
@@ -562,8 +588,23 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
...
+    if (wc == gt_waypt_class_map_intersection || wc == gt_waypt_class_map_line) {
+    	garmin_fs_t::set_duration(gmsd, duration);
```
tsteven4: Is there a reason to save the duration in gmsd?  Nobody is reading it.  If you take this out you can revert garmin_fs.h as well.

One reason to leave it in is it would allow for formats to utilize the data in the future.

gdb.cc
```diff
@@ -562,8 +588,23 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
 
     garmin_fs_t::set_addr(gmsd, fread_cstr());
 
-    FREAD(buf, 5);				/* instruction depended */
-    res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
+    FREAD(buf, 1);
+    unsigned int duration = gbfgetuint32(fin);
+
+   	res->description = FREAD_CSTR_AS_QSTR;	/* instruction */
```
tsteven4: can you please change indentation to use 2 spaces?   You can fix this, and the other indentation/alignment issues below with "astyle --options=astylerc gdb.cc".  If you use astyle you'll pick up some other formatting changes, mostly whitespace.

Using spaces ensures tools like github diff keep things aligned.

gdb.cc
```diff
@@ -562,8 +588,23 @@ read_waypoint(gt_waypt_classes_e* waypt_class_out)
...
+    auto wc = garmin_fs_t::get_wpt_class(gmsd, 0);
+    if (wc == gt_waypt_class_map_intersection || wc == gt_waypt_class_map_line) {
+    	garmin_fs_t::set_duration(gmsd, duration);
+    	if (res->description.length() == 0) {
```
tsteven4: res->description.isEmpty() is preferred instead of checking the length and comparing.
https://github.com/KDE/clazy/blob/master/docs/checks/README-isempty-vs-count.md

I guess your trying to not rely on the fill in by WaypointList::waypt_add(Waypoint* wpt) which would have used notes before shortname.

gdb.cc
```diff
@@ -303,6 +303,32 @@ gdb_add_route_waypt(route_head* rte, Waypoint* ref, const int wpt_class)
   return res;
 }
 
+QString gdb_to_ISO8601_duration( unsigned int seconds ) {
```
tsteven4: unfortunate that this isn't provided in Qt or the standard library.

tsteven4 (commented):
back to the minutia.   If you don't have astyle available I think I could push those changes to lintondf/gpsbabel for you (unless you changed the default to disallow maintainers from making changes).

lintondf:
Using the local variable for waypoint class
I changed to isEmpty(); sorry it has been a thousand years since I touched
Qt
I think I changed the tabs to spaces correctly but feel free to run astyle
and push if I didn't get them all
I left the duration additions to the Garmin-specific data since someone in
the future might want to output it in a true machine readable format
without having to reinterpret the ISO string.
I was surprised Qt didn't support ISO 8601 durations; the Java Duration
class does it quite neatly.  Boost might but I didn't want to go there and
C++20 <chrono> might.  The lack of Qt to/from ISO string also seems a good
reason to leave the durations in the Garmin data structure.

tsteven4:
We are using the c++14 standard.  We don't want to add boost.  Qt does pretty well with string to date/time conversions, but I was surprised by the lack of duration support as well.

I think were good.  Assuming regression runs I will merge.  Thanks.

lintondf:
Thanks for all of the support and patience.  I think the final result is
much better.

I have a Garmin RV 760 and use BaseCamp on both Mac and PC.  If I can be of
any help in the future please let me know

tsteven4:
@lintondf I am having a little buyers remorse over the gpx part of this PR.  I don't think we want the gpx writer deciding to ignore certain waypts.  Note every other writer won't ignore them.  I think it would be better to:

1. always leave them in, or
2. always filter them out in the gdb reader, or
3. filter them out in the gdb reader based on a gdb option.

tsteven4:
@lintondf looking at gdb I see some filtering is already occurring with the via or roadbook option.  I was dismayed to see that there are related tests for descritipion.isEmpty().  I am concerned that your 
addition here (with a missing comment):
https://github.com/gpsbabel/gpsbabel/blob/84445025b4fd793628eeea9cfb18caf02f838f3f/gdb.cc#L600-L602
will interact with these existing options.

I also don't understand why you filtered based on gt_waypt_class_map_intersection or gt_waypt_class_map_line.  There seem to be lots of existing decisions based on relational operators between waypt class and gt_waypt_class_map_point.

lintondf:
The simple answer is that it gave the GPX file that most closely matched
what BaseCamp itself output.

I have not considered the interactions you referred to but will later today.


For a simple but non-trival route, from BaseCamp I saved both a GDB and GPX
file.

Summarizing the BaseCamp GPX file:


   1.

   <metadata>
   2.

   <wpt> # origin of route
   3.

   <wpt> # destination
   4.

   <rte>
   1.

      <rtept>
      1.

         With 203 <gpxx:rpt> subelements 18 with Subclasses
         2.

      <rtept> destination
      1.

         Subclass: 000000000000FFFFFFFFFFFFFFFFFFFFFFFF


I then ran GPSBabel with the change in gpx.cc that excludes
gt_waypt_class_map_intersection and gt_waypt_class_map_line waypoint
classes.

./gpsbabel -i gdb -f ~/Documents/ShortTrip.GDB -o gpx,garminextensions -F
~/Documents/ShortTrip-GPSBabel-After.gpx

Summary of GPSBabel GPX file:

   1.

   <metadata>
   2.

   <wpt> # origin of route
   3.

   <wpt> # destination
   4.

   <rte>
   1.

      <rtept> # origin
      2.

      18 <rtept> corresponding to the gpxx:rpt elements with subclasses
      3.

      <rtept> # destination


I removed the gpx.cc change and ran GPSBabel again:

   1.

   <metadata>
   2.

   18 <wpt> corresponding to the gpxx:rpt elements with subclasses
   3.

   <wpt> # origin of route
   4.

   <wpt> # destination
   5.

   <rte>
   1.

      <rtept> # origin
      2.

      18 <rtept> corresponding to the gpxx:rpt elements with subclasses
      3.

      <rtept> # destination

tsteven4:
I have an experimental branch with the filtering inside the gdb reader instead of the gpx writer.  gdb-sample-v3-autoroute.gpx is the same except I drop 1 cmt element without your copy of the shortname to the description.  It requires the drop option to the gdb reader to get this close.

This isn't ready for prime time, I think the help and name of the drop option needs work, as well as a new documentation file in xmldoc/format/options describing the new option.  But I think it will help this discussion.

https://github.com/tsteven4/gpsbabel/tree/gdbfilter

lintondf:
Additional data point.  The GPX file output by basecamp that matches the
GDB file can not be read by GPSBabel?

./gpsbabel -i gdb -f ~/Documents/ShortTrip.GPX -o gpx,garminextensions -F
x.gpx

gdb: Invalid file "/Users/lintondf/Documents/ShortTrip.GPX"!

tsteven4:
change the input  option to gpx.

> ./gpsbabel -i gdb -f ~/Documents/ShortTrip.GPX -o gpx,garminextensions -F
x.gpx

lintondf:
I can’t find the Homer Simpson emoji....

lintondf:
I think this is quite good as is.  You keep the cmt element that include
turn directions for all maneuvers and preserve all of the durations in desc
elements.  Nice.

Maybe editwaypoints, trimwaypoints, drophiddenwaypoints or
onlyuserwaypoints (inverted of course)?
As to help:  "Omit hidden waypoints" "Include only user waypoints", "Omit
automatically generated waypoints"

Part of the problem is that to BaseCamp these are gpxx:rtept elements not
waypoints per se, so the user of BaseCamp probably isn't thinking that way.
Maybe "waypoints/route points"

As I left things:

grep cmt ~/Documents/ShortTrip-GPSBabel-After.gpx

    <cmt>Schoodic Woods Campground

207-288-1300</cmt>

    <cmt>Sullivan Twn</cmt>

      <cmt>Sullivan Twn</cmt>

      <cmt>Get on Us1 and drive east</cmt>

      <cmt>0002</cmt>

      <cmt>0003</cmt>

      <cmt>0004</cmt>

      <cmt>0005</cmt>

      <cmt>0006</cmt>

      <cmt>Turn right onto S Gouldsboro Rd</cmt>

      <cmt>0008</cmt>

      <cmt>0009</cmt>

      <cmt>0010</cmt>

      <cmt>0011</cmt>

      <cmt>0012</cmt>

      <cmt>0013</cmt>

      <cmt>0014</cmt>

      <cmt>Turn left onto Main St</cmt>

      <cmt>0016</cmt>

      <cmt>Turn right onto Moore Rd</cmt>

      <cmt>0018</cmt>

      <cmt>Schoodic Woods Campground

207-288-1300</cmt>
With your change:

grep cmt ~/Documents/ShortTrip-GPSBabel-Filter.gpx

    <cmt>Schoodic Woods Campground

207-288-1300</cmt>

    <cmt>Sullivan Twn</cmt>

      <cmt>Sullivan Twn</cmt>

      <cmt>Get on Us1 and drive east</cmt>

      <cmt>Turn right onto S Gouldsboro Rd</cmt>

      <cmt>Turn left onto Main St</cmt>

      <cmt>Turn right onto Moore Rd</cmt>

      <cmt>Schoodic Woods Campground

207-288-1300</cmt>

tsteven4:
Are you ok with the loosing the one cmt element in the test?
```
+++ /tmp/gpsbabel.6444/gdb-sample-v3-autoroute.gpx	2020-06-08 11:31:09.834177621 -0600
@@ -59,7 +59,6 @@
     <rtept lat="44.521506736" lon="-68.204822838">
       <ele>40.180</ele>
       <name>0004</name>
-      <cmt>0004</cmt>
       <desc>[PT0S]</desc>
       <sym>Waypoint</sym>
     </rtept>
```

I also wonder if we should use the same class filter as roadbook when deciding if we should set duration and notes, i.e.
`wpt_class > gt_waypt_class_map_point`
instead of
`wpt_class == gt_waypt_class_map_intersection || wpt_class == gt_waypt_class_map_line`

With the test code in my gdbfilter branch durations are funny with the roadbook option.  That filters out points without a description, but those points can have non-zero durations.  Your left with the duration of the points with descriptions, but those durations won't add up to the length of the trip because of the dropped points.

We could use a more realistic (i.e. longer) sample instead of gdb-sample-v3-autoroute.gdb that has a mix of points with and without descriptions.  It would also be good to throw in a third destination, i.e.  I want to go from A to B to C.  

p.s. a command like:
```
 gpsbabel  -r -i gdb -f gdb-sample-v3-autoroute.gdb -x validate,debug -o unicsv -F -
```
is useful to see what is going on in a concise form.  It either shows routes, tracks or waypoints.  -r for routes, -t for tracks, -w for waypoints.

lintondf:
Dropping non-zero duration points would not work for me.

I'll add a stop and a via point to the ShortTrip BaseCamp file and test
it.  I'll send it along an you can add it or replace the current autoroute
gdb.

tsteven4:
Further development on this topic is in #595.

tsteven4:
> I'll add a stop and a via point to the ShortTrip BaseCamp file and test
it. 

Also, please try to get some longer legs so there are some points without descriptions and some points with descriptions in the route.  I think there is one point without a description now at the end that has zero duration, I would like to see more zero length description points interspersed with the non-zero length descriptions in the route.  I think you can do this by creating a longer route.

You can see something like this in this german reference file:
./gpsbabel -r -i gdb -f reference/gdb-sample-v3.gdb -x validate,debug -o unicsv -F -

PR #587, opened 2020-06-02
use http for upgrade

### LINKED ISSUE #580: Download failed: TLS initialization failed.
When using recent pre-release installers on Windows 10 I get a popup when starting the GUI that says
"TLS initialization failed"

Annotation 2020-05-30 103429

tsteven4:
This resolves #580 

Other solutions may be possible.  Deploying openssl on windows should work but seems painful.

### DIFF
--- a/gui/upgrade.cc
+++ b/gui/upgrade.cc
@@ -52,7 +52,7 @@ UpgradeCheck::UpgradeCheck(QWidget* parent, QList<Format>& formatList,
   QObject(parent),
   manager_(nullptr),
   replyId_(nullptr),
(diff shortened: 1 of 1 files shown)


GPSBabelDeveloper (approved):
There's a special case on the server to NOT redirect that one url to https so I think this is ok.

How does Windows ship with broken SSL in modern times? No, don't tell me.

tsteven4:
They have SSL, just not openssl which Qt links to.

tsteven4:
The change to use https was in #450, jan 10, 2020.  release 1.6.0 was still using http.

PR #595, opened 2020-06-08
Adjust gdb reader duration handling

tsteven4:
Instead of filtering out waypoints corresponding to gt_waypt_class_map_intersection or gt_waypt_class_map_line in the gpx writer an option is added to the gdb reader to avoid creation of these waypoints in the first place.  The new gdb option dropwpt will only create waypoints from user points, i.e. gt_waypt_class_user_waypoint.  This exclusion criteria match the criteria already used to designate waypoints as hidden as opposed to the more exclusive filter that is removed from the gpx writer.   In the limited reference files we have the effect is identical.

Also, in #585, an undesired interaction was created when process duration data that could lead to an empty waypoint description being filled in by the shortname.  This compromised the roadbook option, which is checking for an empty waypoint description.


lintondf:
longer multi-stop gdb test file and reference result
[Archive.zip](https://github.com/gpsbabel/gpsbabel/files/4753789/Archive.zip)
I tested end to end and this looks good

tsteven4:
Thanks.  I switched to the new reference files which capture more complexity.  They also had some different waypoint classes represented.

lintondf:
I looked at the write side of gdb.cc and it would be trivial to write the duration from the Garmin specific data.   Would that be of any use?

tsteven4:
I think in this case the reader is much more important than the writer.  It takes some faith to allow us to write a reverse engineered binary format when we don't understand a bunch of the fields.  Without a user actually needing to write duration I wouldn't bother.  I think the knowledge of how to add duration to the writer is captured in the reader so I don't think we need to worry we will forget how if it becomes important some day.

When we use the roadbook or via option which can pare the rtepts in a route we throw away  durations for the pared points as well.  Thus the remaining durations don't add up to the trip time.  That may be considered a defect, but I am not going to chase it.  I also think the roadbook option should make sure to keep some points like the start, end and any intermediate destinations but I don't want to chase that either.  At this point it isn't clear how much roadbook is used or what the exact intention was.  The original developer didn't add a test case and the roadbook option remains untested to this day.

I'd say we spent plenty of effort on this rather old format.  If your happy I think we are good.  Thanks for your contribution.

xmldoc/formats/options/gdb-dropwpt.xml
```diff
@@ -0,0 +1 @@
+
```
GPSBabelDeveloper: A sentence or two here, please.

lintondf:
I'm good.  Thanks again for the support

PR #604, opened 2020-07-08
fix spelling errors flagged by https://lintian.debian.org/sources/gps…

tsteven4:
…babel/1.6.0+ds-11.html

also remove gpsbabel executable that should never have been checked in.


GPSBabelDeveloper (approved):
Thank you.

PR #606, opened 2020-07-13
Make the map optional

City-busz:
This makes the webengine/webkit dependency optional.

### DIFF
--- a/gui/advdlg.cc
+++ b/gui/advdlg.cc
@@ -37,7 +37,11 @@ AdvDlg::AdvDlg(QWidget* parent,
 {
   ui_.setupUi(this);
   ui_.synthShortNames->setChecked(synthShortNames);
+#if HAVE_WEBENGINE || HAVE_WEBKIT
   ui_.previewGmap->setChecked(previewGmap);
+#else // HAVE_WEBENGINE || HAVE_WEBKIT
+  ui_.previewGmap->hide();
+#endif // HAVE_WEBENGINE || HAVE_WEBKIT
   ui_.debugCombo->setCurrentIndex(debugLevel+1);
 #if defined (Q_OS_WIN)
   ui_.buttonBox->button(QDialogButtonBox::Ok)->setIcon(QIcon(":/images/ok.png"));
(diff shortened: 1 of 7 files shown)


robertlipe:
When do you not have either one?

We're really trying to move away from option explosion, not to more of
them.

City-busz:
- Qt WebEngine is a huge dependency. GPSBabel is a tiny tool with minimal dependencies except this one. Not everyone wants to install it just to convert between GPS formats.
- The map uses the proprietary Google Maps JavaScript library, and some people tries to avoid using proprietary software.

tsteven4:
If we accept the idea of having a configuration without webkit/webengine I want to review the implementation and consider other methods of achieving that goal.

tsteven4:
I think your patch is from archlinux.  It is helpful for us to know your are representing a packaging effort by a distribution.
https://git.archlinux.org/svntogit/community.git/commit/trunk?h=packages/gpsbabel&id=c7b3e3b2b74a4633016286a7743a90764e05a461
Thanks for your work with gpsbabel on this distribution.

If you don't want the GUI and the CLI is sufficient for your distribution you could distribute gpsbabel and leave gpsbabelfe and its dependencies out, or package them separability so the user can decide.

If you still want to distribute the GUI, but without a requirement on webengine even though archlinux has the webengine package we would strongly prefer a different implementation.  If that is the case I would like you to review the alternative implementation to make sure it meets your needs.

City-busz:
Yes, I'm the current packager of GPSBabel for Arch Linux.

I would prefer to provide for the users a tiny GUI without forcing them to install the heavy Qt WebEngine dependency. Any solution could be okay for me:
* Disable the map feature at build time (this is what I proposed in this PR).
* Move the map feature into a plugin, which is loaded optionally on runtime depending on the availability of Qt WebEngine.
* Open the map in the default web browser instead of using Qt WebEngine to embed the browser.
* Use a custom map widget implementation instead of requiring a browser, like in [GPXSee](https://github.com/tumic0/GPXSee/tree/master/src/map).

tsteven4:
@City-busz Thanks for your list of potential solutions.  An alternative solution that also disables the map feature at build time is #635.  I would prefer #635 over this PR.  Does that work for you?

tsteven4:
Thanks for the PR and your ideas.  However, I merged #635 instead.  The functionality should be identical. You need to
qmake CONFIG+=disable-mappreview app.pro
to disable the map preview.

PR #607, opened 2020-07-21
Fix NMEA parsing creating Null Island waypoints

viettaml:
Fix parsing of NMEA [GPGGA](https://www.gpsinformation.org/dale/nmea.htm#GGA) sentences: ones with 0 "fix quality" (i.e. "invalid") are allowed in serial (i.e. live GPS) mode because (according to the comment) some GPS devices will report previously-read data in the absence of a current good fix. Adjust this allowance to require an actual coordinate value; at least one popular USB GPS device will issue GPGGA such as `$GPGGA,010222.00,,,,,0,00,99.99,,,,,,*65` if it loses a good fix, and the empty lat/lng coordinates (`,,,,,`) are parsed as 0N 0W (a.k.a. "Null Island").

[`QString::toDouble()`](https://doc.qt.io/qt-5/qstring.html#toDouble)'s optional `bool *ok` argument won't report a problem with an empty input so we simply check for 0 lat/lng.

### DIFF
--- a/nmea.cc
+++ b/nmea.cc
@@ -402,11 +402,13 @@ NmeaFormat::gpgga_parse(char* ibuf)
   double hms = 0;
   if (fields.size() > 1) hms = fields[1].toDouble();
   double latdeg = 0;
-  if (fields.size() > 2) latdeg = fields[2].toDouble();
+  bool lat_ok = false;
+  if (fields.size() > 2) latdeg = fields[2].toDouble(&lat_ok);
   QChar latdir = 'N';
   if (fields.size() > 3) latdir = fields[3][0];
   double lngdeg = 0;
-  if (fields.size() > 4) lngdeg = fields[4].toDouble();
+  bool lng_ok = false;
+  if (fields.size() > 4) lngdeg = fields[4].toDouble(&lng_ok);
   QChar lngdir = 'W';
   if (fields.size() > 5) lngdir = fields[5][0];
   int fix = fix_unknown;
@@ -431,7 +433,7 @@ NmeaFormat::gpgga_parse(char* ibuf)
    * that is more comfortable than nothing at all...
    */
   CHECK_BOOL(opt_ignorefix);
-  if ((fix <= 0) && (read_mode != rm_serial || (latdeg == 0.0 && lngdeg == 0.0)) && (!opt_ignorefix)) {
+  if ((fix <= 0) && (read_mode != rm_serial || !lat_ok || !lng_ok) && (!opt_ignorefix)) {
     return;
   }
 
--- a/nmea.cc
+++ b/nmea.cc
@@ -426,11 +426,12 @@ NmeaFormat::gpgga_parse(char* ibuf)
 
   /*
    * In serial mode, allow the fix with an invalid position through
+   * (unless if the position lat/lng is absent or completely bogus)
    * as serial units will often spit a remembered position up and
    * that is more comfortable than nothing at all...
    */
   CHECK_BOOL(opt_ignorefix);
-  if ((fix <= 0) && (read_mode != rm_serial) && (!opt_ignorefix)) {
+  if ((fix <= 0) && (read_mode != rm_serial || (latdeg == 0.0 && lngdeg == 0.0)) && (!opt_ignorefix)) {
     return;
   }
 
--- a/nmea.cc
+++ b/nmea.cc
@@ -402,13 +402,11 @@ NmeaFormat::gpgga_parse(char* ibuf)
   double hms = 0;
   if (fields.size() > 1) hms = fields[1].toDouble();
   double latdeg = 0;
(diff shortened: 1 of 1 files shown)


tsteven4:
Interesting.  While unlikely, a visit to the null island is legitimate.

https://bugreports.qt.io/browse/QTBUG-32406 seems to imply ok should be false if the string is empty.

Your statement 

> QString::toDouble()'s optional bool *ok argument won't report a problem with an empty input so we simply check for 0 lat/lng.

seems to be incorrect.

With Qt 5.12.9 an empty string does set ok to false:
>
 #include <QtCore/QString>
#include <QtCore/QDebug>
int main()
{
QString str = "";
bool ok;
double d = str.toDouble(&ok);
qDebug() << ok;
}
$ ./a.out
false

viettaml:
> Interesting. While unlikely, a visit to the null island is legitimate.

True, but what are the odds that the fix quality is "invalid" _and_ the last recorded coordinate was _exactly_ 0N 0E?

In any case if using `ok` works (I tested using Google Earth Pro's build which uses Qt 5.5.1, so maybe this was a Qt bug that was fixed later) then it's a moot point as we should use that instead (as I first tried to do, unsuccessfully).

tsteven4:
The confusion over ok was likely not due to Qt 5.5.1, but due to the empty field being replaced by "0" before gga_parse.  This replacement seems to be inappropriate, but it has been there since d08000a6e65dab3c4d136142725b951aed786b72
The replacement means ok is not tripped and your second commit d226d01c862606615aa599f5d3b649cea12f0700 does not do what was intended, while your original commit 71d4d02d324572300e6387558897bf9a98041bf0 did.

https://github.com/gpsbabel/gpsbabel/blob/439948a0113ba684668dbd9df2e261e499a648c2/nmea.cc#L914-L921

It seems like we have to choose between ripping off the bandaid replacement and fixing all the parsers, or throwing out the null island, or just using the ignore_fix option that is already provided.  The ignore_fix option was added in 1328f5013b6f61603bba353f0193e1726e8b6309  It seems this issue is why the ignore_fix option was added.  @gpsbabel can you remember the motivation for ignore_fix?

Even though this is just a couple of LOC, this illustrates the need for a test case.

robertlipe:
Gross. This 15 year old stuff sneaks up on us. We certainly had naive views
of NMEA then.

I'm fine with Tam's first CL. I simply don't care about legit sailors on
Null Island. There's just too much broken stuff in NMEA in this area to
care about. It's kind of a problem that happens at the desk of developers
with GPSes that have never actually been outside. (I have several of
those...)

While that gstrsub is almost certainly misguided in our post-sscanf world.
I was looking at neighboring code and wondering how lines like
if (fields.size() > 3) latdir = fields[3][0];
would work if the fields array had an length of three but the element was
empty; we surely couldn't deref that array. Shouldn't that be a
fields[3].at(0) or a fields.size() > 3 && fields[3].length() > 0 ?  We have
a trainload of occurrences of that. That strsub probably protects us from a
lot of busywork in all our parsers.

ignore_fix was because we had a rash of cameras that wrote every sentence
with an invalid checksum. Every Single One. it's in the release notes at
https://www.gpsbabel.org/news/release-1.4.4.html (Fall of 2012) -

See
[Gpsbabel-code] Problems with Canon NMEA files from GP-E2







[Gpsbabel-misc] Convert Canon PowerShot SX260 HS to GPX
[Gpsbabel-misc] Converter stopped working


[Gpsbabel-misc] NMEA 0183 to GPX



These were all in rapid succession. So once again, our free software became
the bugfix dumping ground for GPS manufacturers that couldn't read specs
and support their own products.

I also see in my notes that this "fix" was pushed live hours before I was
diagnosed and scheduled  for a spinal surgery, so I probably had a tendency
to just end the conversation and not push it back to Canon and tell them to
make better products. :-/

It was kind of like our issue with NMEA CRC checksums, Steven. On a noisy
serial line, they may be important but on an error-correcting USB
connection or a disk file, they were so unimportant that we were dusting up
more problems checking them than just ignoring them sometimes.

So I think Tam's first fix is fine to apply.
The strsub /,,/,0/ is shocking - I didn't know it was still there when I
reworked all those parsers to use QStringSplit instead of sscanf. Maybe we
just keep it.

As for test cases, you're right that I should have insisted on collecting
more of that janky NMEA through the years for formalized testing for it.
Sorry. But here we are...talking about fifteen year old code at 4am.

RJL

viettaml:
Thanks Robert. I certainly noticed that a lot of the rest of the NMEA parsing code could--how do I put it diplomatically--benefit from a bit of coding love, and the temptation to embark on an overhaul did cross my mind. But then I realized that there's likely a big difference between "NMEA according to the spec" and "NMEA as implemented in the real world" and without a decent library of samples from real-world devices I would be doomed to repeating past trials and errors.

PR #609, opened 2020-07-25
support doc creation in qmake.

tsteven4:
pass parameters to tools/mkcapabilites instead of
using text replacement in configure.

text replacement in xmldoc/makedoc was unnecessary, directory
creation is handled in the recipee.

there is some foolishness in GPSBabel.pro to allow overriding of
WEB and DOCVERSION on the command line of make.  If those values
are known when running qmake it could be slightly simpler.

### DIFF
--- a/GPSBabel.pro
+++ b/GPSBabel.pro
@@ -331,15 +331,16 @@ QMAKE_EXTRA_TARGETS += index.html
 
 #
 # The .fo and PDF versions depend on additional tools.
-# 'fop' must be obtained from http://xmlgraphics.apache.org/fop/
+# 'fop' must be obtained from your distribution or http://xmlgraphics.apache.org/fop/
 #   0.92beta seems to work OK, BUT.
 #   * If you have a package called 'docbook-xml-website' it's reported
 #     to prevent the build from working.   Remove it. (Suse)
 #   * Sun Java seems to be required.   GCJ 1.4.2 doesn't work.   You can
 #     force Sun Java to be used by creating ~/.foprc with 'rpm_mode=' (Fedora)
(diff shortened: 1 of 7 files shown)


tsteven4:
This eliminates one of the uses of configure, moving us towards its deprecation/removal.
The other issue is library linking which will be addressed separately.

tsteven4:
another issue to be fixed before we can be free of autotools is text replacement in gbversion.h  and gui/setup.iss

tsteven4:
another issue is parallel valgrind(vtesto) for CI.

GPSBabelDeveloper (approved):
Looks good.
Maybe it's time to get that recipe for building doc out of the build system and into a versioned script in tools/ or such.
Are you sure about that oppo link? Perhaps the site is up but the page is just down right now. (There's a reason we left sourceforge...)
I see now that one of the two systems I use to make doc didn't have the hyphenation library so the 1.7.0 PDF, for example, isn't hyphenated.  Oh, well. Comparing them, they aren't THAT different and I'm not sure how how large our audience for that (huge) doc really is anyway.
With FOP 2.5 being common, is 0.92beta really still in play?

tsteven4:
The comment block was pure cut and paste.  And I think you were the only one on FOP < 1.0.  I'll see if I can find a more up to date reference for hyphenation.

GPSBabelDeveloper:
Yeah, I remember that used to be really painful. I don't think it is now. I see that 'brew' already installs the hyphenation libs for you.

tsteven4:
also need a rule for internal_styles.cc, xcsv_tokens.gperf

PR #613, opened 2020-07-28
move linux CI from configure to qmake.

tsteven4:
- Add recipe for internal_styles.cc, xcsv_tokens.gperf,  gui, unix-gui, toolinfo to GPSBabel.pro.
- Add support for text substitution in setup.iss and gbversion.h to GPSBabel.pro.
- fix qmake warnings from "qmake -Wall".
- split recipes for documents into scripts.
- shellcheck fixes for fixdoc, mkcapabilities.
- parallel build for coverage.
- add INSTALL file with build instructions.
- sort document targets into ones for general consumption (gpsbabel.html, gpsbabel.pdf) and one to support www.gpsbabel.org (now gpsbabel.org instead of index.html).


tsteven4:
We still need

1. recipe for internal_styles.cc
2. recipe for xcsv_tokens.gperf
3. text substitution in gbversion.h,  gui/setup.iss

There are lots of abandoned targets in Makefile.in.  If any others are important please mention them.

tsteven4:
internal_styles, xcsv_tokens, substitutions in gbversion.h and setup.iss are now handled in GPSBabel.pro. I am not aware of any necessary targets that haven't been added to GPSBabel.pro, although many targets from Makefile.in aren't available with GPSBabel.pro.

A new file, INSTALL, is added describing the build process.  Chapter 1, building from source needs revision.   Although there is some duplication I think packagers are more likely to see INSTALL than read the document.  Perhaps we can use the INSTALL text in Chapter 1.

I have rearranged the documentation targets a little to try to isolate support of the www.gpsbabel.org from the non-chunked html documentation and the pdf documentation.  The non-chunked html documentation and pdf documentation is built by some distributions.

tsteven4:
There is a problem with EXTRA_TARGETS.  All the commands are passed to a shell with the -c parameter at once.  The series of commands blissfully continues despite the exit status of any previous command.  The only status seen by make is that of the last command.  This is true for ; lists and && lists.

tsteven4:
I've handed off the doc recipes to scripts to solve the error exit issue with QMAKE_EXTRA_TARGETS.

PR #618, opened 2020-08-06
fix endianness/packing issues with energympro, globalsat_sport.

tsteven4:
this was motivated by test failures seen at
https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=s390x&ver=1.7.0%2Bds-3&stamp=1594393020&raw=0

### DIFF
--- a/energympro.cc
+++ b/energympro.cc
@@ -33,7 +33,7 @@
 
 #include "defs.h"
 #include "energympro.h"
-#include "gbfile.h"             // for gbfgetc, gbfseek, gbfclose, gbfopen, gbfread, gbfgetuint32, gbfcopyfrom, gbfgetuint16, gbfile, gbsize_t
+#include "gbfile.h"             // for gbfgetc, gbfgetuint16, gbfgetuint32, gbfseek, gbfgetint16, gbfclose, gbfopen, gbfcopyfrom, gbfile, gbsize_t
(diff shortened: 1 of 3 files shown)


tsteven4:
@jspricke This was motivated by the s390x debian build log cited above.  If you want to fix your regression failure you could use this as a patch.  You seem to be ignoring the error returned by "dh_auto_test: error: make -j2 check VERBOSE=1 returned exit code 2"

jspricke:
Thanks for working on it, I've pushed the patch to Debian. Btw. test are ignored on i386, still:
``
Binary files ./reference/enigma-gpsb.ert and /tmp/gpsbabel.2475/enigma.ert differ
``
https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=i386&ver=1.7.0%2Bds-4&stamp=1596794554&raw=0
Would be great to get that fixed as well.

tsteven4:
@jspricke Thanks for picking up the patch.  Thanks for reporting the i386 regression failure, see #620 for the fix.

jspricke:
Works for Debian: https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=i386&ver=1.7.0%2Bds-5&stamp=1596813332&raw=0

Thanks for fixing it!

PR #620, opened 2020-08-07
fix i386 regression error with engima.

tsteven4:
this was motivated by the test failure seen at
https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=i386&ver=1.7.0%2Bds-4&stamp=1596794554&raw=0

### DIFF
--- a/enigma.cc
+++ b/enigma.cc
@@ -21,9 +21,16 @@
 
  */
 
+#include <cmath>           // for fabs, lround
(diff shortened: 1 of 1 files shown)


jspricke:
Works for Debian: https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=i386&ver=1.7.0%2Bds-5&stamp=1596813332&raw=0

Thanks for fixing it!
(resent to correct PR :D )

PR #622, opened 2020-08-09
compile in gpsbabel translations and map on non mac unix distribution…

tsteven4:
…s by default.

This eliminates the need to install gpsbabelfe_*.qm, gpsbabel_*.qm and
gmapbase.html on the system.
Note that the Qt supplied translations must still be installed on the system.
The Qt translations may be in a package like qttranslations5-l10n or
qt5-qttranslations.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -57,6 +57,14 @@ unix-gui: Build the graphical user interface and collect the components for
   On macOS an app bundle will be created at gui/GPSBabelFE.app and an apple disk
   image will be created at gui/GPSBabelFE.dmg.
 
+Dependencies:
+On non-macOS unix builds by default we now compile in the gpsbabel generated
+translation files, i.e. gpsbabelfe_*.qm, gpsbabel_*.qm, as well as
+gmapbase.html.  When compiled in these files do not need to be distributed.
+These are used by the GUI.  Additional translation files from Qt will also be
+used if they are found.  They may be in a package such as qttranslations5-l10n
+or qt5-qttranslations.
+
 Windows builds:
   Two build methods are supported with MSVC tools.
   1. "qmake -tp vc" will create a visual studio project that can be built with msbuild.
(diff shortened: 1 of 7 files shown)


tsteven4:
@cjmayo @jspricke 
This doesn't add an install target, and it doesn't regenerate the .qm files, but by default on unix (except mac), it compiles the gpsbabel specific translation files and gmapbase.html into the executable alleviating the need to install them.  This should make your installation task much easier, and it should prevent missing translations (except for the Qt supplied ones, they require you add a package dependency).

This PR incorporates ideas from #394, #391 , #386

I am still testing this, but I would appreciate your opinions on the suitability of this concept for the distributions you are associated with.

cjmayo:
Works for me. gpsbabelfe is 1.1 MB or about half the size of gpsbabel. LANG="de" has the translations. map shows.

To install for Gentoo I have to add to the ebuild:

```sh
		dobin gui/objects/gpsbabelfe
		domenu gui/gpsbabel.desktop
		newicon gui/images/appicon.png gpsbabel.png
```
Which is only two more lines than `emake install` would be, so no big deal.

tsteven4:
@cjmayo I think you also need to add a runtime dependency (DEPEND dev-qt/qttranslations:5)  I didn't embed the Qt supplied translations because the system can provide them and they are large compared to the gpsbabel ones.  If they aren't found then things like the Close button on the main window won't be translated, i.e. it will say Close instead of schließen.  I am concerned that dependencies like this aren't obvious, you just get degraded functionality if they are missing.  How can I communicate this need to packagers?

cjmayo:
I think it's something they will already be aware of for Qt packages. For Gentoo as far as I can see it is even left up to the user to install, there are only 3 packages that depend directly on dev-qt/qttranslations and one of those is qt-creator. I just tried on Ubuntu and if I install gpsbabel-gui it pulls in qttranslations5-l10n.

PR #624, opened 2020-08-11
add macos qt 5.15.0 build with artifactory image.

tsteven4:


tsteven4:
The trick here to get the travis cache loaded for foreign PRs is
1) create a branch on gpsbabel/gpsbabel.git with the changes to .travis.yml
2) make a PR from that branch
3) merge that PR.
Step 3 will create the cache files for the default branch, which will be used by PRs from forked repos as a fallback when the cache for that PR isn't found.

This is because foregin PRs don't have access to encrypted variables, so they can't fetch the Qt image from artifactory.

PR #628, opened 2020-08-13
Fix broken qmake-build of macos/libusb:

umgfoin:
Recent switch to qmake is causing build-failures in libusb on macos:
Header 'threads_posix.h' needs to be consumed prior other dependencies.

Signed-off-by: umgfoin

### DIFF
--- a/libusb.pri
+++ b/libusb.pri
@@ -32,6 +32,14 @@ macx|linux|openbsd {
           # TODO: It would be better to create an archive and link to it
           #       to separate library build requirements from gpsbabel requirements.
           DEFINES += LIBUSB_H_INCLUDE=$$shell_quote(\"mac/libusb/libusb.h\")
+          DEFINES += DEFAULT_VISIBILITY=/**/
+          DEFINES += THREADS_POSIX
+          DEFINES += HAVE_POLL_H
+          DEFINES += ENABLE_LOGGING
+          DEFINES += HAVE_SYS_TIME_H
+          DEFINES += OS_DARWIN
+          DEFINES += POLL_NFDS_TYPE=nfds_t
+          DEFINES += _GNU_SOURCE
           LIBS += -lobjc -framework IOKit -framework CoreFoundation
           INCLUDEPATH += mac/libusb \
                          mac/libusb/Xcode
@@ -44,14 +52,14 @@ macx|linux|openbsd {
                      mac/libusb/os/darwin_usb.c \
                      mac/libusb/os/poll_posix.c \
                      mac/libusb/os/threads_posix.c
-          HEADERS += mac/libusb/os/poll_posix.h \
-                     mac/libusb/os/threads_posix.h \
-                     mac/libusb/hotplug.h \
+          HEADERS += mac/libusb/hotplug.h \
                      mac/libusb/libusb.h \
                      mac/libusb/libusbi.h \
                      mac/libusb/version.h \
                      mac/libusb/version_nano.h \
-                     mac/libusb/os/darwin_usb.h
+                     mac/libusb/os/darwin_usb.h \
+                     mac/libusb/os/poll_posix.h \
+                     mac/libusb/os/threads_posix.h
         } else:equals(WITH_LIBUSB, custom) {
           message("libusb-1.0 is enabled but but must be manually configured")
           message("  e.g. qmake WITH_LIBUSB=custom LIBS+=... INCLUDEPATH+=...")
--- a/libusb.pri
+++ b/libusb.pri
@@ -44,14 +44,14 @@ macx|linux|openbsd {
                      mac/libusb/os/darwin_usb.c \
                      mac/libusb/os/poll_posix.c \
                      mac/libusb/os/threads_posix.c
-          HEADERS += mac/libusb/hotplug.h \
+          HEADERS += mac/libusb/os/poll_posix.h \
+                     mac/libusb/os/threads_posix.h \
+                     mac/libusb/hotplug.h \
                      mac/libusb/libusb.h \
                      mac/libusb/libusbi.h \
                      mac/libusb/version.h \
                      mac/libusb/version_nano.h \
-                     mac/libusb/os/darwin_usb.h \
-                     mac/libusb/os/poll_posix.h \
-                     mac/libusb/os/threads_posix.h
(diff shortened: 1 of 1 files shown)


tsteven4:
It works fine for us on travis CI.  Any idea of what's different for you?

umgfoin:
You're right, I'm seeing different (fail)behaviour on two macos systems, too: They differ in Qt-installation-flavor (app vs. brew).
The root cause seems to be that `mac/libusb/Xcode/config.h` isn't read, if another `config.h` in the search path can be found. Among others, this ends up in `THREADS_POSIX` being undefined while compiling `libusbi.h`.
I think, we need to find a way ensuring explicit use of `Xcode/config.h` or manually declaring the config.h-defines in `libusb.pri `
My previous PR fixed the problem just by coincidence and just for a specific system.

`
/Library/Developer/CommandLineTools/usr/bin/clang -c -pipe -O2  -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -mmacosx-version-min=10.13 -Wall -Wextra -fPIC -DNDEBUG -DHAVE_UNISTD_H -DHAVE_STDARG_H -DSHAPELIB_ENABLED -DHAVE_LIBUSB_1_0 -DLIBUSB_H_INCLUDE='"mac/libusb/libusb.h"' -DMAXIMAL_ENABLED -DFILTERS_ENABLED -DCSVFMTS_ENABLED -DQT_NO_DEBUG -DQT_CORE_LIB -I. -Ijeeps -Izlib -Imac/libusb -Imac/libusb/Xcode -I/usr/local/Cellar/qt/5.15.0/lib/QtCore.framework/Headers -I. -I/usr/local/Cellar/qt/5.15.0/mkspecs/macx-clang -F/usr/local/Cellar/qt/5.15.0/lib -o hotplug.o mac/libusb/hotplug.c
In file included from mac/libusb/core.c:44:
mac/libusb/libusbi.h:302:2: error: unknown type name 'usbi_mutex_t'
        usbi_mutex_t usb_devs_lock;
        ^
`

tsteven4:
The master branch seems to work without change for me using brew supplied Qt.  Where is this other config.h coming from?  Perhaps it is leftover from running gpsbabel's configure before it was removed?  run "git clean -x -d" on your gpsbabel clone?

/usr/local/Cellar/qt/5.15.0/bin/qmake GPSBabel.pro

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -c -pipe -O2  -arch x86_64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk -mmacosx-version-min=10.13 -Wall -Wextra -fPIC -DNDEBUG -DHAVE_UNISTD_H -DHAVE_STDARG_H -DSHAPELIB_ENABLED -DHAVE_LIBUSB_1_0 -DLIBUSB_H_INCLUDE='"mac/libusb/libusb.h"' -DMAXIMAL_ENABLED -DFILTERS_ENABLED -DCSVFMTS_ENABLED -DQT_NO_DEBUG -DQT_CORE_LIB -I. -Ijeeps -Izlib -Imac/libusb -Imac/libusb/Xcode -I/usr/local/Cellar/qt/5.15.0/lib/QtCore.framework/Headers -I. -I/usr/local/Cellar/qt/5.15.0/mkspecs/macx-clang -F/usr/local/Cellar/qt/5.15.0/lib -o hotplug.o mac/libusb/hotplug.c

umgfoin:
> Perhaps it is leftover from running gpsbabel's configure before it was removed? run "git clean -x -d" on your gpsbabel clone?

Yeah - leftovers... thanks for the hint! `git clean -x -d -f` fixed it.

tsteven4:
By the way, this sort of thing is why the pri files say
```
          # TODO: It would be better to create an archive and link to it
          #       to separate library build requirements from gpsbabel requirements.
```

GPSBabelDeveloper:
I come here not to praise libusb on MacOS but to destroy it.

I should have caught this. I've been bugged by this for a while, but I was pretty unique in flipping between configure and qmake. 

There is really a pretty small percentage of libusb that we're actually using. We don't use hotplug or the async stuff, which is where all the thread and mutex stuff comes in and we don't do much beyond bulk/interrupt reads/writes. I'd rather write an equivalent of jeeps/gpslibusb.cc or usbwin.cc - which provide the same services to the bottom of GPSBabel - that spoke IOKit and forget about libusb on MacOS. 

Examples and doc are more plentiful now than they were in 2004 and the needed code is about halfway between https://tewarid.github.io/2012/04/27/access-usb-device-on-mac-os-x-using-i-o-kit.html and the files above, with some inspiration from the innards of the darwin libusb implementation. From the view in 2004, we were going to see multiple USB protocols on different devices, but the world quickly peeled into driverless (yay!) storage devices or half-baked serial devices using on-board USB/Serial adapters, like almost all of our loggers.

As is always the case with the libusb stuff, it's the retest cost that's a killer. That brings us back to our annual discussion.  Garmin's last products that used this stuff were in 2006 and 7. Everything since then in this class has been mass storage with GPX files. Even Garmin never supported these products on MacOS. Both Garmin and Apple had to fix bugs in their respective implementations that we unearthed in this stuff. There was a time the 60Csx was popular with the search and rescue type teams, but they've been unable to get hardware and parts for over ten years, so these have likely rolled off service by then. Given market share realities, it's entirely possible our user base of 'garmin/usb' rounds to zero for all OSes and would be about 15% of zero for MacOS.

Of course, the serial units are older than that. Maybe it's time to let Garmin protocol go completely...

I've left all this for a long time, but if libusb is going to be a squeaky wheel, I'd consider just whacking it.

A) Apply a little more grease and keep it going. Requiring a 'make clean' when we replace build systems isn't onerous.
B) Make a jeeps/gpsusbmac.cc that's inspired by above.
C) Fire Garmin USB protocol.
D) Fire all remaining Garmin USB AND serial devices. The amount of support we offered on devices making their SECOND week 1024 rollover or that have their clocks stuck in the year 2300 was ridiculous.

Big Sur may force this issue somewhat.

We just don't have a good way of tracking when usage falls to zero and knowing when we can quit.

PR #629, opened 2020-08-14
add some explicit docbook ids

tsteven4:
this makes the generated html more reproducible as random ids
aren't generated.


GPSBabelDeveloper:
Yes, please.

Good starter project (for someone else): clean up the docbook warnings and add presubmit tests for the common cases to keep it quiet.

PR #636, opened 2020-08-25
remove unintentional collision with errno.

tsteven4:
this caused compiler failures with gnu++1z.


GPSBabelDeveloper (approved):
Good call. That was never right, even in C89.

PR #638, opened 2020-08-27
fix hiketech datetime bug, prepare for toTime_t override

tsteven4:
the mystery previously noted in the code was due to the use of the %I format
specifier which prints out the hour using a 12 hour clock.

note that it is easy to see the times in the reference files were wrong.  the
times in the first reference file, expertgpx.gpx, are in UTC.  This is always
true for gpx, and is indicated as well by the Z.  the times in the hiketech file
appear in a utc element, so they must be in UTC as well.

this eliminates the usage of QDateTime::toTime_t which is obsolete.

eliminate remaining uses of obsolete QDateTime::toTime_t

This sounds better than it is, we added uses of gpsbabel::DateTime::toTime_t.
But we control that and can add our own toTime_t.

### DIFF
--- a/hiketech.cc
+++ b/hiketech.cc
@@ -128,17 +128,7 @@ hiketech_trk_tlr(const route_head*)
 static QString
 hiketech_format_time(const QDateTime& t)
 {
-  // FIXME: Find out why these two blocks of code aren't equivalent.
-  // it produces times that are 12 hours too late.  Double TZ bump?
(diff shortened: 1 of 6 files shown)


GPSBabelDeveloper:
Seems right. 

Aside, the maker of Hiketech hasn't existed for a long time. We should put a DNR on that module and not offer any extraordinary life saving measures.

PR #639, opened 2020-08-27
Fix potential segmentation faults with exif.

tsteven4:
1. A pointer exif_app_ to an ExifApp on QList<ExifApp> exif_apps was saved.
After this the list was modified.  This can lead to the saved pointer becoming
invalid.
2. The ExifApp structure had a dtor, but no other special functions, i.e.  it
violated the rule of 3 and the rule of 5.  Operations on QList<ExifApp> may
cause an ExifApp on the list to be copied or destroyed.  If an ExifApp is
destroyed, then the ExifApp dtor would close the files, even though there could
be a copy of the ExifApp expecting the files still to be open.

This scenerio occured with Qt6, causing segmentation faults in exif.test.

### DIFF
--- a/exif.cc
+++ b/exif.cc
@@ -177,20 +177,10 @@ struct ExifApp {
   gbfile* fcache{nullptr};
   gbfile* fexif{nullptr};
   QList<ExifIfd> ifds;
-
-  ~ExifApp()
-  {
-    if (fcache) {
-      gbfclose(fcache);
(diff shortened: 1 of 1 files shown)


GPSBabelDeveloper (commented):
Good find.

This is the stuff that makes Python programmers of formerly self-respecting C++ devs...

PR #640, opened 2020-08-29
backward compatible changes for Qt6.

tsteven4:
This is mostly QRegExp -> QRegularExpression.
Some QRegExp usage was was replaced by QString::trimmed.
There is also a QNetworkRequest redirect attribute change that is backward
compatible to Qt5.9.

### DIFF
--- a/gui/aboutdlg.cc
+++ b/gui/aboutdlg.cc
@@ -21,8 +21,13 @@
 //
 
 #include "aboutdlg.h"
-#include "appname.h"
-#include "upgrade.h"
+#include <QtCore/QRegularExpression>  // for QRegularExpression
+#include <QtGui/QTextCursor>          // for QTextCursor
+#include <QtGui/QTextDocument>        // for QTextDocument
+#include <QtWidgets/QTextEdit>        // for QTextEdit
+#include "appname.h"                  // for appName
+#include "upgrade.h"                  // for UpgradeCheck
+
 
 AboutDlg::AboutDlg(QWidget* parent, const QString& ver1,
                    const QString& ver2, const QString& installationId): QDialog(parent)
@@ -31,13 +36,13 @@ AboutDlg::AboutDlg(QWidget* parent, const QString& ver1,
   QTextDocument* doc = ui_.textEdit->document();
   ui_.textEdit->setReadOnly(true);
   QString tt = doc->toHtml();
-  tt.replace(QRegExp("\\$appname\\$"),  QString(appName));
-  tt.replace(QRegExp("\\$babelversion\\$"),  ver1);
-  tt.replace(QRegExp("\\$babelfeversion\\$"),  ver2);
-  tt.replace(QRegExp("\\$installationId\\$"),  installationId);
+  tt.replace(QRegularExpression(R"(\$appname\$)"), QString(appName));
+  tt.replace(QRegularExpression(R"(\$babelversion\$)"), ver1);
+  tt.replace(QRegularExpression(R"(\$babelfeversion\$)"), ver2);
+  tt.replace(QRegularExpression(R"(\$installationId\$)"), installationId);
 
   // Not localized as it should never be seen.
-  tt.replace(QRegExp("\\$upgradetestmode\\$"),
+  tt.replace(QRegularExpression(R"(\$upgradetestmode\$)"),
              UpgradeCheck::isTestMode() ? "**Upgrade test mode**" : "");
 
   doc->setHtml(tt);
--- a/gui/aboutdlg.cc
+++ b/gui/aboutdlg.cc
@@ -21,7 +21,6 @@
 //
 
 #include "aboutdlg.h"
-#include <QtCore/QRegularExpression>  // for QRegularExpression
 #include <QtGui/QTextCursor>          // for QTextCursor
 #include <QtGui/QTextDocument>        // for QTextDocument
 #include <QtWidgets/QTextEdit>        // for QTextEdit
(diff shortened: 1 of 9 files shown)


tsteven4:
@gpsbabel Note that https://doc.qt.io/qt-5/qnetworkrequest.html#RedirectPolicy-enum documents the change to QNetworkRequest as equivalent.  However, this has been an issue recently so maybe a manual test is in order.

> 

QNetworkRequest::NoLessSafeRedirectPolicy | 1 | Only "http"->"http", "http" -> "https" or "https" -> "https" redirects are allowed. Equivalent to setting the old FollowRedirectsAttribute to true
-- | -- | --

gui/aboutdlg.cc
```diff
@@ -31,13 +36,13 @@ AboutDlg::AboutDlg(QWidget* parent, const QString& ver1,
   QTextDocument* doc = ui_.textEdit->document();
   ui_.textEdit->setReadOnly(true);
   QString tt = doc->toHtml();
-  tt.replace(QRegExp("\\$appname\\$"),  QString(appName));
-  tt.replace(QRegExp("\\$babelversion\\$"),  ver1);
-  tt.replace(QRegExp("\\$babelfeversion\\$"),  ver2);
-  tt.replace(QRegExp("\\$installationId\\$"),  installationId);
+  tt.replace(QRegularExpression(R"(\$appname\$)"), QString(appName));
+  tt.replace(QRegularExpression(R"(\$babelversion\$)"), ver1);
```
GPSBabelDeveloper: I know this is mechanical, but this probably doesn't need to e a regex at all; it's a straightforward replace(), right?

gui/formatload.cc
```diff
@@ -67,13 +67,11 @@ bool FormatLoad::processFormat(Format& format)
...
-  QString htmlPage = lines_[currentLine_++];
-  htmlPage.replace(QRegExp("^[\\s]*"), "");
-  htmlPage.replace(QRegExp("[\\s]$"), "");
+  QString htmlPage = lines_[currentLine_++].trimmed();
 
-  QRegExp regex("^option");
+  QRegularExpression regex("^option");
   QList <FormatOption> optionList;
-  while (currentLine_ <lines_.size() && regex.indexIn(lines_[currentLine_]) == 0) {
+  while ((currentLine_ < lines_.size()) && regex.match(lines_[currentLine_]).hasMatch()) {
```
GPSBabelDeveloper: So many of these regex uses are overkill. Isn't this currentline_.startswith("option") ? There are a few occurrences like this.

gui/help.cc
```diff
@@ -20,22 +20,22 @@
...
+
+#include "format.h"                   // for Format
 
 //------------------------------------------------------------------------
 void ShowHelp(const QString& urlIn)
 
 {
   QString url = urlIn;
-  if (!url.contains(QRegExp("^http://"))) {
+  if (!url.contains(QRegularExpression(R"(^https?://)"))) {
```
GPSBabelDeveloper, line 37: likewise

gui/mainwindow.cc
```diff
@@ -510,8 +507,8 @@ void MainWindow::outputFileNameEdited()
 QString MainWindow::filterForFormat(int idx)
 {
   QString str = formatList_[idx].getDescription();
-  str.replace(QRegExp("\\("), "[");
-  str.replace(QRegExp("\\)"), "]");
+  str.replace('(', '[');
```
GPSBabelDeveloper, line 510: This would be a good starter project for someone new to the project. Add a flag for -^4 or -^json or something that emitted something better defined than TSV that could be read by code like this.

gui/upgrade.h
```diff
@@ -19,13 +19,17 @@
...
+#include <QtCore/QObject>                   // for QObject
+#include <QtCore/QString>                   // for QString
+#include <QtCore/QUrl>                      // for QUrl
+#include <QtNetwork/QNetworkAccessManager>  // for QNetworkAccessManager
+#include <QtNetwork/QNetworkReply>          // for QNetworkReply
+#include <QtWidgets/QWidget>                // for QWidget
+#include "babeldata.h"                      // for BabelData
+#include "format.h"                         // for Format
 
-class QNetworkAccessManager;
```
GPSBabelDeveloper, line 27: Forward decls when we need only the pointer and not any member access means we can skip the #includes. We probably don't do enough of this in the porject. Is this just your script getting overambitious?

GPSBabelDeveloper (approved):
Looks good. Easier to read in most cases, too. Optional observations, nits and questions, but approved.

gui/aboutdlg.cc
tsteven4: geez.   thanks for the second pair of eyes.

gui/upgrade.h
tsteven4: I was swayed by https://google.github.io/styleguide/cppguide.html#Forward_Declarations
I am running include-what-you-use and specifically telling it to avoid forward declarations.

gui/help.cc
tsteven4: this one actually fixes a bug,  http or https, so it is a regex.  There may be a QUrl method we could use.

gui/mainwindow.cc
tsteven4: Agreed.  Where is our army of enthusiastic eager young volunteers?

gui/formatload.cc
tsteven4: yes.

gui/upgrade.h
GPSBabelDeveloper: That's compelling.
Ironic because my mentally cached version of this documented went the other way. Compile times were dominating development and Qt's headers were some of the worst. We spent a ridiculous amount of effort tracking down scoping issues that were exposed when the project essentially did blah.cc: #include *.cc just so the tokenizer and parser didn't have to chomp on 0000's of copies of <QString> for every build. Qt headers were always at the top of our build profiles.

That document has changed a few times over the years as tooling and techniques have evolved.

We traded developer pain for build times. Looks like the pendulum has gone back.

I agree, though, that we should probably follow a formal guideline on this kind of thing instead of makign up our own.

PR #643, opened 2020-09-11
Update tpo.cc

SRE-babel:
Some .tpo file tracks were/are parsed such that wild jumps occur in lat/lon, creating invalid coords which were not caught by output routines. Some logic flaws and int size errors have been fixed, but some tracks still contain invalid points. Tracks not corrected by this patch now result in warning messages to STDOUT. Entirely valid tracks will also generate warnings if lat or lon changes by a full degree between adjacent track points (should never happen). Non-fatal warning allows manual removal of the few bad track points  in an output format like GPX. Damage is apparently limited to half a dozen incremental points before the next full point, but that can't be guaranteed.


tpo.cc
```diff
@@ -518,38 +514,43 @@ static void tpo_process_tracks()
...
     Q_UNUSED(tmp)
-    // printf("Skipping unknown byte 0x%x after color\n",tmp);
+    if(DEBUG > 2) { printf("Skipping unknown byte 0x%x after color (? always zero ?)\n",tmp); }
 
-    // byte for name length, then name
+    // byte for track style name length, then name itself
     tmp = gbfgetc(tpo_file_in);
     // wrong byte order?? tmp = tpo_read_int(); // 16 bit value
-    // printf("Track %d has %d-byte (0x%x) name\n", ii, tmp, tmp);
+    if(DEBUG > 1) { printf("Track style %d has %d-byte (0x%x) name\n", ii, tmp, tmp); }
```
tsteven4, line 536: Arg 1 is unsigned, so 
"Track style %u has %d-byte (0x%x) name\n"

tpo.cc
```diff
@@ -565,39 +566,35 @@ static void tpo_process_tracks()
     // clumsy way to skip two undefined bytes
     for (unsigned xx = 0; xx < 2; xx++) {
       tmp = gbfgetc(tpo_file_in);
-      // printf("Skipping final byte 0x%x\n", tmp);
```
tsteven4, line 567: Can we delete all these //printfs?

tpo.cc
```diff
@@ -51,6 +46,10 @@
  Older (pre-3.0) format does not have the above string.  Contains
  only tracks.  Waypoints are saved in a separate .TPG file.
 
+ Track parsing has been problematic and may still not be right!
+ See further notes and clues in tpo_process_tracks()
+ (can #define Tracks2012 to revert to pre-2020 code in tpo_process_tracks)
```
tsteven4, line 51: can we delete all the Tracks2012 code?

tsteven4 (changes requested):
Thanks for taking this on, sorry the review has been stalled.

Please add a test to demonstrate the problem you were addressing has been solved.

This needs reformatting to conform to our style, "astyle --options=astylerc tpo.cc".  If you don't have astyle I can push the style changes to your branch.

The elimination of the commented printfs and Track2012 clauses as suggested will simplify maintenance as we don't have to maintain unused code.

There are additional opportunities for improvement that we could address later
1. use QScopedArrayPointer instead of malloc for buf.
2. eliminate shadowing of route_count() and track_count()
3. could use "if constexpr(DEBUG ..." statements if we merge (c++17)

tpo.cc
tsteven4: This was fixed in #735

Issue #646, opened 2020-10-24
Question by filtering


velocat:
Tell me, I can't discard part of the data for the output file if it's not gpx.
For example, I want to filter data from a KML file to the same format
, but I get absolutely all the data in the output.
-w -t -r - not effect
nuketypes - not effect
```
gpsbabel -D9 -w -i gpx -f 'file1.gpx' -x nuketypes,tracks,routes -i kml -f 'file2.kml' -x nuketypes,tracks,routes -o kml,unit=m,points=0 -F /tmp/temp.kml
```

I see as an option only save the file as gpx (filtered), and the second pass to save it in the desired format. But this is unnecessary, I think. :(

Maybe I'm doing something wrong?

tsteven4:
Perhaps you meant to write .file2.kml?  If so you need to use -F instead 
of -f.

To monitor waypoints, routes, tracks you can add -x validate,debug.

The following shows bounds-test.gpx has 11 waypoints and three routes 
before the nuketypes filter.   After the nuketypes filter only 11 
waypoints remain.

> ./gpsbabel -i gpx -f reference/bounds-test.gpx -x validate,debug -x 
> nuketypes,routes,tracks -x validate,debug -o gpx -F out.gpx
>
> Processing waypts
> point ct: 11, waypt_count: 11
>
> Processing routes
> route 0 ct: 60, waypt_count: 60
> route 0 ct: 35, waypt_count: 35
> route 0 ct: 13, waypt_count: 13
> route head ct: 3, route_count: 3
> total route point ct: 108, route_waypt_count: 108
>
> Processing tracks
> track head ct: 0, track_count: 0
> total track point ct: 0, track_waypt_count: 0
>
> Processing waypts
> point ct: 11, waypt_count: 11
>
> Processing routes
> route head ct: 0, route_count: 0
> total route point ct: 0, route_waypt_count: 0
>
> Processing tracks
> track head ct: 0, track_count: 0
> total track point ct: 0, track_waypt_count: 0

velocat:
Well, I wrote that there are no problems with filtering the gpx file.
But how can I filter out another format, such as xml?

my example shows 2 gpx and kml files (-f), from which I want to get the third one (-F), but I want to discard tracks and routes in it and leave only waypoints

tsteven4:
The filters don't know or care where the waypoints/routes/tracks came from.

Try this:

[...]

PR #647, opened 2020-10-27
ozi: don't deinit a write stream if we don't have one.

robertlipe:
We can get into close() with a null stream_ TextStream. I think it has to do with ozi_openfile() reopening a file with a new name to match an extension of a certain type so it's renaming a file that hasn't been written. See issue #646


ozi.cc
```diff
@@ -172,6 +172,9 @@ ozi_open_io(const QString& fname, QIODevice::OpenModeFlag mode)
 static void
 ozi_close_io()
 {
+  if (!stream) {
```
tsteven4, line 175: This makes the check around the call to ozi_close_io in ozi_openfile redundant (line 263).

tsteven4:
We should move this along, with or without resolving the redundancy I pointed out above.  It's a real bug and a good fix.

tsteven4:
> 
> 
> We can get into close() with a null stream_ TextStream. I think it has to do with ozi_openfile() reopening a file with a new name to match an extension of a certain type so it's renaming a file that hasn't been written. See issue #646

The only case I have found that generates the fault is when waypt_count, route_count and track_count are all zero.  I don't think there is an issue with reopening a file with a new name as speculated above.

I have enhancements to the ozi test that covers both of the above cases.  I can push those to this branch or merge this branch and put them in a separate PR.   @robertlipe please tell me how you would like to proceed.  thanks

tsteven4:
In the absence of further communication I am merging this.  The test cases demonstrate resolution of the reported segmentation fault.

robertlipe:
Agreed. Thank you. That sat almost at the top of my TODO list for too long.

Sidebar for anyone else with too many computers and too many source trees in your life: include the computer and the directory name when creating git branches.

PR #651, opened 2020-11-11
manually install Qt 5.12.10 for Appveyor msvc 2017 builds

tsteven4:
Appveyor has not updated Qt 5.12 in the MSVC 2017 build environment for some time.  A ticket https://github.com/appveyor/ci/issues/3306 from February 2020 says these environments will be updated, but it hasn't happened yet.

This uses Appveyor install scripts to install the Qt version of our choice as suggested in the above ticket.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -41,13 +41,13 @@ init:
       }
   - ps: |
       if ($env:qt -eq "5.12\msvc2017") {
-          $env:path = "C:\Qt\5.12.9\msvc2017\bin"; $env:path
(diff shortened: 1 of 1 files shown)


robertlipe (approved):
Looks good. Is there an easy way to reduce the number of occurrences of "5.12" and "5.12.10" in this? It doesn't look like a big maintenance burden and I don't remember batch files well enough to remember the rules of variables, but if it's easy, please refactor to a variable. If it's not, it's still a good win and we should take it. Thanx!

tsteven4:
This installation method is of historical interest, but we are moving to github actions.

PR #656, opened 2020-12-15
backward compatible changes for Qt6

tsteven4:
- Avoid implicit conversion of int to QChar, in Qt6 the constructor QChar(int code) is explicit by default.
- use focal for coverage generation
- use binary codacy coverage reporter.  the java version was failing on travis with "java.io.IOException: Error writing to server"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,7 @@ jobs:
           - $HOME/Cache
         timeout: 600
     - os: linux
-      dist: bionic
+      dist: focal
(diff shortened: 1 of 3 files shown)


robertlipe:
I know all the cool kids are doing it, but this trend of "download and run
a script at build time" sure seems like a security problem just waiting for
a cool code name and web  write-up. It just realy hampers

No need for us to change it here. I'm just being a grouch...

Issue #659, opened 2020-12-28
Consider 64bit executable for windows


tsteven4:
It is possible to fail memory allocation using the 32 bit windows executable with a large number of waypoints.

An example is "-i gpx -f big.gpx -o gpx -F xxx.gpx", where big.gpx contains 105,596 waypoints or thereabouts.
1. In this case std::bad_alloc is thrown and uncaught.  The program dies without an error message.  It would be easy to catch this exception and at least try to print a message.
2. With a 64 bit windows executable the correct output file is produced.

Using the MSVC 64 bit compiler generates many more warnings compared to the MSVC 32 bit compiler.  ~~In general these have to do with the word size mismatch of the long type between the LLP64 memory model used by MSVC x86_amd64 and the LP64 memory model that was used to develop the majority of the code.~~  Note that CI builds with both MSVC 32 bit and 64 bit compilers, but historically we have always distributed only a 32 bit windows executable.

tsteven4:
Note that the QStarz BT-Q1000XT can log 400,000 points.

tsteven4:
Almost all the errors are C4267, The compiler detected a conversion from size_t to a smaller type.

Note that with  gcc and clang we do not enable Wconversion, Warn for implicit conversions that may alter a value.   If we enable this warning many of the warnings flagged by C4267 would be detected by gcc/clang.  Wconversion is not enabled by Wall.

I recommend we disable C4267 which brings the MSVC compiler warnings closer to gcc/clang.  The C4100 warning can easily be fixed as we have done in other places, but I would  like to avoid touching shapelib.

We currently appear to run Inno Setup in 32 bit mode and the resulting executables are placed in C:\Program Files (\x86), even for our 64 bit build.

MSVC 32 bit build, with C4267 disabled (2 warnings):
Severity	Code	Description	Project	File	Line	Suppression State
Warning	C4100	'hSHP': unreferenced formal parameter	gpsbabel	C:\Users\strabert\source\repos\gpsbabel\shapelib\shpopen.c	3030	
Warning	C4477	'printf' : format string '%ld' requires an argument of type 'long', but variadic argument 2 has type 'time_t'	gpsbabel	C:\Users\strabert\source\repos\gpsbabel\lowranceusr.cc	500	

MSVC 64 bit build, with  C4267 disabled (2 warnings):
Severity	Code	Description	Project	File	Line	Suppression State
[...]

PR #660, opened 2020-12-29
switch windows release build from 32 bit to 64 bit (only).

tsteven4:
This switches our release build to 64 bits for windows, although 32 bit versions are still build in CI.  This will preclude the use of x86 hardware with our official release.  I have seen estimates of 0.66% market share for 32 bit.

By default the windows install will now go to "Program Files" instead of "Program Files (x86)".

As mentioned in #659, turning off MSVC warning C4267 brings us closer to the warning set we use with gcc/clang (Wall).


robertlipe:
If it really is .68% we should let it go. The recent pieces on LWN.net really brought home how far into disrepair 32-bit platforms have become in general computing, but I thought there were more pre-Core and low-grade Celeron-is Intel systems out there in use. 

Are there issues with in-place update? Will our Settings from 32-bit still be readable by our "new" app?

tsteven4:
The settings will be preserved unless the user uninstalls the 32 bit app before installing the 64 bit app.  Whenever we run the uninstaller we delete all the registry keys (Flags: uninsdeletekey).

However, if the user has the 32 bit app installed, and then installs the 64 bit app, both apps will be present but windows will only know about the 64 bit app in "add remove programs".  The user can manually resolve this by deleting C:\Program Files (x86)\GPSBabel, or can leave the 32 bit app in place.

By "in-place" update, I assume you mean installing the 64 bit app in C:\Program Files (x86)\GPSBabel on a 64 bit version of windows.  This can be done, but I think it is not what MS intended.  We could end up with both 32 and 64 bit pieces in the folder if something isn't replaced by the new install, e.g. the msvc run time installer.  I think chrome used to do this but now conforms to expectations for new installs, https://www.ghacks.net/2020/06/11/google-chrome-is-soon-going-to-be-installed-in-a-different-directory-on-windows/

tsteven4:
The market share numbers I referred to was from https://www.pcbenchmarks.net/os-marketshare.html

tsteven4:
> However, if the user has the 32 bit app installed, and then installs the 64 bit app, both apps will be present but windows will only know about the 64 bit app in "add remove programs". The user can manually resolve this by deleting C:\Program Files (x86)\GPSBabel, or can leave the 32 bit app in place.

We could probably maintain both uninstallers in that situation, and share registry values, if we change the AppId for the 64 bit version.  We might want to change AppVerName for the 64 bit version to "GPSBabel 1.7.0 (64-bit)"  or "GPSBabel 1.7.0 (x64)" so it would be different in the add/remove programs dialog.  It is too late to change the 32 bit version to "GPSBabel 1.7.0 (32-bit).

tsteven4:
> We could probably maintain both uninstallers in that situation, and share registry values, if we change the AppId for the 64 bit version. We might want to change AppVerName for the 64 bit version to "GPSBabel 1.7.0 (64-bit)" or "GPSBabel 1.7.0 (x64)" so it would be different in the add/remove programs dialog. It is too late to change the 32 bit version to "GPSBabel 1.7.0 (32-bit)

But, running either uninstaller will delete all the registry keys.

GPSBabelDeveloper:
This all sounds pretty reasonable to me. Unlike some apps, we just don't
retain THAT much state in our settings, so I'm OK with there being minor
interference there. I'd think that MOST people will just install an update
and not try to manually reclaim the tiny droppings left over. We could
force the issue and manually nuke the 32-bit version (that may still be
referenced by scripts and automation) but I'd rather leave it there. (We'll
have to be prepared for some people running an older version than they
think: they'll install 1.8.0 but their batch file still calls 1.7.0....)


I do wonder if we should go back to streaming writers instead of apparently
trying to hold all the DOM in VM at once, though. Using > 2GB of RAM to
make a 30MB file seems WAY out of proportion to me, but for one affected
user I'm not digging into measuring it. OTOH, there are some defaults in
our KML writer that are overdue to be changed anyway. All that balloon
stuff should become ExtendedData in modern times, for example.

I'm OK with all of this and stopping here for now. Thanx for your diligence.

RJL

tsteven4:
OK.  To summarize the expected behavior is:  

-  It 32 bit gpsbabel is installed it will be left in place when 64 bit gpsbabel is installed, but the registry (and Control Panel Add/Remove Programs) will be connected to the 64 bit install (or more accurately to the last gpsbabel version installed).  The 32 bit install and it's uninstaller will be orphaned.  If the user wants to clean up they can just delete the 32 bit installation with File Explorer.
- The registry keys are shared, so if the user doesn't run uninstall then they will be preserved and shared between 32 bit and 64 bit versions.   Running either uninstaller deletes are the registry keys where we save settings as it always has.

PR #663, opened 2021-01-11
fix some codeql warnings.

tsteven4:


tsteven4:
I suspect the z length modifier didn't work with earlier versions of MSVC 2015 (the documentation says its not supported), but CI shows it works with the final version of MSVC 2015 so I think we are good.  I've about had it with MSVC 2015 anyway, the next time it misbehaves I think I will move the floor to MSVC 2017.

robertlipe:
Looks good to me. Thanx.
Clearly I'm not the only one that forgets about %z even though I know it's a nightmare across systems.

It's been ten years and I've still not moved to I/O streams for debugging and errors, so let's not even start the clock on the new (and pretty awesome) {fmt} that landed as std::format in C++20.

Go for it.

tsteven4:
> Go for it.

I will once I can move #662 forward, which requires resolution of repository permissions, i.e. migrating to an organization.  Having some PR's to merge will be a good test of #662 as some things only happen then.

PR #664, opened 2021-01-14
Remove dead formats: alan, an1, netstumbler, naviguide.

robertlipe:
As discussed in https://github.com/gpsbabel/gpsbabel/issues/659


tsteven4 (commented):
Yay!

For consistency do you want to move naviguide.cc to deprecated instead of deleting it?
For consistency we have been moving deprecated tests to testo.d/deprecated instead of deleting them.
We should also move/delete make-an1sym.pl.
It appears we have been deleting reference and documentation files, although I suspect some reference files associated with deprecated formats/filters are still in their original locations.

tsteven4:
Robert, Do you want some help with this?  I can create another branch, squash merge this branch, restore some deleted files and move them to deprecated while preserving history (for the source and test scripts).

With regards to the doc and reference files we some orphans today.  We could created deprecated directories for them as well, or delete them as you have done, or leave them as orphans.

GPSBabelDeveloper:
Thanks. I'll pick it back up soon.

PR #670, opened 2021-01-26
Update standards to c++17, Qt >= 5.12, MSVC >= 2017

tsteven4:
This required changes to the code to accommodate macOS 10.12 and 10.13, which don't support std::optional::value.  Instead of the value method the dereference operator was used.  Thanks to github actions for discovering this issue.

There are probably opportunities to create odr-used constexpr static data members at namespace scope in include files without referencing them in a .cc file now.  This was previously done with a no longer necessary conditional reference in garmin_fit.cc, height.cc, lowranceusr.cc.

This change precludes compilation on Ubuntu bionic with the Ubuntu distributed Qt (5.9.5).  However, we are still compiling on bionic with a manually installed Qt 5.12.10.  Note that the next Ubuntu LTS release, focal, is now available and being used on some CI jobs.  Ubuntu distributes Qt 5.12.8 with focal.  I think we could drop bionic from CI at this point and request the army of ubuntu developers that flood us with PRs to use focal.


robertlipe (approved):
I agree. The deref vs value () thing is disappointing, but not a major defeat. 

If there was an automated constexpr  finder, I'm shocked that's all it found. We do sizeof struct all over. Oh, well.

Solid cleanup here. Thanx!

PR #671, opened 2021-01-27
trust Qt to strip out characters illegal in xml.

### LINKED ISSUE #637: xmlstreamwriter use of codec
In Qt6 QXmlStreamWriter always writes utf-8.  In our xmlstreamwriter we create a codec to filter out some characters which is no longer supported.  It looks like this was fixed in Qt in https://github.com/qt/qtbase/commit/3b5b8f1d4ab8092e5dd337b7b4e32d85fda2e0b7 likely making our fix unnecessary.  The Qt fix will set hasError.  The fix appeared in 5.11.0, so it isn't in our floor of 5.9.

tsteven4:
this feature was added in Qt 5.11.0.

this resolves #637

### DIFF
--- a/src/core/xmlstreamwriter.cc
+++ b/src/core/xmlstreamwriter.cc
@@ -19,75 +19,28 @@
 
 #include "src/core/xmlstreamwriter.h"
 
-#include <QtCore/QFile>
-#include <QtCore/QString>
-#include <QtCore/QTextCodec>
-#include <QtCore/QXmlStreamWriter>
+#include <QtCore/QString>           // for QString
+#include <QtCore/QXmlStreamWriter>  // for QXmlStreamWriter
+#include <QtCore/QtGlobal>          // for QT_VERSION, QT_VERSION_CHECK
 
 // As this code began in C, we have several hundred places that write
 // c strings.  Add a test that the string contains anything useful
 // before serializing an empty tag.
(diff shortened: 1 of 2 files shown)


robertlipe (approved):
This is a good double hit for us.
1) That code was excessively clever. Our need for that code was gross. For form feeds in description tags no longer croak us? (I think Steve relied on that in on that in TPO.) Good! The original source of this horror was Groundspeak GPX from pocket queries and we helped shame them into fixing that long ago.
2) This gets us out of the need for 1 codec that was problematic under Qt 6.
Nice!

src/core/xmlstreamwriter.cc
```diff
@@ -19,75 +19,28 @@
...
 
-#include <QtCore/QFile>
-#include <QtCore/QString>
-#include <QtCore/QTextCodec>
-#include <QtCore/QXmlStreamWriter>
+#include <QtCore/QString>           // for QString
+#include <QtCore/QXmlStreamWriter>  // for QXmlStreamWriter
+#include <QtCore/QtGlobal>          // for QT_VERSION, QT_VERSION_CHECK
 
 // As this code began in C, we have several hundred places that write
```
robertlipe, line 26: This comment has gotten orphaned from its code. Please consider shoving it closer to the body of writeOptionalTextElement.

PR #672, opened 2021-01-28
update readme.md with current continuous integration provider

tsteven4:


tsteven4:
@robertlipe, are you cool with these direct links to the pre-releases and the downloads on gpsbabel.org.  They used to point to the CI logs, but that is redundant now.

robertlipe:
I won't reject it, but I'm a bit clammy.
1) do we manage version numbers in these automatically? Are they handled correctly by auto update?
2) I'm not a fan of processes that encourage drive by downloads. Sending them to the web site at least gives us the chance to imprint our online help, FAQ, mailing list addresses and (this won't be popular) our donation pages. It short circuits our chance to develop a community member instead of just a drive-by download.  We've already arguably lost that via Homebrew and all the Linux distros , so maybe that's a weak position to cling to.

We have the ability to serve betas on our download server if we need to distribute builds before we do a full release.

We've admittedly not done great with the community-building aspect in recent years. Do you think this helps or hurts that?

tsteven4:
All the CI builds embed the version number from GPSBabel.pro VERSION variable.  Currently the pre-releases masquerade as 1.7.0.
 
I like having the pre-release continuous releases available for testing by myself or a user I am engaged with under my direction.  Maintaining them as github releases is significantly easier now the CI is running via github actions.  I don't feel we need to advertise them. 

I have deleted any mention of pre-releases in the readme as well as the link to our [presubmit testing](https://www.gpsbabel.org/quality.html).  The official download page is mentioned.

tsteven4:
BTW you can see the rendered new readme on the branch for this PR: https://github.com/tsteven4/gpsbabel/tree/readme

PR #673, opened 2021-01-28
clang tidy readability fixes

tsteven4:
readability-make-member-function-const
readability-convert-member-functions-to-static
readability-const-return-type
readability-qualified-auto

astyle was used to match our formatting standards for readability-qualified-auto, which resulted in some formatting changes in geojson.cc

### DIFF
--- a/bend.cc
+++ b/bend.cc
@@ -52,7 +52,7 @@ void BendFilter::init()
 }
 
 Waypoint* BendFilter::create_wpt_dest(const Waypoint* wpt_orig, double lat_orig,
-                                      double long_orig, double lat_orig_adj, double long_orig_adj)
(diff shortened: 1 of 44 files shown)


robertlipe:
If clang-tidy and astyle disagree, I'd probably defer to clang-tidy these days. We (I) still have some dated opinion that date back to Bell-heritage.

I'd favor consistency over any one point, though.

auto foo vs auto* foo is kinda dumb, though. That's what auto means...

Whatever you want to do. Go for it.

tsteven4:
I agree with the motivation behind the tidy recommendation, I think it does make the code more readable.

> advises to make it obvious if a auto typed variable is a pointer.

I don't care much what the style is, but I also value consistency.

PR #679, opened 2021-02-04
replace obsolete and overloaded QProcess signal error(QProcess::Proce…

tsteven4:
…ssError error)

with the recommended replacement, QProcess signal errorOccurred(QProcess::ProcessError error).
It seems this should have caused an issue in Qt6, but we hadn't noticed it yet.


robertlipe:
IIRC, there's not much, if any, type checking in the signal/slot plumbing. That may be less true now, but there was a time you could just make up a connection and get no error. This case seems to support that. 

Good catch.

PR #681, opened 2021-02-05
delete FAKE_LANGUAGE_MENU code.

tsteven4:
This code hadn't compiled for a while, and is no longer necessary
to test translations.


robertlipe:
Ok. That was for OSes that didn't let you change the OS after it was
installed. Windows 7, maybe?

I'd say it's pretty solid these days, so bye, bye.

PR #685, opened 2021-02-07
prefer implicitly-defined dtors instead of user defaulted dtors or dtors with empty bodies.

tsteven4:
Not only is this simpler, it avoids clang-tidy complaining about
warning: class * defines a non-default destructor but does not define a copy
constructor, a copy assignment operator, a move constructor or a move
assignment operator [cppcoreguidelines-special-member-functions]


robertlipe:
I like it.

These were surely added to hush some hyperactive warnings some time ago.
Language moves on.

PR #687, opened 2021-02-08
cleanup filterwidgets.

tsteven4:
clang-tidy modernize-use-override
clang-tidy modernize-use-equals-default
clang-tidy readability-inconsistent-declaration-parameter-name
manual fix for clang-tidy readability-named-parameter

rule of zero implemented for FilterWidget class by using QSharedPointers.
rule of five implemented for FilterOption class by deleting copy and move
special member functions.

use c++17 std::clamp.


tsteven4:
The goals of this PR were a part of #1387

PR #688, opened 2021-02-10
make src/core and jeeps includes relative to top cli directory

tsteven4:
Don't use unix directory aliases in includes as suggested in https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes
Note that we mostly already did this in src/core.

We intentionally don't do this for libusb, shapelib and zlib.  Ideally those third party sources would be compiled as separate archives or libraries.  We also resist modifying them to simplify syncing with the upstream sources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,7 +206,6 @@ if(UNIX)
   set(SOURCES ${SOURCES} gbser_posix.cc)
   set(HEADERS ${HEADERS} gbser_posix.h)
   set(JEEPS ${JEEPS} jeeps/gpslibusb.cc)
-  include_directories(AFTER jeeps)
   add_compile_options(-O2 -Wall)
 endif()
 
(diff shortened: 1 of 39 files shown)


tsteven4:
In case we need do this again:

```
#!/bin/bash -ex
for file in *.cc *.h
do
echo $file
  ed -s $file <<"EOJ"
1,$g/#include/s/"\([^"]*\)"/"jeeps\/\1"/
1,$g/#include/s/"jeeps\/\.\.\/\([^"]*\)"/"\1"/
w
q
EOJ
done

# by hand gpsserial.cc "gbser_posix.h", "gbser_win.h"
```

robertlipe:
Can we remove the -I so it doesn't slip back in?

Good call. Thanks!

tsteven4:
Yes, the include was removed from GPSBabel.pro and CMakeLists.txt.

PR #692, opened 2021-02-13
use object tree to cleanup UpgradeCheck, UpgradeCheck::manager_ and UpgradeCheck::replyId.

tsteven4:
~QNetworkAccessManager() promises to delete any QNetworkReply objects that are returned by the class.


robertlipe:
Good idea.QNetworkAccessManager didn't always have this option. This code
began life with a different network manager back in the Qt 4.1 days or so.

PR #693, opened 2021-02-14
rework progresswait

tsteven4:
eliminate potential memory leak in runGpsbabel that could occur
if the process immediately returned NotRunning after being started.

eliminate potential race condition (that apparently never happens)
in checking process state after staring.  It seems best to hook up
all the signals and then start the process, ensuring that any errors
or finished signals will be caught.  This also simplifies the logic,
the start check is repaced by normal signal handling.


tsteven4:
Abandoned in favor of #699

PR #699, opened 2021-02-23
Introduce RunMachine to manage gpsbabel process.

### LINKED ISSUE #696: GUI error highlighting?
Should we highlight errors in the GUI from running the CLI gpsbabel?  Note that standard error and standard output from the gpsbabel process are returned as a merged QString, so we currently can't distinguish them.  We could keep them separate, but then we lose the relative order of any messages.  But, we can distinguish any error message the GUI generates based running the CLI.  For example

image

### LINKED ISSUE #697: GUI process wait dialog issues
There are multiple issues with the process wait dialog.  The dialog is rarely seen because most operations are fast.

On windows with 1.7.0:
1. A help icon ('?') is shown in the title bar, but doesn't appear to do anything.  In Qt6 may not happen: Qt::WindowContextHelpButtonHint may not be set by default.
2. pressing the escape key will kill the dialog and we report "Translation successful".  However, this is not the case.  The output may not have been created, or may be partially created.
3. pressing the "Stop process" button doesn't appear to do anything.
4. pressing the enter key doesn't appear to do anything.  This should act like the dialogs default button was pressed.
5. pressing the x button in the title bar doesn't appear to do anything.
6. if a large file is send to standard out the process wait dialog locks up.

On macOS and linux the issues are different.

image

tsteven4:
This is a recommended alternative to #693.

This resolves #696, implementing the suggested enhancement.

This fixes #697, with a few caveats:

- the process wait dialog intentionally ignores close events (pressing the x button in the title bar doesn't appear to do anything.)  This doesn't seem like good interface design, but that is what we have always done.  Pushing the "Stop process" button, or pressing enter (the default button), or pressing escape all result in the process being aborted and the process wait dialog closing.
- If the user selects '-' for the output file, and the file is very large, e.g. "-i gpx -f big.gpx -o gpx -F -", there will be a delay of many seconds after the process wait dialog closes while the output is converted to a QString and appended to the output window.  In process wait we now truncate the plain text window to the last 16384 characters which keeps it speedy.  We could do something similar in mainwindow, perhaps keeping the first and last 16384 characters with a "..." indication in between.  Unlike with process wait the mainwindow output window stays open and the user does have an opportunity to scroll back and see the entire window contents.

Additionally, the proper codec is used to translate standard output and error before display.  This was another unreported bug.

### DIFF
--- a/gui/mainwindow.cc
+++ b/gui/mainwindow.cc
@@ -43,6 +43,8 @@
 #include <QtGui/QDesktopServices>      // for QDesktopServices
 #include <QtGui/QIcon>                 // for QIcon
 #include <QtGui/QImage>                // for QImage
+#include <QtGui/QTextCharFormat>       // for QTextCharFormat
+#include <QtWidgets/QAbstractButton>   // for QAbstractButton
 #include <QtWidgets/QApplication>      // for QApplication, qApp
 #include <QtWidgets/QCheckBox>         // for QCheckBox
 #include <QtWidgets/QDialogButtonBox>  // for QDialogButtonBox
@@ -70,7 +72,7 @@
 #include "help.h"                      // for ShowHelp
 #include "optionsdlg.h"                // for OptionsDlg
 #include "preferences.h"               // for Preferences
-#include "processwait.h"               // for ProcessWaitDialog
+#include "runmachine.h"                // for RunMachine
 #include "upgrade.h"                   // for UpgradeCheck
 #include "version_mismatch.h"          // for VersionMismatch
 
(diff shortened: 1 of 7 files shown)


gui/processwait.h
```diff
@@ -23,64 +23,48 @@
...
-  int                  ecode_;
-  QProcess*     process_;
-  QProgressBar* progressBar_;
-  QPlainTextEdit* textEdit_;
-  QDialogButtonBox* buttonBox_;
-  QTimer*           timer_;
-  QString          errorString_;
-  QString          outputString_;
+  int progressIndex_{0};
+  QProcess* process_{nullptr};
```
robertlipe, line 62: This isn't necessarily in scope for this change, but these raw pointers should probably be RAII as they don't need to outlive this class, I don't think.

gui/runmachine.cc
```diff
@@ -0,0 +1,218 @@
...
+            std::optional<QProcess::ExitStatus>(exitStatus));
+  }, Qt::QueuedConnection);
+  connect(process_, &QProcess::started,
+  this, [this]() {
+    execute(processStarted,
+            std::nullopt,
+            std::nullopt,
+            std::nullopt);
+  }, Qt::QueuedConnection);
+  connect(progress_, &ProcessWaitDialog::rejected,
```
robertlipe, line 87: Lambdas make this sort of thing so much cleaner. Nice!

gui/runmachine.cc
```diff
@@ -0,0 +1,218 @@
...
+    }
+    if (exitStatus.has_value()) {
+      debugStream << *exitStatus;
+    }
+    if (exitCode.has_value()) {
+      debugStream << *exitCode;
+    }
+  }
+
+  switch (state_) {
```
robertlipe, line 132: A FSM captures this so nicely.

gui/processwait.cc
```diff
@@ -82,138 +61,60 @@ ProcessWaitDialog::ProcessWaitDialog(QWidget* parent, QProcess* process):
...
+// Thus, we set the plain text to the last bit of output.
 //
-void ProcessWaitDialog::appendToText(const char* ptr)
+void ProcessWaitDialog::appendToText(const QByteArray& text)
 {
-  outputString_ += QString(ptr);
-  for (const char* cptr = ptr; *cptr != 0; cptr++) {
-    if (*cptr == '\r') {
-      continue;
+  static constexpr int textLimit = 16384;
```
robertlipe, line 100: Good find. I'm not sure that writing to stdout inside the GUI is something that normal people do.
Is it worth a warning that it's been truncated if outputString_.size() > textLimit? (Maybe not. See above...)

gui/mainwindow.cc
```diff
@@ -1038,7 +1017,14 @@ void MainWindow::applyActionX()
     }
 #endif
   } else {
+    QTextCharFormat defaultFormat = ui_.outputWindow->currentCharFormat();
+    QTextCharFormat errorFormat = defaultFormat;
+    errorFormat.setForeground(Qt::red);
```
robertlipe, line 1022: Styling this is such a good idea. Why didn't we think of that 15 years ago? :-)

robertlipe (approved):
I like this a lot. So much cleaner a good use of more modern C++.

Thanx for tackling another long-standing chunk of code that's been neglected for a long time.

gui/processwait.h
tsteven4: All the corresponding pointed to objects go on the object tree which takes care of their deletion.

PR #700, opened 2021-03-04
modernize nmea reader

tsteven4:
use QDate, QTime, QDateTime instead of struct tm, time_t.

enhance nmea date rollover testing, and fix a rollover bug when
a new track point without a date follows a previous point on the
other side of midnight.

use const Qt containers when convenient avoiding the need for Qt
to check for shared usage.

enhance nmea checksum verification. previously this could fail if
a sentence was truncated past the '*' checksum indicator.


robertlipe:
I've started this a few times, so I'm glad you did it. Thanx and approved.

PR #707, opened 2021-03-26
PVS-Studio fixes

tsteven4:
This is a set of fixes for various issues flagged by PVS-Stuido.

V1064 The '360' operand of integer division is less than the '65535' one. The result will always be zero.
V547 Expression is always false.
V758 The 'pos' reference becomes invalid when temporary object returned by a function is destroyed.
V668 There is no sense in testing the 'wpt_tmp' pointer against null, as the memory was allocated using the 'new' operator. The exception will be generated in the case of memory allocation error.
V567 The modification of the 's' variable is unsequenced relative to another operation on the same variable. This may lead to undefined behavior.
V646 Consider inspecting the application's logic. It's possible that 'else' keyword is missing.
V576 Incorrect format. Consider checking the second actual argument of the 'asprintf' function. The SIGNED integer type argument is expected.

### DIFF
--- a/brauniger_iq.cc
+++ b/brauniger_iq.cc
@@ -54,7 +54,7 @@ inline state_t& operator++(state_t& s) // prefix
 inline state_t operator++(state_t& s, int) // postfix
 {
   state_t ret(s);
-  s = ++s;
+  ++s;
   return ret;
 }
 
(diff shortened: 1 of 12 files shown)


robertlipe:
Fascinating what these tools can find, even when this body of code
regularly goes through so many *other* gauntlets.

Thanx, as always.

PR #710, opened 2021-04-16
Minorfixes to "new" gtm_stringread() functions to eliminate leading and trailing spaces

robertlipe:


dmtlog.cc
```diff
@@ -425,8 +418,8 @@ read_CTrackFile(const int version)
 
   /* S1 .. S9: comments, hints, jokes, aso */
   for (int i = 0; i < 9; i++) {
-    char* s = read_str(fin);
-    xfree(s);
+    auto s = read_str(fin);
+    Q_UNUSED(s);
```
tsteven4: or without Qtisms"(void) read_str(fin);"

gtm.cc
```diff
@@ -79,20 +79,8 @@ static QString
...
-  gbfread(val, 1, len, fd);
-  while (len != 0 && val[len-1] == ' ') {
-    len--;
-  }
-  val[len] = 0;
-  QString v(val);
-  xfree(val);
-  return v;
+  QByteArray bytes = gbfreadbuf(len, fd);
+  return QString(bytes).trimmed();
```
tsteven4: or "return gbfreadbuf(len, fd).trimmed();"
QByteArray also has trimmed, and it's not necessary to use the QString constructor explicitly.

gtm.cc
```diff
@@ -104,17 +92,8 @@ fread_string_discard(gbfile* fd)
...
-  while (len != 0 && val[len-1] == ' ') {
-    len--;
-  }
-  val[len] = 0;
-  QString v(val);
-  xfree(val);
-
-  return v;
+  QByteArray bytes = gbfreadbuf(len, fd);
+  return QString(bytes);
```
tsteven4: or "return gbfreadbuf(len, fd);"

dmtlog.cc
```diff
@@ -329,21 +330,16 @@ tlog3b_xgcb_wpten(xg_string, const QXmlStreamAttributes*)
...
 
-  char* res = (char*) xmalloc(i + 1);
-  res[i] = '\0';
-  if (i) {
-    gbfread(res, 1, i, f);
-  }
-
-  return res;
+  auto res = gbfreadbuf(i, f);
+  return QString(res);
```
tsteven4: or "return gbfreadbuf(i, f);", no need to call QString constructor.

vitosmt.cc
```diff
@@ -128,20 +125,16 @@ vitosmt_read()
...
-    tmStruct.tm_hour =timestamp[3];
-    tmStruct.tm_min =timestamp[4];
-    tmStruct.tm_sec  =(int)floor(seconds);
-    tmStruct.tm_isdst =-1;
+    QDate wdate(timestamp[0] + 2000, timestamp[1], timestamp[2]);
 
     double usec = fmod(1000000*seconds+0.5,1000000);
-    wpt_tmp->SetCreationTime(mkgmtime(&tmStruct), lround(usec/1000.0));
+    QTime wtime(timestamp[3], timestamp[4],
+                (int) floor(seconds), lround(usec/1000.0));
```
tsteven4: This is problematic if usec/1000.0 rounds up to 1000, in that case wtime will be invalid.

vitosmt.cc
```diff
@@ -128,20 +125,16 @@ vitosmt_read()
...
-    tmStruct.tm_year =timestamp[0]+100;
-    tmStruct.tm_mon =timestamp[1]-1;
-    tmStruct.tm_mday =timestamp[2];
-    tmStruct.tm_hour =timestamp[3];
-    tmStruct.tm_min =timestamp[4];
-    tmStruct.tm_sec  =(int)floor(seconds);
-    tmStruct.tm_isdst =-1;
+    QDate wdate(timestamp[0] + 2000, timestamp[1], timestamp[2]);
 
     double usec = fmod(1000000*seconds+0.5,1000000);
```
tsteven4: This is problematic.  It can result in incorrect rounding, e.g.
seconds: 59.9999999 usec: 0.3999999985098839

vitosmt.cc
```diff
@@ -128,20 +125,16 @@ vitosmt_read()
...
-    wpt_tmp->SetCreationTime(mkgmtime(&tmStruct), lround(usec/1000.0));
+    QTime wtime(timestamp[3], timestamp[4],
+                (int) floor(seconds), lround(usec/1000.0));
+    wpt_tmp->SetCreationTime(QDateTime(wdate, wtime, Qt::UTC));
```
tsteven4: I suggest something like:
wpt_tmp->SetCreationTime(QDateTime(QDate((timestamp[0] + 2000, timestamp[1], timestamp[2]), QTime(timestamp[3], timestamp[4]), Qt::UTC).addMSecs(lround(seconds*1000.0));

PR #714, opened 2021-05-19
fix map preview bug that dropped tracks with 2 pts.

tsteven4:
This came up with https://opendata.swiss/en/dataset/luftfahrthindernisdaten-schweiz/resource/2cca4304-023a-48c0-9a41-f2802e814795
The cited example was track 296-TI-4 although other tracks were dropped as well.


robertlipe:
LGTM

New loop style is easier to read. Thanks.

PR #716, opened 2021-06-15
Add warnings for important lines in NMEA that may be ignored due to known invalid state.

robertlipe:
Add warnings for important lines in NMEA that may be ignored due to known invalid state.
I don't care that the error path might be slow. I care that it's automatable, self resetting between reads, and that the displayed messages stay in sync.

TODO: decide between std::map/std::out and QMap/qWarning & test case.


nmea.h
```diff
@@ -169,6 +170,8 @@ class NmeaFormat : public Format
 
   int wpt_not_added_yet{};
 
+  std::map<std::string, int> counter;
```
tsteven4: For consistency I'd prefer to stick with Qt containers unless there is a compelling advantage to std lib in a particular use case.

nmea.cc
```diff
@@ -25,6 +25,7 @@
 #include <cstdio>                  // for snprintf, sscanf, fprintf, fputc, stderr
 #include <cstdlib>                 // for atoi, atof, strtod
 #include <cstring>                 // for strncmp, strchr, strlen, strstr, memset, strrchr
+#include <iostream>                // for std::cout
```
tsteven4: For consistency I'd prefer to stick with Qt streams and our subclasses.  I acknowledge that the Qt namspace changes introduced in 5.14, e.g. endl -> Qt::endl, are either generating warnings are breaking compatibility with 5.12.

tsteven4: The above concern about endl has been taken care of in src/core/logging.h, you can use Qt::endl for all Qt versions we support.

nmea.h
GPSBabelDeveloper: Agreed. Changed and renamed.

GPSBabelDeveloper:
This code is in master and has shipped.

PR #718, opened 2021-08-26
Add ttyACMO as option in gpsbabel gui

dk5ee:
On newer linux systems serial devices often are mapped to ttyACMx, not traditional ttyUSBx.

To facilitate the usage, instead of forcing the user to fiddle with udev/systemd device mapping, this small commit adds ttyACM0 as an option in the gui.

### DIFF
--- a/gui/serial_unix.cc
+++ b/gui/serial_unix.cc
@@ -97,6 +97,7 @@ static const char* deviceNames[] = {
   "/dev/ttyS2",
   "/dev/ttyS3",
   "/dev/ttyUSB0",
+  "/dev/ttyACM0",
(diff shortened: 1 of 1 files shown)


robertlipe:
Thanks, but I thought this code was on the way out several years ago. I
remember using https://doc.qt.io/qt-5/qserialportinfo.html#availablePorts
so we didn't have these hardcoded lists at all.

Can you confirm that static method works for your ports?

Maybe we chicken out and keep the static list, add whatever we find with
this, and just use the unique union of those sets...

tsteven4:
Thanks for pointing out the issue, however I am closing this without merging in favor of the implementation in #755.

PR #723, opened 2021-09-27
fixes for #722, Mechanically remove QStringRef use for QStringView to help with Qt6.

tsteven4:
This contains cherry-picks of the 3 commits in the stringview branch that were added after #722 was merged.

This fixes CI of the master branch.


tsteven4:
Note QStringViews are still being passed as const references instead of by value as recommended.

PR #734, opened 2021-10-13
tpo fixes for #643

tsteven4:
mismatched format string

astyle


tsteven4:
I'll take a second swab at this in another PR.

PR #738, opened 2021-10-15
fix an out of bounds access to a QString in tpo.

tsteven4:
This was caught with Qt6.2.0, which has an assertion to check for
out of bounds access.
ASSERT: "i >= 0 && i < size()" in file .../Qt-6.2.0/6.2.0/gcc_64/include/QtCore/qstring.h, line 1325


robertlipe (approved):
Please just whack that comment as it no longer makes sense.

PR #739, opened 2021-10-17
Add support for Qt6.

tsteven4:
This also maintains the current Qt floor of 5.12.


robertlipe (approved):
Thanx for working through a number of icky issues to get this landed. The code looks great to me.

PR #743, opened 2021-10-23
replace the last uses of QRegExp.

tsteven4:
also, correct the discard documentation.  The given example didn't
work in recent releases.

### DIFF
--- a/discard.cc
+++ b/discard.cc
@@ -19,13 +19,16 @@
 
  */
 
-#include "defs.h"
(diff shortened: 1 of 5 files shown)


robertlipe:
If asked to write a regex that matched a fix number of characters, I'd use dots over question marks but that's just the world we have with multip;le regex syntaxes, I suppose. Thanx for the fixes. Go for it.

PR #744, opened 2021-10-23
update included libusb for mac to 1.0.24

tsteven4:


tsteven4:
Our files that changed are CMakeLists.txt, libusb.pri, mac/libusb/libusb.pro, mac/libusb/README.  The rest of the changes are straight from the 1.0.24 distribution of libusb.

libusb 1.0.24 is used by homebrew without patches.

robertlipe (approved):
Though USB Garmin protocol was kind of a peak for us,  I'm really not sure if anyone's still using it. Garmin hasn't engineered a device using this since 2006, though they still sold a few into the late 200x's.

Does anyone on the list still care about pre USB filesystem era Garmins?

PR #748, opened 2021-10-28
fix windows release with Qt6.

tsteven4:
windeployqt failed to deploy Qt6Core5Compat.dll because it was
not used by the gui and we didn't tell windeployqt there was
another executable to scan for dependencies.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -111,10 +111,10 @@ jobs:
 
     - name: Qt install
       if: steps.cache.outputs.cache-hit != 'true'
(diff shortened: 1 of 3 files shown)


tsteven4:
This all started with make_windows_release.ps1 not passing both executables to windeployqt.  Then we ran into problems with  cache and aqt install on macos.  The conditionals in macos.yml weren't working for a while, apparently due do a runner change.  The cache was found, but then we tried to run aqt anyway which failed.  Now the conditionals are working, but we also have a second check in travis_install_osx.

PR #750, opened 2021-10-31
C++ improvements in VCard. Add tests.

robertlipe:


vcf.cc
```diff
@@ -64,15 +64,14 @@ vcf_print_utf(const utf_string* s)
...
-  xfree(p2);
-  xfree(p3);
-  xfree(stripped_html);
+  char *string = strip_html(s);
+  QString stripped_html = string;
+  xfree(string);
+
+  stripped_html.replace("\n", "\\n");
+  stripped_html.replace("<p>", "\\n");
+  stripped_html.replace(";", "\\;");
```
tsteven4, line 73: I think commas need to be escaped as well, see https://datatracker.ietf.org/doc/html/rfc2426#section-2.4.2
`stripped_html.replace(",", "\\,");`

GPSBabelDeveloper:
Thanx. I've incorporated that and am awaiting build results before I
collapse it.

There's a lot of HTML we probably could/should do better with,  This format
is sitting right on the edge of the chopping block so I didn't get to
ambitious with it by, but I actually used it for sending coordinates to a
Tesla, of all things, just a few months ago. Still, this format has been
with us for many years and nobody's noticed the comma thing, so how popular
is it really? {shrug}

Using qtextdocument() to replace the "hard" part of HTML is tempting, but
linking our CLI against QtGUI would probably make someone unhappy...
especially for a format on death's door.

PR #751, opened 2021-10-31
Eliminate internal copy/free inside navilink.

robertlipe:
- C++ improvements inside Vcard. Add test.
- Eternal impedance mismatch on C vs C++ pointer ownership.


tsteven4:
This branch is identical to #750 and has nothing to do with navilink.

robertlipe:
Git sure made a mess of keeping my two branches separate, leading to a lot of busy-work for a one line change. :-(

PR #753, opened 2021-11-02
optimize speed and distance conversions.

tsteven4:
structure constexpr conversions such that a minimal amount of
arithmetic is done at run time and run time division is avoided.

add tests for conversions.


robertlipe (approved):
Clever!

Issue #337, opened 2019-04-16
serial port identification


tsteven4:
Can gui/serial_mac.cc, gui/serial_unix.cc, gui/serial_win.cc all be replaced with
```
#include <QtCore/QList>                    // for QList
#include <QtSerialPort/QSerialPortInfo>    // for QSerialPortInfo
#include <QtWidgets/QComboBox>             // for QComboBox

#include "mainwindow.h"


void MainWindow::osLoadDeviceNameCombos(QComboBox* box)
{
  const auto ports = QSerialPortInfo::availablePorts();
  for (const auto& info : ports) {
    box->addItem(info.systemLocation());
  }
}
```
This has the potential to resolve [https://bugzilla.redhat.com/show_bug.cgi?id=1337062](url)

robertlipe:
Yes. See
https://github.com/gpsbabel/gpsbabel/pull/42#diff-for-comment-58104108

I meant to do this years ago and, in fact, thought I had.

tsteven4:
On Linux (ubuntu bionic) it looks like we need systemLocation() to get things like /dev/ttyS0 instead of ttyS0, but on windows systemLocation gives things like \\.\COM1, and we probably want portName() instead.

robertlipe:
That's kind of gross. The qt doc is uncharacteristically unhelpful on the
difference.

ISTR that once the "X" in "COMx:" grew outside the bounds of [1-9]
 (somewhere around Windows 3.0) that they added the \.\COMx convention
<https://support.microsoft.com/en-us/help/115831/howto-specify-serial-ports-larger-than-com9>
so maybe this is something that Windows people just accept as normal these
days. That's when they added the whole notion of "friendly names" so you
could pick "3Com Modem" instead of "COM2" and certainly the common use of
BT and USB "serial" devices really doesn't fit well into the "COMx:"
convention. So how about we display one but use the other?

addItem(info.portName(), QVariant(info.systemLocation));

Then when we pull the name out of the array we read
box->currentIndex()->toString() in the onChange reader of the slot or
whatever.

That way we display the pretty names for humans but use the /dev and
"\.\COM" style names for the actual command line and resulting calls to
open ... though I couldn't fake shock if open() and CreateFile don't do the
same thing on Windows.

Man, so much for our awesome two line refactoring...

tsteven4:
I have something to test, but no means to really test it.

On windows the GUI looks like this, this is in a VM with nothing connected to COM1 or COM2.

[...]

PR #755, opened 2021-11-03
serial port names

tsteven4:
Try to leave identification of serial ports to Qt.  I am trying to leave the interface to the CLI (gpsbabel) alone, and confine the changes to the GUI.

I do see a difference on a macos data center system.  With the change I see
image
Without the change (baseline) the tty option is missing.

I haven't been able to test com ports >9 on windows.  I think we have some difficulty there in any case with gbser_win.cc define DEV_PREFIX "\\\\.\\\\" containing an extra trailing backslash.  But on my system the choices are identical, but this isn't really a test with a live device.

Also, on ubuntu vm I see the same choices, but this isn't really a test with a live device.

This has had a long history of not getting committed, e.g. https://github.com/GPSBabel/gpsbabel/issues/337, https://github.com/GPSBabel/gpsbabel/pull/718, ...

Note usb: is still added by mainwindow.cc for formats that support usb.

@dk5ee, does this work for you?  Can we test this enough to do it?  It does eliminate (or move into Qt) a bunch of OS specific code.


tsteven4:
Considerations:
1. Windows.  There is a history of windows version specific issues and \\\\:\\COM10 issues noted in gbser_win.cc.  The names the GUI has passed to the CLI seem to match QSerialPortInfo::portName values and not QSerialPortInfo::systemLocation values.  Furthermore, there appear to be bugs in our \\\\:\\ handling.  I would prefer to leave the CLI alone and continue to pass the same names we always have, from both the GUI and the human.    The hard coded list we have is conditionally compiled out.  Maybe we just leave serial_win.cc alone. 
2. Mac.  There isn't a hard coded list.  QSerialPortInfo::systemLocation values seem to match current behavior.  The proposed change seems relatively safe to me.
3. Non Mac Unix.  There is a hard coded list that is appended to anything found with udev.  The list is
```"/dev/ttyS0",
  "/dev/ttyS1",
  "/dev/ttyS2",
  "/dev/ttyS3",
  "/dev/ttyUSB0",
  "/dev/rfcomm0",
```
There is a history here of not finding things, e.g. #718.  QSerialPortInfo appears complicated, with distinctions for osx, win, freebsd and unix.  The Qt unix code has a longer hard coded list than we do.  I have seen a report of QSerialPortInfo success finding ttyUSB0 and ACM0 in other projects.  QSerialPortInfo::systemLocation values seem to match current behavior.  I think relying on Qt here is a good idea.

The problem with all serial port changes is there are nearly impossible to test.  Can we get some manual testing done on mac and linux, and just leave windows alone, i.e. revert the proposed change to serial_win.cc?

tsteven4:
FYI here is what I see on a ubuntu VM:
baseline:
image
With change.  Only one device is returned, so "If only one choice, just disable it.":
image
If I enable another serial port in the VM then I see them both:
image

robertlipe:
Yes, please. I think I've written the meat of this patch a couple of times (as you mention) but never committed it. (Add #178 and #42 to the list. It probably predates our move to GitHub...) It's time to just go for it.

I'm surprised that you ever see 32 ports on Linux without associated hardware.

Please just commit this. I'll spot test USB adapter recognition (not necessarily serial GPSes because level shifting 3.3V to serial devices is a problem) with the three currently trendy USB/Serial adapters: FTDI(ish) CH340, and CP2102.

There's some irony in our world that most "RS232" devices we originally supported are electrically incompatible with most USB adapters you can buy, or at least require careful recognition of 3.3V vs. 5V settings. The original +/-12 VDC just isn't an option; buck-boost converters are like impossible to find on these this days. MOST cases that we call serial in modern times is a USB logging device using one of those three chips internally.

tsteven4:
I'll include this change for all systems, including windows.  On windows I will make a few more changes so the text matches what device manager shows:
image

and in device manager we see:
image

dk5ee:
Looks great, good work. Unfortunatly I have no time right now to test it, but thanks.

PR #759, opened 2021-11-08
Fix build on Catalina, HTML doc formatting improvements

GPSBabelDeveloper:
- Build fix for Catalina (10.15.7)


mac/libusb/libusb.h
```diff
@@ -25,6 +25,8 @@
 #ifndef LIBUSB_H
 #define LIBUSB_H
 
+#include "mac/libusb/Xcode/config.h"
```
tsteven4: This should be unnecessary, libusb.pri should take care of it.  Is there an issue with cmake or xcode?  In any event let's keep any needed changes in a separate commit and PR.

tsteven4: cmake works for me on macos 11:
```
-- Defines are: "HAVE_UNISTD_H;HAVE_STDARG_H;HAVE_LIBUSB_1_0;MAXIMAL_ENABLED;FILTERS_ENABLED;SHAPELIB_ENABLED;CSVFMTS_ENABLED"
-- Libs are: "Qt5::Core"
-- Include Directores are: "/Users/administrator/serialnames2.1/zlib;/Users/administrator/serialnames2.1/mac/libusb;/Users/administrator/serialnames2.1/mac/libusb/Xcode"
-- Configuring done
-- Generating done
```

tsteven4: This also works for me, but now I am just guessing what you might be doing
```
cmake -G Xcode .
xcodebuild -project GPSBabel.xcodeproj -configuration Release
```

"qmake -spec macx-xcode" does not work because of the configuration tests.

tsteven4: This also works for me, but now I am just guessing what you might be doing
```
cmake -G Xcode .
xcodebuild -project GPSBabel.xcodeproj -configuration Release
```

"qmake -spec macx-xcode" does not work because of the configuration tests.

tsteven4: I think the long range plan here is to have a hierarchical build where we can better sequester the defines and includes for dependencies from gpsbabel.  The change you have isn't compatible with that goal.

tsteven4: I have a fix for generating xcode projects with qmake if that's an issue.

GPSBabelDeveloper:
It is was qmake that fails.

tsteven4:
Were using qmake to make a Makefile or a Xcode project?

Can you back out your change and reproduce the failure?  I would like to see a transcript.  I have never seen a failure in CI or at mac stadium.  CI is running on catalina and big sur using qmake to generate makefiles.

tsteven4:
By any chance did you have an old config.h file sitting in the top of your build tree?

PR #761, opened 2021-11-11
Fix kml bounding box issues.

tsteven4:
The kml bounding box had problems around the antimeridian, that were unsuccessfully mitigated in 2011. Instead of fixing the obvious mistake in that mitigation we use nvectors to compute the actual center of the bounding box.

A test case is added that shows the gross error around the antimeridian.

Nvectors will be useful to avoid problems at discontinuities and singularities that are inherent in the latitude, longitude coordinate system.  Our implementation has gone extensive testing that isn't apparent in this commit.  Planned enhancements will use more of the capability of nvectors.

We also restore some lost whitespace in gc_logs data.  The whitespace change in LineStyles.kml is a vertical tab that is handled differently than the distant past.


src/core/nvector.cc
```diff
@@ -0,0 +1,451 @@
...
+  if (result < 0.0) {
+    result += 360.0;
+  }
+  if (result >= 360.0) {
+    result = 0.0;
+  }
+  return result;
+}
+
+#if 0
```
robertlipe, line 208: See ifi the four 'if 0' blocks in this file really do deserve to stay.

src/core/nvector.cc
```diff
@@ -0,0 +1,451 @@
...
+  double sab_rad = distance_radians(n_EA_E, n_EB_E);
+  double sar_rad = fraction*sab_rad;
+  double az_rad = azimuth_radians(n_EA_E, n_EB_E);
+
+  // equation 9 (normalized)
+  Vector3D kaeast = crossProduct(Vector3D(1.0, 0.0, 0.0), n_EA_E).normalize();
+  // equation 10 (normalized, n_EA_E and kaeast are perpendicular unit vectors)
+  Vector3D kanorth = crossProduct(n_EA_E, kaeast);
+
+  Vector3D d_EA_E = kanorth*cos(az_rad) + kaeast*sin(az_rad);
```
robertlipe, line 228: Please run astyle or clang-tidy to format. This is one of many places that binary operand spacing is inconsistent. Here it's both wrong (*) and right (+) in the same line.

src/core/nvector.h
```diff
@@ -0,0 +1,94 @@
...
+constexpr double WGS84_ECCENTRICITY_SQUARED = 1.0 - (WGS84_ASPECT_RATIO * WGS84_ASPECT_RATIO);
+
+class PVector;
+
+class LatLon
+{
+public:
+  LatLon(double latitude, double longitude);
+
+  double lat;
```
robertlipe, line 55: Can class members be foo_ instead of foo? Or is that too mucky given how prevalent thinks like this and vec3d are in our code.

reference/track/gtrnctr_power-kml.kml
```diff
@@ -8,10 +8,10 @@
         <end>2010-05-28T02:41:44Z</end>
       </gx:TimeSpan>
       <longitude>-122.139608</longitude>
-      <latitude>37.382794</latitude>
+      <latitude>37.382814</latitude>
```
robertlipe, line 11: @viettaml , note this CL is going to change the LookAt (and thus, FlyTo) on files we handle. It's small on most common data and cases involving > hemisphere of polygons or places spanning hemispheres are subject to weird stuff already.
We can at least reduce OUR weird stuff.

src/core/nvector.cc
```diff
@@ -0,0 +1,451 @@
...
+
+double NVector::longitude() const
+{
+  // this implements equation 5.
+  double longitude_radians = atan2(y, -z);
+  double longitude_degrees = longitude_radians * 180.0/M_PI;
+  return longitude_degrees;
+}
+
+// great circle distance in radians
```
robertlipe, line 115: Sentence casing and punctuation, please.

src/core/nvector.cc
```diff
@@ -0,0 +1,451 @@
...
+double NVector::distance_radians(const NVector& n_EA_E, const NVector& n_EB_E)
+{
+  // this implements equation 16 using arctan for numerical accuracy.
+  double result = atan2(crossProduct(n_EA_E, n_EB_E).norm(),
+                        dotProduct(n_EA_E, n_EB_E));
+  return result;
+}
+
+// great circle distance in meters
```
robertlipe, line 124: Timid suggestion. ("No" is OK and I know we have a million other places already suffering from this...)
Instead of everything being a double should we have classes for Radians, Degrees and Feet or Meters along with overloads that convert between them? That would let us strongly typecheck arguments so we don't swap a Lat with a Lon or a degrees with a radian again. A 5_meters could convert to feet, but not to degrees.

Maybe c++ proposal P1935 is relevant or at least inspirational, but that's been punted to C++23 at least. We probably need a much simpler subset.

I'll also accept "sounds good, but I don't care enough to change it". :-) Things like "class Waypoint should store an NVector instead of a lat/lon tuple" fall into that same category.

src/core/nvector.cc
tsteven4: The only thing astyle changes with our style file seems incorrect, it eliminates the space before the '*' in
WGS84_ECCENTRICITY_SQUARED = 1.0 - (WGS84_ASPECT_RATIO * WGS84_ASPECT_RATIO)

I don't want to get into formatting flip flops with two rulers, i.e. astyle and clang-format.

tsteven4: astyle could do what I think you want with "--pad-oper", but that hasn't been part of our style file.

src/core/nvector.h
tsteven4: The LatLon member variables are public.

I actually tried to change the Vector3D fields as you suggest but I got into a battle with clion I didn't want to fight.  I could try again.  I note our getters for that class violate https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c131-avoid-trivial-getters-and-setters, and if we wanted to heed that guideline then the trailing underscore is undesirable.  But, at this point, the fields are not public and could use a trailing underscore.

reference/track/gtrnctr_power-kml.kml
tsteven4: Note that the added test case in kml.cc demonstrates the existing code flying to the wrong place half way around the world.  In working a another enhancement I got so sick of GE flying to the wrong side of the world I made a detour (this PR) to fix it.

The small changes in latitude are not due to finite precisions arithmetic, they are due to the span of longitude in the bounding box which moves the midpoint along a great circle diameter towards the pole.

Also note the code currently has COMPARE_BEARING_TO_GRTCIRC defined, which matches our grtcirc calculations, but uses a spherical earth with the equatorial radius.  If we correct(undefine) that and use nvectors instead of grtcirc in the future a bunch of reference files will need to be tweaked.

src/core/nvector.cc
tsteven4: I note that with
NVector n1;
NVector n2;
n1+n2 has type Vector3D, not NVector.
The non-explicit single parameter constructors for NVector and PVector are how we can get an NVector back.  Of course clang-tidy will point out they are non-explicit single parameter constructors.  I fooled around with this, even making Vector3D a template class, but didn't get anywhere I liked.

src/core/nvector.cc
tsteven4: I pushed another commit with improvements to capitalization and spelling.

src/core/nvector.h
tsteven4: I renamed the member variables of Vector3D as you suggested.

tsteven4:
There is another old related issue.  The bounding box longitudes approach -180 and 180 for a track like sim.csv that crosses the anitmeridian, where they should approach the max of the negative longitudes and the min of the positive longitudes.  This effects our range calculation.

tsteven4:
What I believe will be a superior solution is draft PR #771

Issue #719, opened 2021-09-11
Locations.fit on Garmin devices


fabzzap:
Garmin Edge Explore devices, and possibly more Garmin Edge device models, store saved locations in a file called Locations.fit .

I wanted to convert such a file to a GPX with only waypoints (section `wpt`), but GPSBabel produced an empty GPX file.
```<?xml version="1.0" encoding="UTF-8"?>
<gpx version="1.0" creator="GPSBabel - https://www.gpsbabel.org" xmlns="http://www.topografix.com/GPX/1/0">
  <time>2021-09-11T14:10:32.576Z</time>
  <trk/>
</gpx>
```
Digging into the code, it turns out that the `global_id` of that file is 0x1d (29), and garmin_fit.cc's code does not support it.

Garmin makes a FIT SDK available for download on a restrictive licence. While I couldn't find anything in that licence that prohibits working on open-source software to support the FIT format, I'd like to know others' opinion about that before downloading it (which requires accepting the licence).

Also, more in general, are there other ways to approach the developers? I couldn't find an IRC channel, and, while the Sourceforge page looks abandoned, the mailing list gpsbabel-misc@sourceforge.net had posts as late as June 2021, which makes its active/abandoned status unclear.

tsteven4:
Thanks for checking about the license.  We have rejected the fit sdk license before, e.g. see https://sourceforge.net/p/gpsbabel/mailman/message/36145201/  If you want to discuss this further @robertlipe needs to make the call.

sourceforge is abandoned except for the gpsbabel-misc, gpsbabel-code, gpsbabel-announce.  See https://www.gpsbabel.org/lists.html .  Other than that communication is via github as you found.

GPSBabelDeveloper:
Hi, and thanx for asking.

Restricted licenses just don't work for us. It's not like we can enforce
people to accept a third party license when they download a copy that we
include.

As for the quietness on our lists, I've noticed that. Our download and use
counters have dropped from peak, but this tracks use of dedicated gps in
general. We're a twenty year old program and have reached a quite stable
base of code and users.

fabzzap:
For sure one can't copy any code from the FIT SDK into an open-source project. But: are those who accepted the FIT SDK licence banned from developing FIT support in GPSBabel, even if they just reimplement the protocol with 100% new code?

If the answer is "yes", I will forget about accepting it (which I haven't done yet) and try and reverse-engineer that specific file.

robertlipe:
[...]

PR #762, opened 2021-11-13
Add support for Locations.fit files

fabzzap:
See issue #719 

6 out of 9 fields are supported. The others are unknown, except the location icon field, which is not especially interesting

### DIFF
--- a/garmin_fit.cc
+++ b/garmin_fit.cc
@@ -34,6 +34,7 @@
 #include <QDateTime>           // for QDateTime
 #include <QFileInfo>           // for QFileInfo
 #include <QLatin1Char>         // for QLatin1Char
+#include <QMetaType>           // for QMetaType, QMetaType::UInt
 #include <QString>             // for QString
 #include <Qt>                  // for CaseInsensitive
 #include <QtGlobal>            // for qint64
@@ -214,7 +215,7 @@ QString
 GarminFitFormat::fit_getstring(int size)
 {
   if (fit_data.len < size) {
-      throw ReaderException("record truncated: expecting " + std::to_string(size) + " bytes, but only got " + std::to_string(fit_data.len) + ".");
+    throw ReaderException("record truncated: expecting " + std::to_string(size) + " bytes, but only got " + std::to_string(fit_data.len) + ".");
   }
   QByteArray buf(size + 1, 0);
   gbsize_t count = gbfread(buf.data(), size, 1, fin);
@@ -313,7 +314,7 @@ GarminFitFormat::fit_parse_definition_message(uint8_t header)
   fit_data.message_def.insert(local_id, def);
 }
 
-uint32_t
+QVariant
 GarminFitFormat::fit_read_field(const fit_field_t& f)
 {
   /* https://forums.garmin.com/showthread.php?223645-Vivoactive-problems-plus-suggestions-for-future-firmwares&p=610929#post610929
@@ -345,6 +346,9 @@ GarminFitFormat::fit_read_field(const fit_field_t& f)
       }
       return -1;
     }
+  case 0x7:
+    return fit_getstring(f.size);
+
   case 0x83: // sint16
   case 0x84: // uint16
     if (f.size == 2) {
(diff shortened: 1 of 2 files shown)


garmin_fit.cc
```diff
@@ -655,6 +730,21 @@ GarminFitFormat::fit_parse_data(const fit_message_def& def, int time_offset)
...
+    }
+    if (global_opts.debug_level >= 7) {
+      debug_print(7,"%s: storing fit data location %d\n", MYNAME, def.global_id);
+    }
+    auto* lappt = new Waypoint;
+    lappt->latitude = GPS_Math_Semi_To_Deg(lat);
+    lappt->longitude = GPS_Math_Semi_To_Deg(lon);
+    lappt->shortname = name;
+    lappt->description = description;
+    waypt_add(lappt);
```
tsteven4: This doesn't have anything to do with a lap, does it?  How about using locpt instead of lappt?

it seems you may have altitude data from kFieldLocAltitude you aren't setting.

garmin_fit.cc
```diff
@@ -395,198 +412,256 @@ GarminFitFormat::fit_parse_data(const fit_message_def& def, int time_offset)
...
-    } else {
+    if (f.type == 7) {
+      QString val = fit_getstring(f.size);
       switch (def.global_id) {
-      case kIdDeviceSettings: // device settings message
+      case kIdLocations:
         switch (f.id) {
-        case kFieldGlobalUtcOffset:
+        case kFieldLocationName:
           if (global_opts.debug_level >= 7) {
```
tsteven4: Perhaps it is better to teach fit_read_field to handle f.type 7 data and return a QVariant.  Then the handling of kIdLocations can be handled in the def.global_id switch block kIdLocations case below along with latitude, longitude, altitude.

garmin_fit.cc
```diff
@@ -210,6 +210,21 @@ GarminFitFormat::fit_getuint32()
   }
 }
 
+QString
+GarminFitFormat::fit_getstring(int size)
+{
+  if (fit_data.len < size) {
+      throw ReaderException("record truncated: expecting " + std::to_string(size) + " bytes, but only got " + std::to_string(fit_data.len) + ".");
```
tsteven4: please correct the indentation here.
Also please delete the trailing white space on two other lines.

This can be done manually or with "astyle --options=astylerc garmin_fit.cc"

tsteven4 (changes requested):
Thanks for taking this on.

Please add a test case.  See testo.d/garmin_fit.test.  We would prefer a fit file that is sufficient to test your changes but isn't too large.  Without a test case maintenance is problematic and things tend to get broken with future changes.

tsteven4:
@fabzzap  Thanks for this.  Other users have complained of the same shortcomings in our fit handling, but you are the first to come up with a fix.
 
Can you try the attached patch to your commit?https://github.com/GPSBabel/gpsbabel/pull/762/commits/b6ab6ca9b2ef7ae38395c2b3160397344bb36c48

This resolves all the review issues and a few more with debug messages(qPrintable, appropriate message), except for a test case.  Hopefully you can supply a test case.
[pr762_b6ab6ca9b2e.patch.txt](https://github.com/GPSBabel/gpsbabel/files/7550484/pr762_b6ab6ca9b2e.patch.txt)

I will squash this PR when I merge, so you can just add commits with doing a forced push.

Thanks again.

fabzzap:
> @fabzzap Thanks for this. Other users have complained of the same shortcomings in our fit handling, but you are the first to come up with a fix.
> 
> Can you try the attached patch to your commit?[b6ab6ca](https://github.com/GPSBabel/gpsbabel/commit/b6ab6ca9b2ef7ae38395c2b3160397344bb36c48)
> 
> This resolves all the review issues and a few more with debug messages(qPrintable, appropriate message), except for a test case. Hopefully you can supply a test case. [pr762_b6ab6ca9b2e.patch.txt](https://github.com/GPSBabel/gpsbabel/files/7550484/pr762_b6ab6ca9b2e.patch.txt)
> 
> I will squash this PR when I merge, so you can just add commits with doing a forced push.
> 
> Thanks again.

I applied the patch to my tree. It seems to work well, Thank you

tsteven4:
@fabzzap This looks great.  Can you supply a fit file that includes the kIdLocations entries, including kFieldLocationName and/or kFieldLocationDescription ?  I would like to add this to our regression test suite.  If so can you also tell me the model of the device that generated the fit file?

Thanks

tsteven4:
Another user contributed a fit file for regression test.  I will add a test after this PR merges.  Thanks for fixing this issue.

fabzzap:
Anyway, I pushed a reference Locations.fit (taken from a Garmin Edge Explore, manually edited to remove sensitive locations), and a test case to this branch

tsteven4:
Wow, your commit just missed the merge.  I understand the file can contain sensitive information.  Your file gives slightly more coverage (description), but it is nice to have fit files straight from the device in case we discover oddities in the future.  I think we are good with the file from the other user.  Thanks for your efforts.

tsteven4:
This fixed #719

PR #763, opened 2021-11-16
Introduce resampling filter and nvectors.

tsteven4:
This implements classic digital signal processing resampling techniques.  It can be used to change a track sample rate by a rational factor, e.g. twice, half, two thirds.  It uses nvectors to average position, so it works across the antimeridian and at the poles.  The track points are assumed to be sampled at a constant rate.  The track points do not need to have timestamps.  It can also be used to smooth a track without changing the sample rate.

It can be used to create similar output to the existing interpolate filter, but for points that are very far apart the nvector averaging is somewhat distorted from great circle calculations.  Also, the propagation of unknown altitude values is worse than the existing interpolate filter.

The version of nvectors/vector3d included here incorporates the improvements suggested in #761 and a few more.

The ability to decimate data was desired some time ago by someone working with very large data sets.  This is an enhanced version of code he tested.  As I recall he was experiencing run times with the simplify filter in excess of a day.  In such a case it may be desirable to either use the resample filter to decimate the data and then the simplify filter, or use the resample filter alone to decimate and smooth the data.


FJBDev:
@tsteven4 
Thanks for this superb feature that I was just looking for.
Would it be possible to have the ability to specify the number of points ? 

i.e.: My GPX file has 200 points and I'd like to have 956 points in the end. I used a factor of 5 to be closer to 956 but it would be great if I could match exactly a specific number.

tsteven4:
you can change the sample rate by rational factor by using both interpolate and decimate, e.g.

$ ./gpsbabel -t -i random,points=200 -f dummy  -x validate,debug -x resample,interpolate=24,decimate=5,average=24 -x validate,debug

Processing waypts
point ct: 0, waypt_count: 0

Processing routes
route head ct: 0, route_count: 0
total route point ct: 0, route_waypt_count: 0

Processing tracks
track 0 ct: 200, waypt_count: 200
track head ct: 1, track_count: 1
total track point ct: 200, track_waypt_count: 200

Processing waypts
point ct: 0, waypt_count: 0

Processing routes
route head ct: 0, route_count: 0
total route point ct: 0, route_waypt_count: 0

Processing tracks
track 0 ct: 956, waypt_count: 956
track head ct: 1, track_count: 1
total track point ct: 956, track_waypt_count: 956

FJBDev:
Thanks, not very intuitive at first (how do you end up with 956 with those parameters?) I'll try that.

tsteven4:
agreed.  I started with finding a rational factor that would change the sample rate by 956/200 = 239/50.  But while this changed the sample rate as expected, it yields 952 points, so I tried 240/50, which gave me 956 points.  I reduced 240/50 = 24/5.

But we can find a general solution.  If you start with n points and you want to end up with m points covering the same span of time then  (n-1)*Ts = (m-1)*Ts' where Ts is the original sample rate and Ts' is the modified sample rate.  So in your case n=200 and m = 956.  Ts/Ts' = (m-1)/(n-1) = 955/199.  199 is prime so we can't reduce it further.
$ ./gpsbabel -t -i random,points=200 -f dummy  -x validate,debug -x resample,interpolate=955,average=955,decimate=199 -x validate,debug

Processing waypts
point ct: 0, waypt_count: 0

Processing routes
route head ct: 0, route_count: 0
total route point ct: 0, route_waypt_count: 0

Processing tracks
track 0 ct: 200, waypt_count: 200
track head ct: 1, track_count: 1
total track point ct: 200, track_waypt_count: 200

Processing waypts
point ct: 0, waypt_count: 0

Processing routes
route head ct: 0, route_count: 0
total route point ct: 0, route_waypt_count: 0

Processing tracks
track 0 ct: 956, waypt_count: 956
track head ct: 1, track_count: 1
total track point ct: 956, track_waypt_count: 956

FJBDev:
Actually, my goal is to use timeless GPX files, would it be possible in this case to specify any `m` numbers of endpoints ? or do you need time data for resampling ?

tsteven4:
Resample doesn't use timestamps at all.  Normally this kind of signal
processing is applied to data that is sampled at a constant rate, but the
filter doesn't actual use any time information from the points.  My last
answer, (m-1)/(n-1), will result in the number of points you asked for.

PR #764, opened 2021-11-17
Eliminiate redundant route_head data member

tsteven4:


tsteven4:
The action here is in defs.h and route.cc.

robertlipe:
I like it. I meant to to this years ago when we moved to 'real' containers.
Thanx.

PR #765, opened 2021-11-18
add test for fit Locations(Lctns) files.

tsteven4:


tsteven4:
This illustrates the value of regression test cases and Continuous Integration.  We found a sensitivity to a change in behavior from Qt5 to Qt6 in the constructor [QString::QString(const QByteArray &ba)](https://doc-snapshots.qt.io/qt6-dev/qstring.html#QString-9).  I wouldn't be surprised if this bites us again somewhere else.

Thanks again to @fabzzap and Osvald for contributing to the resolution of this issue.

PR #771, opened 2021-11-26
implement universal solution for kml viewpoint

tsteven4:
The existing bounding box used to create the kml viewpoints has problems at the poles and antimeridian.  It aslo has problems at whenever the bounding box has both positive and negative longitudes, even around the prime meridian.

The intent is to solve these issues in a universal way, that works for all sets of points on the earth.  Specifically, it works for points around the antimeridian and points around the poles.  We also fix the bug around the prime meridian.

The proposed solution uses Welzl's algorithm to find the minimum bounding "circle" (actually a spherical cap) of the points.  It implements this in n-vector space.  Welzl's algorithm is O(n), so this solution should scale.   I beleive the combination of the Welzl algorithm with n-vector representation of position is novel.

At present there are some reproducibility issues due to randomization of the order of the points.  This appears to introduce very small errors in some of the reference files.


tsteven4:
At the moment regression passes except for 64 bit windows builds.  They fail with a stack overflow.

More critically, the expected linear performance has not been realized.

tsteven4:
performance appears linear if the set of points is in a hemisphere.  However, if the set of points does not fit in any half sphere it becomes very non-linear.

memory utilization also seems to be an issue somewhere above 10,000 points.

PR #773, opened 2021-11-28
migrate arcdist, polygon and xcsv style file reading to gpsbabel::TextStream

tsteven4:
fix lgtm detected ill-defined for loop.

remove support for non standard entity replacements, which has been disabled since 2005.


tsteven4:
I am going to repackage this and resubmit.

PR #775, opened 2021-11-30
wean mkshort from cet

tsteven4:
clean utf-8 strings by removing QChar::ReplacementChar

break utf-8 strings at grapheme boundaries, not unicode code point boundaries.

remove the unused cet functions that are unused as a result of the above.


tsteven4:
repackaged in #777

robertlipe:
Good. Thanx.

PR #777, opened 2021-12-01
retire cet

tsteven4:
in mkshort:
when cleaning a string we convert to utf-16, remove any replacement characters, and convert back to utf-8.
when truncating a string we split at grapheme boundaries instead of unicode code point boundaries.

in gbfile:
gbfgetucs2str is renamed gbfgetutf16str and reworked to use QChar and QString.  It now supports line endings of \r\n, ~~\r,~~ or \n, and detects errors when:

- the file ends with an incomplete utf-16 character
- the file ends with an incomplete surrogate pair
- a high surrogate is not followd by a low surrogate
- a low surrogate isn't preceeded by a high surrogate

replacement character handling, splitting at grapheme boundaries, and the various flavors of line endings are covered in new tests.


robertlipe:
Hooray!

PR #781, opened 2021-12-02
hopefully the real fix for garmin_xt tool issues.

tsteven4:
Fix LGTM detected "Comparison between i of type uint8_t and Count of
wider type int."

This latest find by LGTM is one in a long series of issue detections by
various tools.  However, close analysis reveals the previous solutions
to these deteced issues didn't find or fix the real root issues:

commit 233f3c8b0bf69397403b6c1f29af8e10a65a8928, 10/22/2013, Pad
internal buffer to appease -fsanitize=address.  While the fix appeased
the sanitizer, it didn't fix the root problem.  The root problem was
a bug in the original translation of format_garmin_xt_decrypt_trk_blk
when converting to zero-based array indexing.

commit dcf0dd85a71c6fa5fc3dac72520c2070ba051108, 7/8/2015, A bunch of
busy work to satisfy hyperactive warnings in newer GCC builds.  Again,
the fix didn't fix the root problem.  The root problem was a bug in the
original translation of format_garmin_xt_proc_strk.  This bug resulted
in a mismatch between our test output and that of the original project.
In our output the timestamp of the last two points in a track was
identical.

This fix results in our output matching the reference file of the original
project (with allowances for precision).

### DIFF
--- a/garmin_xt.cc
+++ b/garmin_xt.cc
@@ -153,9 +153,9 @@ format_garmin_xt_decrypt_trk_blk(int Count, uint8_t TrackBlock[])
 {
   int j = 12;
   while (j<(Count-1)) {
-    for (uint8_t i = j; i < Count; i++) {
+    for (int i = j; i < Count; i++) {
       TrackBlock[i] = TrackBlock[i] >> 1;
-      if (i<(Count)) {
+      if (i<(Count-1)) {
         TrackBlock[i] = TrackBlock[i] + (TrackBlock[i+1] % 2) * 128;
       }
     }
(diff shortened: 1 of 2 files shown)


tsteven4:
I note that with this fix the first track in the saved track reference file garmin_xt_strk.gpx matches the active track reference file garmin_xt_atrk.gpx, except for the latitude of the first point.  This single mismatch is also present in the original project.  Previously the times did not match for all but the last point.  The atrk and strk read code is separate.  The near alignment of the active and saved tracks supports the changes in this PR.  The disagreement about the latitude of the first point is unexplained, and not effected by this PR.

tsteven4:
The mismatch in the first latitude between the active and saved tracks appears to be because the latitude is stored at higher resolution in the active track file compared to the saved track file.

PR #782, opened 2021-12-05
harden gbfgetcstr, gbfgetpstr.

tsteven4:
Hangs were detected with afl and the gdb reader and gbfgetcstr.  One such
hang reads a fuzzed 32 bit url count with FREAD_i32, and then proceeds
to fetch that many strings with FREAD_CSTR_AS_QSTR.  This results in
fetches from beyond EOF that are undetected.  gbfgetcstr_old is modified
to throw a fatal error if reading is attempted beyond EOF.

A similar failure was not detected with gbfgetpstr, but it is also
modified to throw a fatal error if reading is attempted beyond EOF.

### DIFF
--- a/gbfile.cc
+++ b/gbfile.cc
@@ -990,8 +990,12 @@ gbfgetcstr_old(gbfile* file)
   for (;;) {
     int c = gbfgetc(file);
 
-    if ((c == 0) || (c == EOF)) {
+    if (c == 0) {
       break;
+    }
+
+    if (c == EOF) {
+      fatal("%s: Unexpected end of file (%s)!\n", file->module, file->name);
     }
 
     if (len == file->buffsz) {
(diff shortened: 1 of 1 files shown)


GPSBabelDeveloper:
LGTM

We may have only Mapsend using the Pascal String stuff and it's on a
chopping block on my laptop awaiting a different evening to take it (and a
several others) out.

Still, this change is good taste until we can confidently remove these, so
thank you.

tsteven4:
Usage is:
grep gbfgetpstr *cc | cut -f1,1 -d:| sort | uniq -c
      5 easygps.cc
      2 gbfile.cc
      6 mapsend.cc
      2 tpg.cc
grep gbfgetcstr *cc | cut -f1,1 -d:| sort | uniq -c
      3 easygps.cc
      5 gbfile.cc
      4 gdb.cc
      2 teletype.cc

PR #783, opened 2021-12-05
Kill CarteSurTable format.

tsteven4:
Apparently this format hasn't ever been used.  Our sample has coordinates in UTM,
but we have been interpreting them as degrees/100000.  "Le pic du midi"
referred to in the sample is The Pic du Midi de Bigorre.


tsteven4:
There are lots of cartesurtable samples at http://www.varouler.com/?q=vrgps

robertlipe:
Thank you. Nuke it!

tsteven4:
The additional samples at the site above use a different SRS, "GRID 3 (Lambert II étendu)".  Our sample uses "GRID 0 (UTM/UPS)".  I haven't found a sample that uses lat/lon in units of1e-5 degrees like our code, but I suppose its possible some other  "GRID ..." does.

The CarteSurTable website went dark in 2018, but is archived at https://web.archive.org/web/20180826191216/http://phgiraud.free.fr/CarteSurTable/index.html.  The last update was in 2013.  Windows 7 was the last supported version.

PR #785, opened 2021-12-08
Remove compegps, cst, g7towin, gopal, gpsman, gpsutil, jogmap, jtr, maggeo, mapsend, nm4, pcx, pocketfms, skyforce, stmsdf, stmwpp, tiger, vpl, vidaone, vitosmt, vitovtt, wfff, and yahoo.

GPSBabelDeveloper:
Merge reaper1201 into trunk.

Turn down several formats with no maintenance commits or evidence of users in many years. Those PocketPC-era formats have been a maintenance burden.


tsteven4:
git mv yahoo.h deprecated

tsteven4:
I don't see any need to maintain
DEPRECATED_FMTS
DEPRECATED_HEADERS
DEPRECATED_SHAPE
in GPSBabel.pro  and CMakeLists.txt.  I would vote to just delete those variables from the files.

tsteven4:
I can use nmea to get the gbfile coverage back into gbfile.test that we lost with the deprecation of the pcx format.

tsteven4:
please merge the current master so the changes are reasonable (not 227 files!)
git checkout master
git pull
git checkout reaper1201
git merge master

tsteven4:
the gbfile test is updated in #786 to avoid the pcx format.

GPSBabelDeveloper:
I've been battling Git tonight to get around that 277 file mess - and to understand how I keep doing this to myself.

I'll take out the MAXIMAL/MINIMAL/CSV stuff after landing this. I haven't heard from @viettaml and that's getting pretty obnoxious for us. If it's needed to keep that distinction, we can move to instantiation-based ctors that makes features appear only when certain modules are present.  The days of kilobytes mattering are long gone once we added -lqtcore. :-)

I moved the gbfile compression test to NMEA (somewhere in those 277...) so this seems pretty natural.

Thanx,
RJL

tsteven4:
Thanks for the merge.  I found the following issues which we can fix in this PR or a subsequent one.

1. vidaone was removed from GPSBabel.pro, CMakeLists.txt, the serialization reference files and testo.  The source and documentation files are still present.  Was the intent to retire vidaone?
2. jogmap was removed from GPSBabel.pro, CMakeLists.txt, the serialization reference files and testo.  The source and documentation files are still present.  Was the intent to retire jogmap?
3. yahoo was removed, except for yahoo.h which needs to be moved with git to deprecated.
4. gpsman.test was modified.  The original file should  just be moved to testo.d/deprecated.
5. I have gbfile.test modified to replace the coverage for the supplemental plane due to the retirement of pcx in #786 

Please use the "Squash and merge" button to merge this PR instead of the "Create a merge commit" button.  This will result in the master branch history just showing one commit for this PR instead of the 13 commits in this branch.

tsteven4:
vitosmt is designated as a pocket pc format in our doc.  If we are killing pocket pc related things it might be another one to consider.

tsteven4:
jeeps/gpsutil.h should stay in GPSBabel.pro and CMakeLists.txt

GPSBabelDeveloper:
Fixed in c1d78d15..ceb9ea92
(And probably some of the earlier builds. My dumb grep for 'gpsutil' (the
format) took these out. Having a format gpsutil and a file inside jeeps has
long been a source of pain, so I'm glad to see these go. My brute force
tooling has struggled with it, though.

GPSBabelDeveloper:
Lessons learned (OK, I already knew some of these, but this reinforced it) in this CL: juggling this many formats at once is too hard. Don't do this.
- "Not all things worth automating are worth automating well". We have too many fiddly bits that look similar in too many places and many of them have obscure interactions. (e.g. main doc failing because an introduction chapter links to the format you just resolved) tools/nuke_format is as dumb as a rock, but it should make dropping a format possible to do in a minute and once per commit so that a revert should be possible, which should make our (somewhat overkill) deprecated/ not as tempting. Maybe we should leave deprecated/ to version control.
- Changing the first line of the pull request breaks and restarts the thread in gmail. Not a bug, but it can make waiting for results frustrating when they show up in a different place.

kermit-the-frog:
What's the reason for removing the support for CompeGPS?

GPSBabelDeveloper:
As given in the mail list on recent such changes, reasons included:

Low usage count.
Inactive community presence (blogs, inclusion in web pages, etc.)
https://trends.google.com/trends/explore?date=all&q=compegps
No active maintainer since 2008.
Excessive maintenance costs.
Log history found that only mechanical edits had been performed on it in
years showing that we were spending time "fixing" it to make/keep the code
contemporary, but reflecting no user contributions or results from user
problem reports.
This was one of many of one former maintainer's file formats that came out
of nowhere without any users actually requesting it.
https://github.com/GPSBabel/gpsbabel/commit/5b7c23392d75ac1c47f920b15faeee2333e83de8

For eleven years, we actually encouraged people to quit using this format.
The doc page used to say, "Release 6.1 of CompeGPS Land and CompeGPS Air
added GPX as an import export format for waypoints, routes, and tracks. As
GPX is a more heavily used format, you should probably use that for new
applications."

We're a 20 year old program. Over the years we collected a lot of formats
that nobody (or statistically close to "nobody") was still using and we
have been increasingly hampered by development because of the cost of these
old formats.  We've tossed out iPaq-era code only fairly recently. We can't
support everything forever.

The good news for you is that we're open source. You're welcome to keep
using and maintaining the old version if you actually did use it, unlike
when, say, Adobe removes your favorite feature. If you want to roll back
the deletion and keep maintaining it against the trunk (i.e. you pay the
maintenance cost that we won't) you're even free to do that.

RJL

PR #787, opened 2021-12-10
organize source files in GPSBabel.pro, CMakeLists.text`

tsteven4:
The files are listed one per line in alphabetical order.  This makes it much easier to compare the two.  It also makes it easier to see what happened should the reaper return.


robertlipe (approved):
Yes!, Please!
I've needed to do that for years.
If you have time, please change the remove_word to remove_line calla in tools/nuke_format
If not, that's still fine.
Thanks!

tsteven4:
> If you have time, please change the remove_word to remove_line calla in tools/nuke_format

I made a few changes there, but that script is far from portable.   brew has shellcheck which I have found useful.  ubuntu linux has sed which is GNU sed, but not named gsed.  mystyle.sh has been tortured trying to find sed.  At least the script should need to be run by anyone else.

I just deleted the msvc project and solution files.  I haven't been keeping these up to date.  It's best just to generate the project files with "qmake -tp vc GPSBabel.pro".  The solution file will be created by MSVC from that.  The need to create MSVC project files from GPSBabel.pro does constrain us a bit in what we can do in qmake, but it's the best solution I have found.

PR #790, opened 2021-12-12
retire some csv char based routines.

tsteven4:
The last csv_lineparse user, garmin_txt, is converted to csv_linesplit.

xcsv_parse_val converted to use csv_stringtrim(QString,QString,int),
eliminating the other overloads.

xcsv_parse_style_line parsing of FIELD_DELIMITER, FIELD_ENCLOSER,
RECORD_DELIMITER and BADCHARS updated to use the above overload of
csv_stringtrim.  FIELD_DELIMITER, FIELD_ENCLOSER and RECORD_DELIMITER
changed to trim, with a double quote enclosure,  before character substitution.
There was a subtle bug in csvs_stringtrim(char, char, int).  When trimming
a source that was all white space the first white space character would
be retained, while others would be trimmed.  This did occur with CRNEWLINE,
but the bug allowed other substitutions to work, e.g. SPACE, NEWLINE, TAB, CR.

A small bug in csv_stringclean was fixed.  If the to_nuke was empty an invalid
regular expression was created.

### DIFF
--- a/csv_util.cc
+++ b/csv_util.cc
@@ -52,80 +52,15 @@ QString
 csv_stringclean(const QString& source, const QString& to_nuke)
 {
   QString r = source;
-  // avoid problematic regular rexpressions, e.g. xmapwpt generated [:\n:],
-  // or one can imagine [0-9] when we meant the characters, '0', '-', and '9',
-  // or one can imagine [^a] when we meant the characters '^' and 'a'.
-  QRegularExpression regex = QRegularExpression(QString("[%1]").arg(QRegularExpression::escape(to_nuke)));
-  assert(regex.isValid());
-  return r.remove(regex);
-}
-
-// csv_stringtrim() - trim whitespace and leading and trailing
-//                    enclosures (quotes)
-//                    returns a copy of the modified string
-//    usage: p = csv_stringtrim(string, "\"", 0)
-char*
-csv_stringtrim(const char* string, const char* enclosure, int strip_max)
-{
-  static const char* p1 = nullptr;
-  char* tmp = xstrdup(string);
-  size_t elen;
-
-  if (!strlen(string)) {
-    return (tmp);
+  if (!to_nuke.isEmpty()) {
+    // avoid problematic regular rexpressions, e.g. xmapwpt generated [:\n:],
+    // or one can imagine [0-9] when we meant the characters, '0', '-', and '9',
+    // or one can imagine [^a] when we meant the characters '^' and 'a'.
+    QRegularExpression regex = QRegularExpression(QString("[%1]").arg(QRegularExpression::escape(to_nuke)));
+    assert(regex.isValid());
+    r.remove(regex);
   }
-
-  if (!enclosure) {
-    elen = 0;
-  } else {
-    elen = strlen(enclosure);
-  }
-
-  char* p2 = tmp + strlen(tmp) - 1;
-  p1 = tmp;
-
-  /* trim off trailing whitespace */
-  while ((p2 > p1) && isspace(*p2)) {
-    p2--;
-  }
-
-  /* advance p1 past any leading whitespace */
-  while ((p1 < p2) && (isspace(*p1))) {
-    p1++;
-  }
-
-  /* if no maximum strippage, assign a reasonable value to max */
-  strip_max = strip_max ? strip_max : 9999;
-
-  /* if we have enclosures, skip past them in pairs */
-  if (elen) {
-    int stripped = 0;
-    while (
-      (stripped < strip_max) &&
-      ((size_t)(p2 - p1 + 1) >= (elen * 2)) &&
-      (strncmp(p1, enclosure, elen) == 0) &&
-      (strncmp((p2 - elen + 1), enclosure, elen) == 0)) {
-      p2 -= elen;
-      p1 += elen;
-      stripped++;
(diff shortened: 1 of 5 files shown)


garmin_txt.cc
```diff
@@ -1049,11 +1049,13 @@ bind_fields(const header_type ht)
...
 static void
-parse_grid()
+parse_grid(const QStringList& lineparts)
 {
-  char* str = csv_lineparse(nullptr, "\t", "", 1);
-
-  if (str != nullptr) {
+  if (lineparts.size() < 1) {
+    fatal(MYNAME ": Missing grid headline!\n");
+  } else {
```
GPSBabelDeveloper, line 1054: fatal should be attribute __noreturn__ so you can just drop the else and lower the quote level on everything else.

csv_util.cc
```diff
@@ -212,121 +147,6 @@ csv_dequote(const QString& string, const QString& enclosure)
...
-/*****************************************************************************/
-/* csv_lineparse() - extract data fields from a delimited string. designed   */
-/*                   to handle quoted and delimited data within quotes.      */
-/*                   returns temporary COPY of delimited data field (use it  */
-/*                   or lose it on the next call).                           */
-/*    usage: p = csv_lineparse(string, ",", "\"", line)  [initial call]      */
-/*           p = csv_lineparse(NULL, ",", "\"", line)    [subsequent calls]  */
-/*****************************************************************************/
-char*
-csv_lineparse(const char* stringstart, const char* delimited_by,
```
GPSBabelDeveloper, line 224: This is great news. Alex's work in this area took a while to become battle-hardened but it's certainly been problematic in C++ vs. C. It's served us well, but thank you for taking this behind the barn and ending it.

garmin_txt.cc
```diff
@@ -908,22 +911,19 @@ parse_date_and_time(char* str, time_t* value)
 }
 
 static uint16_t
-parse_categories(const char* str)
+parse_categories(const QString& str)
```
GPSBabelDeveloper, line 914: I was going to worry about the static cbuf here, but I found a bigger worry instead:
Can you find any callers of this functino?

garmin_txt.cc
```diff
@@ -943,16 +943,16 @@ parse_categories(const char* str)
 }
 
 static int
```
GPSBabelDeveloper: Please make this a proper bool.

garmin_txt.cc
```diff
@@ -966,20 +966,20 @@ parse_temperature(const char* str, double* temperature)
     }
     return 1;
   } else {
-    fatal(MYNAME ": Invalid temperature \"%s\" at line %d!\n", str, current_line);
+    fatal(MYNAME ": Invalid temperature \"%s\" at line %d!\n", qPrintable(str), current_line);
```
GPSBabelDeveloper, line 957: I was going to nudge this toward FataMsg() but then I noticed 51 other occurrences of this. That's a good search-and-replace project for a budding Babel-Head to undertake...
Don't worry about it, tsteven4.

garmin_txt.cc
```diff
@@ -1063,27 +1065,23 @@ parse_grid()
...
+parse_datum(const QStringList& lineparts)
 {
-  char* str = csv_lineparse(nullptr, "\t", "", 1);
-
-  if (str != nullptr) {
-    datum_index = gt_lookup_datum_index(str, MYNAME);
-  } else {
+  if (lineparts.size() < 1) {
     fatal(MYNAME ": Missing GPS datum headline!\n");
+  } else {
```
GPSBabelDeveloper: Drop the else?

garmin_txt.cc
```diff
@@ -1304,7 +1311,7 @@ parse_track_waypoint()
       }
       break;
     case 9:
-      WAYPT_SET(wpt, course, atoi(str));
+      WAYPT_SET(wpt, course, atoi(CSTR(str)));
```
GPSBabelDeveloper, line 1302: Is this str.toDouble() ?

GPSBabelDeveloper (commented):
This is much more relaxing to read. Thanx.
Most review comments are optional before submit.

garmin_txt.cc
tsteven4: usage at L1179.

garmin_txt.cc
tsteven4: done

garmin_txt.cc
tsteven4: The converstion to QString::toTYPE requires knowledge that the value being converted does not have a trailing non-convertable part.  atoTYPE will ignore trailing non-convertable data while QString::toTYPE will not.

garmin_txt.cc
tsteven4: I went ahead and did that.  The lateset cppcheck doesn't always figure out [[noreturn]], many of the remaing warnings are related to that cppcheck issue.  It is also hard for tools to figure out that if is_fatal(condition, ...) returns it implies condition == false.

garmin_txt.cc
tsteven4: ok, I moved the QString/char array a bit further along.

garmin_txt.cc
tsteven4: ok.

PR #791, opened 2021-12-15
Fix warnings on Mac related to printing qint64s

robertlipe:
Fix default warnings on Mac by printing qint64s as %lld, but portably.

### DIFF
--- a/garmin_fit.cc
+++ b/garmin_fit.cc
@@ -25,6 +25,7 @@
 #include <cstdint>             // for uint8_t, uint16_t, uint32_t, int32_t, int8_t, uint64_t
 #include <cstdio>              // for EOF, SEEK_SET, snprintf
 #include <deque>               // for deque, _Deque_iterator, operator!=
+#include <inttypes.h>          // for PRItNN
 #include <memory>              // for allocator_traits<>::value_type
 #include <string>              // for operator+, to_string, char_traits
 #include <utility>             // for pair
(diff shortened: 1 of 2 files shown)


lowranceusr.cc
```diff
@@ -139,7 +140,7 @@ LowranceusrFormat::register_waypt(const Waypoint* wpt) const
   }
 
   if (global_opts.debug_level >= 2) {
-    printf(MYNAME " adding waypt %s (%s) to table at index %d\n",
+    printf(MYNAME " adding waypt %s (%s) to table at index %" PRId64 "\n",
```
GPSBabelDeveloper, line 143: This might be a problem in Qt on MacOS. The warning I'm trying to fix was:
lowranceusr.cc:144:70: warning: format specifies type 'int' but the argument has type 'qsizetype' (aka 'long long') [-Wformat]
but if we look at the doc at https://doc.qt.io/qt-5/qlist.html#length we see they really are just plain ints.

Maybe we just cast these to plain ole ints and accept that we're going to shorten what we're printing, but that's OK because if we had > 4.2bn (or the square of that for 64-bit) we're going to have way too much debug print action at best and run out of memory soemwhere else; it's not like you're going to pass a 4.2bn geojson file to a lowrance...

tsteven4: Be aware of the Qt 6 change "QList's (and hence QVector's) size type is changed from int to qsizetype."

https://doc.qt.io/qt-6/qtcore-changes-qt6.html#qvector-qlist

garmin_fit.cc
```diff
@@ -25,6 +25,7 @@
 #include <cstdint>             // for uint8_t, uint16_t, uint32_t, int32_t, int8_t, uint64_t
 #include <cstdio>              // for EOF, SEEK_SET, snprintf
 #include <deque>               // for deque, _Deque_iterator, operator!=
+#include <inttypes.h>          // for PRItNN
```
tsteven4, line 28: this should be <cinttypes> for c++, although the Qt5 -> Qt6 change from int -> qint64 needs to be dealt with

tsteven4:
For garmin fit I would prefer #795.  Yes, it was quite a bit more work.

GPSBabelDeveloper:
Agreed. That's much better. This change is dumb.  Thanx.

PR #793, opened 2021-12-15
Deprecate navicache

robertlipe:
Navicache was always a little used format. The one known user confirmed
he'd quit using it and passed away earlier this year.

A former owner of the site declared it was abandoned. It had been idle for
years before this declaration in 2016.

Background at
https://www.navicache.com/cgi-bin/ib312a/ikonboard.cgi?s=34cde18c7da3dbbc22ad6c7b3f7aef86;act=ST;f=23;t=1728#entry10053
and "it has been dead so long now not sure the site could ever be revived" just two posts lower.

It's dead, jim.


tools/nuke_format
```diff
@@ -31,12 +31,13 @@ remove_line $1.cc GPSBabel.pro
...
 
 remove_manually $1 reference/help.txt
+remove_manually $1 reference/format3.txt
 
 # Take out the boilerplate entries in vecs.h ...
 remove_line_containing extern.*$1 vecs.h
 remove_line_containing Format.*$1 vecs.h
 remove_line_containing include.*$1 vecs.h
-# ... then let a human nerd snip3 the actual table entry awy
+# ... then let a human nerd snipe the actual table entry away.
```
tsteven4: snipe -> snip

tsteven4 (approved):
This seems well thought out and the rationale is nicley documented in the PR.

tools/nuke_format
GPSBabelDeveloper: Fixed

PR #794, opened 2021-12-15
Possible unterminated string in humminbird

GPSBabelDeveloper:
Leave space for null terminator when copying string into fixed-size buffer.

(From 2021-03-20.)

### DIFF
--- a/humminbird.cc
+++ b/humminbird.cc
@@ -875,7 +875,7 @@ humminbird_rte_tail(const route_head* rte)
     be_write32(&humrte->time, humrte->time);
 
     QString name = mkshort(rtename_sh, rte->rte_name);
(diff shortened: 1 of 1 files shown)


humminbird.cc
```diff
@@ -875,7 +875,7 @@ humminbird_rte_tail(const route_head* rte)
     be_write32(&humrte->time, humrte->time);
 
     QString name = mkshort(rtename_sh, rte->rte_name);
-    strncpy(humrte->name, CSTR(name), sizeof(humrte->name));
+    strncpy(humrte->name, CSTR(name), sizeof(humrte->name)-1);
```
tsteven4, line 878: I can beleive this made some tool happy, but it is unclear to me if the format really uses a NULL terminator on a fixed length string, it may just be zero filled.  At least this change is consistent with L737.

PR #795, opened 2021-12-15
use streams for garmin_fit debug messages.

tsteven4:
This eliminates the Qt 6 warnings related to the change in types for container sizes and the format specifier in the garmin_fit format "warning: format ‘%d’ expects argument of type ‘int’, but argument 5 has type ‘qsizetype’ {aka ‘long long int’} [-Wformat=]"

Most of these were done with a couple vi commands, but ~15 were converted by hand.

Testing revealed a mismatch that was releated to 24 PVS studio warnings "Incorrect format. Consider checking the fourth actual argument of the 'debug_print' function. The SIGNED integer type argument is expected."  This resulted in the static_casts of val to int32_t in the stream.

If I add four Qt::uppercasedigits modifiers and modify garmin_fit.test to used -D9 the logs match.

debug_print is now unsued and could be eliminated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,6 +206,7 @@ set(SUPPORT
   rgbcolors.cc
   route.cc
   session.cc
+  src/core/logging.cc
   src/core/nvector.cc
   src/core/textstream.cc
   src/core/usasciicodec.cc
--- a/GPSBabel.pro
+++ b/GPSBabel.pro
@@ -207,6 +207,7 @@ SUPPORT = \
   rgbcolors.cc \
   route.cc \
   session.cc \
+  src/core/logging.cc \
   src/core/nvector.cc \
   src/core/textstream.cc \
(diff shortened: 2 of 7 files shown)


GPSBabelDeveloper (approved):
This is much better taste - and a bigger undertaking - than my ham-handed swat at this problem last night. Thanx for taking the high road.

Should we keep a constructed Debug() (and or Warning/Fatal)  in the format structure, have a method to setFormatName that calls MYNAME (and, really, that could be automated in vecs.h) and elininated all the MYNMAME token concatenation stuff?

tsteven4:
Yeah, we need to ditch MYNAME in formats and filters.  main could register the information or create the Debug,,  Warning, Fatal classes.  But that isn't going into this PR.

GPSBabelDeveloper:
Agreed on all counts. Just thinking out loud.

PR #797, opened 2021-12-16
retire is_fatal for better static analysis

tsteven4:
todays static analysis tools generally cannot  realize that if "is_fatal(condition, ...);" returns it means condition == false, but they can make this inference from "if (condition) fatal(...);"


GPSBabelDeveloper:
That's impressive sed mojo!
I always disliked is_fatal() but Olaf was fond of it. Thanx for removing an eyesore.
Don't worry about the dtor warning. I need to just disable that in lgtm settings.

PR #798, opened 2021-12-16
remove misleading, dangerous and unneeded redefinitions for snprintf, vsnprintf

tsteven4:
> Beginning with the UCRT in Visual Studio 2015 and Windows 10, snprintf is no longer identical to _snprintf. The snprintf function behavior is now C99 standard conformant.

> Beginning with the UCRT in Visual Studio 2015 and Windows 10, vsnprintf is no longer identical to _vsnprintf. The vsnprintf function conforms to the C99 standard; _vnsprintf is kept for backward compatibility with older Visual Studio code.

delete an unneeded MSVC compiler pragma.


GPSBabelDeveloper (approved):
Compilers are less of a wild west than in 2002. :-)

tsteven4:
> Compilers are less of a wild west than in 2002. :-)

yeah, it only took a decade and a half.

PR #802, opened 2021-12-26
migrate from directory level cmake commands to target commands.

tsteven4:
fix a few bugs with our CMakeLists:

set WIN32 for gui target.  remove _CONSOLE, it's obsolete.

when undefining UNICODE, also undefine _UNICODE.  it's not clear to
me we have a sensitivity to UNICODE/_UNICODE.

take optimization flags out of CMakeLists, rely on CMAKE_BUILD_TYPE
to select appropriate flags.

don't force ZLIB_INHIBITED to be undefined on windows.

add cmake support for some of our configuration options:
disable-mappreview, embed_map, embed_translations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,4 @@
-
-# It isn't clear what the minimum required version is.
-# tested against 3.10.2 on ubuntu bionic 2018/07/05
-# tested against 3.5.1 on ubuntu xenial 2018/07/05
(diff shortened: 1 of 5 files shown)


robertlipe:
I think UNICODE comes from our Expat days. When Earth was shipped over a
modem, it made sense for us to use their libexpat.dll which was confiled to
write 16-bit "characters" for everything internally.

I could be wrong.

PR #803, opened 2021-12-27
make internal_styles.cc a generated file

tsteven4:
on all platforms.
This adds a build requirement for python3.


robertlipe:
Nice!

tsteven4:
While this worked everywhere I think I have a much better solution..

tsteven4:
Closed without merging in favor of #804

GPSBabelDeveloper:
Are the files linked into the executable it are they tucker inside an app
bundle? (I'm on mobile today...)

Maybe the day of copying/isolating one executable and expecting it to work
the same are over anyway.

tsteven4:
Regarding #804, the way we are doing in they are compiled into the executable much like internal_styles.cc was.  Note the gui was already using the Qt resource system for icons, translations and the gmapbase html file.

https://doc.qt.io/qt-5/resources.html:

> Currently, Qt always stores the data directly in the executable, even on Windows, macOS, and iOS, where the operating system provides native support for resources. This might change in a future Qt release.

https://doc.qt.io/qt-6/resources.html:

> Note: Currently, the Qt resource system does not make use of any system-specific capabilities for handling resources, such as the ones on Windows, macOS, and iOS. This might change in a future Qt release.

robertlipe:
Cool. I'm hip with that, then.
I think there's an expectation that a CLI is self-contained (perhaps
translations aside) where a GUI *expects* to be splattered out in the
filesystem. Given we dynamic link to Qt, that ship has already sailed. I
need to update my own mental model.
Thank you.

tsteven4:
If the Qt implemenation compicated installation I would have favored #803(closed).  As it stands #804(merged) simplifies build, makes minor simplifications to the code, and compiles in all resources like we have always done.  If Qt changes how resources are stored we may need to revisit this.

robertlipe:
Your reasoning is, as usual, solid.

I just needed to fill in my own knowledge gap here.

Thanks!

PR #805, opened 2021-12-29
fix bad merge in mtk_logger.

tsteven4:
1. Jan 01, 2017 269c55eb5 Robert fixes ftruncate issue.
2. Jan 23, 2017 604178aa8 HarelM fixes same a different way.
3. Jan 25, 2017 8456d1d21 Robert merges resulting in duplicate fixes.

I am backing out HarelM's fix which is effectively equivalent to what we have been using.

Screenshot 2021-12-29 092401
lM's fix.

### DIFF
--- a/mtk_logger.cc
+++ b/mtk_logger.cc
@@ -71,11 +71,6 @@
 
 #define MYNAME "mtk_logger"
 
-#ifdef __WIN32__
-#include <io.h>
-#define ftruncate _chsize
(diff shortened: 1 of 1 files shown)


robertlipe:
We didn't get fs::resize_file until c++17 and adding qt in here is even
worse. What a crazy situation.

Thanks for the cleanup.

PR #806, opened 2021-12-30
make gbversion.h, setup.iss generated files.

tsteven4:
This makes gbversion.h and setup.iss generated files for all build systems and platforms.  Furthermore, we generate gbversion.h for both the CLI and GUI builds.
This eliminates some issues when doing out of source builds were the wrong gbversion.h could be found.  We had parent path elements (../) embedded in various include directives, including windows rc files.  This is difficult because the child/parent relationship only is guaranteed to exist in the source tree, and may or may not exist in the build tree.  Since our build isn't hierarchical yet, there is no guaranteed relationship between the CLI build and the GUI build.  See gbversion.pri, gbversion.cmake for further explanation.  In contrast, today version is only defined in GPSBabel.pro, the GUI and cmake builds use the gbversion.h file the CLI builds, assuming the correct one can be found.  In an out of source build we often use the old one from the source tree instead of the newly generated one.
Unfortunately, this does require duplication of the version in two places: gbversion.pri for qmake builds, and qbversion.cmake for cmake builds.  There is no further duplication introduced to handle CLI and GUI builds which both rely on gbversion.h.  This duplication will be resolved when we drop support for qmake sometime in the future.  Also in the future, with a hierarchical build, we can define version once in the project command of the top-level CMakeLists.txt, and the sub-directory CMakeLists.txt will inherit those settings.

add cmake documentation targets.

We currently have some complexity to allow WEB (the location of the gpsbabel.org files, default ../babelweb) to be set on the make command line (assuming we are using a makefile generator).  Note this can also be set with qmake or cmake:
'qmake WEB=/tmp' or 'cmake -DWEB:PATH=/tmp'.
There is an explanation in CMakeLists.txt.  If we don't need to have the ability to set WEB on the make command line we can simplify the GPSBabel.pro and CMakeLists.txt a bit.


tsteven4:
There is now a new cmake build running on ubuntu, see https://github.com/GPSBabel/gpsbabel/runs/4681381558?check_suite_focus=true
We have been running a windows cmake build for some time.

Our biggest unimplemented feature with cmake is library configurability.

PR #807, opened 2022-01-02
use hierarchy with cmake.

tsteven4:
Ideally we would have a directory parallel to gui with the cli, but this allows us to do a hieraracal build with cmake without changing our directory structure.

It's unfortunate that we change the name of the GUI executable depeding on platform.  This changes the name of the target used to build the GUI.  We could name the target something like gui and then change the RUNTIME_OUTPUT_NAME property based on the platform.


robertlipe:
I'm open to addressing both of those fundamental issues.

We've danced around most of our interesting files being in our root
directory for years. Maybe we have a cmd/{formats,filters} as a sibling to
gui. I started that with src/core but never went back and addressed
existing code and new filters and formats kept flowing in.

Wanna change? I'm open to a new plan.

The renaming thing is something I never liked. We were working a problem
that may not even exist. There was some OS that couldn't cope with having
an executable and a launchable app of the same name. Maybe we could make
our main.cc invoke the GUI if we have no arguments or something.

I'd accept a breaking change that quit with the gpsbabelFE stuff.

Do we know any prior art for a program that has both a GUI and that
doesn't?  Apps that are vaguely like us (convert from imagemagick, objcopy
from binutils, ogrtoogr from GDAL) don't have GUIs. If you type "gpsbabel",
do you launch the CLI or the GUI?

We could make them both just one executable; the presence of a string of
arguments is a clear hint. The downside is that if you're running on a
headless Pi or a web server or something, you might not even be able to
depend on QtGui. Maybe we leave that behind.

Oh. There were issues in the pre-Qt translation files where we had two
"main.cc line 1234" so we distinguished with a rename there, but I'm not
convinced those are useful anyway.

The code that became GPSBabel turned 20 last week or so. I can accept that
things have changed and that we can make better mistakes now. :-) I'm open
to radical change.

Maybe after a release...and that's on me to prioritize now that I'm back.

tsteven4:
Let's consider it post 1.8.0.  Is make time WEB support needed still, or can we move that to configure(generate) time only?

GPSBabelDeveloper:
Thats fine. It's weird that it writes outside our own tree, but htat was jut expedience/laziness on my part. As long as I can find where it goes, it could be in ${TMPDIR} for all I care.

PR #808, opened 2022-01-03
add intial support for CTest

tsteven4:


CMakeLists.txt
```diff
@@ -430,6 +430,141 @@ message(STATUS "Libs are: \"${LnkLibs}\"")
 get_target_property(IncDirs gpsbabel INCLUDE_DIRECTORIES)
 message(STATUS "Include Directores are: \"${IncDirs}\"")
 
+set(TESTS
```
GPSBabelDeveloper, line 433: Could that be file(GLOB, testo.d/), perhaps with a get_filename_component(NAME, ....) to strip off the .test "extension"?

If it has to be, it has to be, but adding another list of format names is a drag.

PR #809, opened 2022-01-03
add support for cmake build options

tsteven4:
Unlike our qmake build the cmake build generates and links with static libraries if
we are using our internally supplied copies.  The configuration options
are not as sophisticated as in our qmake build.  The configuration of
zlib will likely be acceptable as cmake provides support via findZLIB.
The configuration of shapelib and libusb-1.0 likely will likely need to
be enhanced.


robertlipe:
These libraries and the dual "we provide our own, except when we don't"
have caused us (you) disproporation pain through the years.

While I Really Like that you can (mostly) unpack our source and go, what if
we just made these external dependencies that we didn't ship? I get that
libusb is way smaller that Qt, but we do have prior art.

Kind of a shame that zlib is in the Qt library, but they don't expose the
symbols or headers.

It's also a shame that two of the places that justified having gzip (well,
minizip which we actually don't even build) in our tree never reached my
"get around to it" list and they've become increasingly irrelevant.

Garmin .ggz files are gziped files that contain an array of GPX and an
external index file. It seems like I wrote the code but never finished it.
.kmz files are zipped files with special rules about the order of the
contents, the name of the top level file, rules about relative pathnames,
etc.

I know I also prototyped a media format-specific data type once that read
image files from Garmin GPS receivers and wrote them into KML and GPX
files; it seems a natural to put the images in a KMZ but I never got around
to it.

Interest in all of these peaked around '09 or '10 and kind of faded away...
RJL

tsteven4:
It was dynamic for check,check-vtesto using likely distribution specific find...xargs...basename...sort, before I added CTest, not GLOB, and dynamic at test time. I like the ctest support as packagers may be familar enough with it to pick it up, and it gives a nice user experience.
https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#dont-use-fileglob-in-projects

PR #812, opened 2022-01-05
add dg100 heartbeat.

tsteven4:


tsteven4:
this is @robertlipe's fix for @jose1711

PR #813, opened 2022-01-05
only support changing gpsbabel.org location at build generation time.

tsteven4:
We still allow the user to change the default location of generated files destined for gpsbabel.org, but only when generating a build system, i.e. when running qmake or cmake.  If using a makefile generator you are no longer able to change the destination at make time.  The default remains ../babelweb.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -520,22 +520,8 @@ endif()
 
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
   set(WEB "../babelweb" CACHE PATH "Path where the documentation will be stored for www.gpsbabel.org.")
-  # If using a makefile generator WEB can be overridden on the make
-  # command line, e.g.  make gpsbabel.org WEB=/tmp.  If not overridden on
-  # the make command line, or not using a makefile generator, or if using
-  # "cmake --build dir --target gpsbabel.org", then the cache variable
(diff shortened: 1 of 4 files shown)


GPSBabelDeveloper (approved):

GPSBabelDeveloper:
Related:
I've now staged https://www.gpsbabel.org/htmldoc-1.8.0/index.html
I've not updated capabilities and won't until we release. Ditto for the nag box on what's current and what's development.

tsteven4:
#813 merged. I'm good with the next few commits after d9894a8350751e3792cf19a1e056dd29b37c6e96 creating 1.8.0
1. gbbversion.cmake, gbversion.pri: GB.VERSION 1.7.0 -> 1.8.0, GB.BUILD 31->32.
2. win32/gpsbabel.rc, gui/app.rc: 2020 -> 2022
3. Do you want any of #759 merged?  the libusb.h changes are looking like they will be an issue, but babelpdf.xsl may be desirable.  I can revert libusb.h in that PR and merge if you agree.

We need to consider what flavors and Qt versions we are going to release.  I can grab them from CI and stage a draft assuming the desired versions are already in CI, but I think the retention times might be down to 7 days so we need to be a bit snappy.

GPSBabelDeveloper:
I'll cull the xslt part of #759 and commit that in a few minutes. I can't think of anything else that's really outstanding. 

Kalman filter isn't actually useful so I'm fine just leaving it as a published branch for a while.

PR #829, opened 2022-01-23
Replace which by POSIX command -v

jspricke:


tsteven4:
Thanks.

I'm generating a docker image to make sure we don't discover anything there.

If you have any questions about the new build system for 1.8.0 please ask.  It looks like Fedora is going with cmake, although we still officially recommend qmake.  cmake is lightly tested compared to qmake.  I pushed a bunch of cmake changes shortly before the release to add things like the ability to use system libraries so it should be feature equivalent.  Also note gmapbase and the translations are embedded by default.  See INSTALL.  ctest support that is equivalent to testo was added as well (but not valgrind).

jspricke:
>If you have any questions about the new build system for 1.8.0 please ask.  It looks like Fedora is going with cmake, although we still officially recommend qmake.  cmake is lightly tested compared to qmake.  I pushed a bunch of cmake changes shortly before the release to add things like the ability to use system libraries so it should be feature equivalent.  Also note gmapbase and the translations are embedded by default.  See INSTALL.  ctest support that is equivalent to testo was added as well (but not valgrind).

I went with qmake and uploaded a first version earlier:

https://buildd.debian.org/status/package.php?p=gpsbabel

The build failed for some architectures and I worked around the i386 one 
for now with:

https://salsa.debian.org/debian-gps-team/pkg-gpsbabel/-/blob/master/debian/patches/0007-Add-workaround-for-failing-test-on-i386.patch

Also I'm not sure if the translations where actually embedded.

tsteven4:
The translations and map got compiled and linked in https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=amd64&ver=1.8.0%2Bds-1&stamp=1642972522&raw=0
so they should be embedded.  I'd check a package if I could remember how.

g++ -Wl,-O1 -Wl,-rpath-link,/usr/lib/x86_64-linux-gnu -o objects/gpsbabelfe objects/aboutdlg.o objects/advdlg.o objects/donate.o objects/dpencode.o objects/filterdata.o objects/filterdlg.o objects/filterwidgets.o objects/format.o objects/formatload.o objects/gmapdlg.o objects/gpx.o objects/help.o objects/latlng.o objects/main.o objects/mainwindow.o objects/map.o objects/optionsdlg.o objects/preferences.o objects/processwait.o objects/runmachine.o objects/upgrade.o objects/version_mismatch.o objects/serial_unix.o objects/qrc_app.o **objects/qrc_translations.o** **objects/qrc_map.o** objects/moc_advdlg.o objects/moc_donate.o objects/moc_filterdlg.o objects/moc_filterwidgets.o objects/moc_gmapdlg.o objects/moc_mainwindow.o objects/moc_map.o objects/moc_optionsdlg.o objects/moc_preferences.o objects/moc_processwait.o objects/moc_runmachine.o objects/moc_upgrade.o   /usr/lib/x86_64-linux-gnu/libQt5WebEngineWidgets.so /usr/lib/x86_64-linux-gnu/libQt5PrintSupport.so /usr/lib/x86_64-linux-gnu/libQt5Widgets.so /usr/lib/x86_64-linux-gnu/libQt5WebEngineCore.so /usr/lib/x86_64-linux-gnu/libQt5Quick.so /usr/lib/x86_64-linux-gnu/libQt5Gui.so /usr/lib/x86_64-linux-gnu/libQt5QmlModels.so /usr/lib/x86_64-linux-gnu/libQt5WebChannel.so /usr/lib/x86_64-linux-gnu/libQt5Qml.so /usr/lib/x86_64-linux-gnu/libQt5Network.so /usr/lib/x86_64-linux-gnu/libQt5SerialPort.so /usr/lib/x86_64-linux-gnu/libQt5Xml.so /usr/lib/x86_64-linux-gnu/libQt5Positioning.so /usr/lib/x86_64-linux-gnu/libQt5Core.so -lGL -lpthread

jspricke:
>The translations and map got compiled and linked in https://buildd.debian.org/status/fetch.php?pkg=gpsbabel&arch=amd64&ver=1.8.0%2Bds-1&stamp=1642972522&raw=0
>so they should be embedded.  I'd check a package if I could remember how.

Right now you can get it from incoming:

https://incoming.debian.org/debian-buildd/pool/main/g/gpsbabel/

(until it get's added to the unstable distribution in some hours)

Thanks for checking!

tsteven4:
Note that we don't embed the Qt supplied translations, but we also use them.  It would be nice if the gpsbabel gui package required/recommended/suggested that they be installed.  For Fedora it's qt5-qttranslations, but I don't remember the debian package off the top of my head.

tsteven4:
Thanks.  shellcheck?

jspricke:
> Thanks. shellcheck?

You mean if I proposed this because of shellcheck? No, we currently have a discussion in Debian to drop `which` eventually, see https://lwn.net/Articles/874049/.

robertlipe:
Funny. About a week ago I was reading old UNIX archives and saw the FIRST
round of this fight between the BSD and USG camps, which resulted in
command -v being added to spec1170/POSIX because neither would budge and it
was better to create something never heard of than to give the other side
any concession.

I think it was in the mid-late 1980s. And here we are.

PR #830, opened 2022-01-24
Support Hurd in qmake

jspricke:


tsteven4:
Can you please try the attached instead, maybe we can get ahead of adding the next specifc flavor of unix.  It divides the unix world into mac and !mac.  This is what I have been doing with cmake, although that certainly hasn't seen much usage or variety.

[qmake.zip](https://github.com/GPSBabel/gpsbabel/files/7925805/qmake.zip)

jspricke:
>Can you please try the attached instead, maybe we can get ahead of adding the next specifc flavor of unix.  It divides the unix world into mac and !mac.  This is what I have been doing with cmake, although that certainly hasn't seen much usage or variety.

Works on Linux and Hurd. Can you open a PR with it?

tsteven4:
> Works on Linux and Hurd. Can you open a PR with it?
See #831
Thanks for testing, you've got quite the array of odd machines.

PR #833, opened 2022-01-26
Use smart pointers over xmalloc and manual life management

GPSBabelDeveloper:
- Use reference counted pointers instead of managing our own. 
Fixes a leak when using debug printing.


tsteven4 (approved):
We should include <memory> for std::unique_ptr.

This does split our implementation of smart pointers, using both Qt and std, but I'm ok with that.

please squash and merge so the detailed commit history of this PR doesn't live forevere in our commit history.

tsteven4:
@GPSBabelDeveloper, @robertlipe  can you squash and merge please.  I am ripping through format conversions, I have it pretty wired now.  I'm converting stuff that has copyright dates >= 2010, or recent user activity.

PR #841, opened 2022-01-27
convert Mapbar track to Format class

tsteven4:


GPSBabelDeveloper:
I'm sure you have a good mechanical system for this down, and I love the
cleanups, but you should feel free to "fix" some of these with
tools/nuke_format. Mapbar is a good example of one that I can't tell has
ever been used.

We have 1.8 out. 2,000+ downloads and nobody missed the ones we nuked last
time. We can clean dead wood.

RJL

PR #851, opened 2022-02-16
use precompiled headers to speed up compile time

tsteven4:
In my testing on Linux this reduced the compile time by a factor of 2 to 3.

I did experiment with including other often used Qt headers, but all the combinations I tested didn't improve the situation over just precompiling the Qt and C++ standard library headers used in defs.h.


GPSBabelDeveloper:
Looks good. Thank you.

PR #854, opened 2022-02-20
Use PImpl idiom to hide Format and Filter headers from users of Vecs, FilterVecs.

tsteven4:
This greatly improves the compile time of main, garmin, magproto.

There were some memory leak failures with the standard implementation using unique_ptr and make_unique.  These all related to the destruction of Formats or Filters.  These failures depended on platform and compiler.  Some were caught by valgrind, some by address sanitizer.  There were three categories:
1) nmea can create a waypoint that isn't deleted.  Normally there is a data member, curr_waypt, that points to it, so it is "still reachable".  But, if the format is destroyed then it becomes "definitely lost".
2) fatal calls exit, preventing main from cleaning up route_denit, waypt_deinit, session_exit, FilterVecs::exit, Vecs::exit, inifile done.  argument value strings can be left allocated.  These errors were found by address sanitizer.
3) GpxGlobal instance isn't cleaned up if GpxFormat::exit isn't called, which can happen with a fatal error.

Perhaps the root of these was related to the ambiguity in "Early dynamic initialization".

These can be avoided by making the implemntations static.  Vecs and FilterVecs are meyers singletons already, so this is natural.


robertlipe:
Github's online diff doesn't do this justice. I want to patch this in on a
real computer instead of my fone.

I'm not sure I care about leaks in explicit calls to exit.  The whole idea
(though maybe we should have split user error from internal consistency
problems) with Fatal was that our internal state was so screwed up that we
couldn't continue. The less code we run after that the better, I think.

If we need some other callback so that GPX (and nmea?) can toss that
tracking point and GPX can tear down that GPXGlobal (which needs a rethink)
I'm really open to that.

I always imagined the formats (and filters) subclassing the base format.
(Think of the classic triangle::area() example.) Instead of including
everything into each other, instantiating a format derived from the base
class should be enough, no? Let me think more on this.

RJL

tsteven4:
> Github's online diff doesn't do this justice.

Sorry about the diff.  It's almost all due to:
1) moving stuff unchanged from vecs.h to vecs.cc
2) reorganizing vecs.h per the google style guide.  I resisted this for a while with Format conversions to make the diff easier, but I have come to value the organized header files more than the one time diff difficulty.

The meat of this change is quite simple, and lies in Vecs::Instance which is only 3 lines.  This is the heart of the Meyers singleton,  now with an additional static variable for the private implementation.  The opaque pointer d_ptr_, is used by the public Vecs class methods to access the private part in Vecs::Impl.  So every time the vec_list is accessed it is done through d_ptr_.

In a full blow PImpl implementation all the private methods and data members would have moved to the private implementation.  This might require a wrapper for const propagation and a backward pointer q_ptr_.  But, in this case, I could achieve all my goals by just moving vec_list and the Format instances.  This greatly improves the compile time (~5x) of anybody using  Vecs (main, magproto, garmin), and removes the dependencies of those users on each of the Format based formats.  Perviously they had to be recompiled every time any one of the Format based format header files changed.

I got on to this using clang's -ftime-trace option, https://github.com/nico/ninjatracing and https://www.speedscope.app/ when looking at the performance of precompiled headers.

filters and formats already are subclasses of the Filter and Format base classes.  But, they are not dynamically allocated.  The dynamic allocation actually seems quite difficult.  We know GpxGlobal and gpx version and the stack filter count on storage between invocations, but how many other cases count on storage between invocations.  How do we find them?  How do we sort them out from the accidental cases of communication that are bugs?  If we could solve that we could ditch ::rd_init, ::rd_denint, etc and use ctors and hopefully the default dtors.  A Format base class might contain and Reader base and an Writer base.  Anyway, I don't see a reasonable way to deal with all the intentional use of memory between a formats reader and writer,  or repeated invocations of a reader or writer, or filter.

Regarding fatal, my motivation behind the obtuse syntax fatal(FatalMsg ..) was to allow fatal to throw an exception.  Then by adding a try/catch around "rc = run(prog_name);" in main we can do a lot of cleanup.  In practice this seems to work well, and reduces the amount of "still reachable" clutter in a valgrind log.  A benefit of that is that it is easier to find any "still reachable" real issues like GpxGlobal.

tsteven4:
Do you want more time to mull this over?

GPSBabelDeveloper:
Now that I'm home and able to see this on a real computer, I can work
through it better.  It's still tough to distinguish code motion from Real
Work, but I think I have it now.

As for the issues you've raised along the way.
I think we've "proven" there's little communication between the formats by
having all the data and members as static going back to the C days. There
are a few violations of "buddy" formats (there's some geocaching stuff in
gpx.cc that just needs to be alone, I added some hooks in GPX for another
format that made chopped up GPX files, but that format didn't land,
magproto and magellan are related, etc.) but I don't think we'll find a lot
of cross-module chatter.

There are a few cases where we intentionally persist things, like the
gpx_global stuff. We just move that out of the format class itself and into
some other format or global (d_impl) that does persist.  I'm really not
convinced that merging all those headers is a great idea anyway.

There are a few cases where we leak after fatal. Do we really care? By
definition, we call fatal when something is so screwed up that we've
decided we can't carry on. Maybe this is the systems programmer in me
making it think of a panic() where, say, our wpt_list isi corrupt and any
further accesses risks a crash anyway. But if we leak that NMEA
accumultaor point during a fatal(), I don't see that as bad.

The design 20 years ago was indeed for rd_init/rd_deinit to model ctors and
dtors. C++ just wasn't yet a viable thing at the time.  Whether they
actually ARE ctors/dtors doesn't bother me so much.

At one point, I'd prototyped a dynamic vec_list[] to get all that format
stuff out of vecs*[ch] and just leave it in the format and have a global
ctor in the format that push_back'ed the data. I was conflicted on how it
felt. It was nice be able to have formats come and go just by their
presence in the build, but global static ctors are trending downward
(there's a name for the "fiasco", but I think we're mostly immune from such
ordering issues) and I was concerned about project startup time. In all, I
dropped the prototype.

So there's fertile tinkering opportunities in this area, but I'm not sure
how much tangible gain there is. Certainly you're wise to tackle that ugly
compiliation time issue, which I had noticed, but I'm not sure how much
further to take it.

So, once again, I'm hip with this and thank you for tackling it and for
waiting for me to get my act together to review it properly.

RJL

PR #855, opened 2022-02-22
Move compile-time tests for debug message to runtime tests at -D3 or more.

robertlipe:
Move compile-time tests for debu message to runtime tests at -D3 or more.


tsteven4 (commented):
You should delete lines 75-77 now.  The macro is now defined at line 90.
image

PR #856, opened 2022-02-23
use Qt for os independent environmental variable access.

tsteven4:


robertlipe:
Looks good. Thanx.
(I checked offline with the creator of that original getenv osdep code.)

PR #858, opened 2022-02-24
fix various printf related issues.

tsteven4:
printf arugment reording is not available on MSVC with printf (it requires_printf_p).  don't use it (garmin_gpi)

lowranceusr has trouble printing time_t.  avoid this by switching to int64_t and using PRId64.  Note QDateTime is using qint64 for SecsSinceEpoch.

lowrancceusr has trouble printing size() related Qt methods.  These changed from int in Qt5 to qsizetype in Qt6.  Qt6 does offfer PRIdQSIZETYPE, but this is available before 6.2.  Punt and just let Qt do the conversion with QByteArray::number().constData() and print the resulting string.

### DIFF
--- a/garmin_gpi.cc
+++ b/garmin_gpi.cc
@@ -1119,19 +1119,19 @@ GarminGPIFormat::load_bitmap_from_file(const char* fname, const unsigned char** 
   }
 
   if (GPI_DBG) {
-    printf("data size:             0x%1$x (%1$d)\n", src_h.size);
-    printf("image data offset:     0x%1$x (%1$d)\n", src_h.image_offset);
-    printf("header size:           0x%1$x (%1$d)\n", src_h.header_size);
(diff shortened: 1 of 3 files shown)


GPSBabelDeveloper (approved):
Most of these are reasons to move to "normal" C++ streaming operators. We just have a ton of printf-y code. :-/

LGTM. Thanx.

tsteven4:
> Most of these are reasons to move to "normal" C++ streaming operators. We just have a ton of printf-y code. :-/
> 
> LGTM. Thanx.

Yes.  But manually converting 130 printfs was a bit beyond my dedication.  The QByteArray::number().constData() thing sort of splits the difference, but it does rely on QByteArray::number() accepting qsizetype, fortunatley qsizetype is either int or long long int today.

GPSBabelDeveloper:
I get it. It's like we're a 20 year old C project. It's just another option
to keep in mind.

Thanx for spending the dedication you do.

PR #860, opened 2022-02-26
fix Qt6 deprecation warning with QVariant::canConvert.

tsteven4:


tsteven4:
This showed up on the macOS build with Qt6.

PR #861, opened 2022-03-07
misc improvements.

tsteven4:
fix memory leak in nmea reader.
eliminate extra string conversion using xml_parse_time.
use local QByteArray instead of xstrdup.
convert human_to_dec interface to QString.


csv_util.cc
```diff
@@ -424,9 +415,8 @@ human_to_dec(const char* instr, double* outlat, double* outlon, int which)
     case '9':
     case '0':
     case '.':
-    case ',':
       numres[numind] = atof(cur);
-      while (cur && *cur && strchr("1234567890.,",*cur)) {
+      while (cur && *cur && strchr("1234567890.",*cur)) {
```
GPSBabelDeveloper: is that inbytes.indexOf(*cur) ? It's very odd to see a string literal as the heystack for strchr. Can it ant eh line above reduce to:

numres[numind] = strtoul(cur, &cur, 10): ? That lets the library parse it into numres, from cur, and writes cur with the end of the number it parsed. 

Maybe inbytes.indexOf() lets it stay totally in QString place without cur at all.

Either way, it's a big improvement, so if there's no easy fruit to havest here, it's OK.

Maybe that lets cur go away completely and we can operate totally in QString space.

tsteven4: strtod lets us get rid of the end of conversion detection.  This doesn't seem to be directly supported in Qt.

robertlipe:
Cool. If we can't eliminate the case of the same data existing in both
Qt-land and C-string-land at the same time, this is still better. I find
this easier to read.

Thanx!

PR #862, opened 2022-03-10
retire strsub, gstrsub.

tsteven4:
These routines either had no usage, or were used only by explorist_ini.  

The only possibility I see for using any of the explorist_ini code is on APPLE (os_get_magellan_mountpoints()).  This also is one of two cases where a format is invoked outside of main.

explorist_ini_get purports to search through a list of paths, but the pointer is never advanced.

explorist_ini has NO test coverage, i.e. 0%.

The replaced code has multiple leaks from xstrdup and gstrsub.

I suspect explorist_ini should be excised in it's entirety.


GPSBabelDeveloper:
Awesome. Thanx.

That format was a weird appendage and explorist is about to go walk the
gallows anyway.

Bonus. That those functions were horrible...

PR #872, opened 2022-03-18
update to Qt 6.2.4

tsteven4:


robertlipe (approved):
I dig how easy you make this look these days.

tsteven4:
> I dig how easy you make this look these days.

Yeah, well there was a few hours in #871.  That issue started to throw some warnings in 6.2.3, but with 6.2.4 map preview flat out didn't work when gmapbase.html was a local file.

GPSBabelDeveloper:
I suspected there was some blood behind the scenes but it's really pretty
awesome how we (you) can flip a digit in a control file and get all the
builds updated.

I know it's hard to make it *look* easy.

RJL

PR #874, opened 2022-03-31
Use QStandardPaths to locate data files

turboencabulator:
On Linux where gpsbabel and gpsbabelfe are presumably placed in /usr/bin (and whether or not gmapbase.html and the translations are embedded), the program will still be searching for /usr/bin/gmapbase.html and /usr/bin/translations/*.  No sane Linux package would place those data files in /usr/bin.  A better location is to search for them in a subdirectory of /usr/share, and this change will do just that -- searching in /usr/share/GPSBabel/GPSBabel (taken from the OrganizationName and ApplicationName that are set in gui/main.cc).

Note that I'm unable to test this for Mac/Win; any scripts to create packages for them may need to move the files to an appropriate location.

### DIFF
--- a/gui/map.cc
+++ b/gui/map.cc
@@ -30,6 +30,7 @@
 #include <QLatin1String>          // for QLatin1String
 #include <QMessageBox>            // for QMessageBox
 #include <QNetworkAccessManager>  // for QNetworkAccessManager
+#include <QStandardPaths>         // for QStandardPaths
 #include <QStringLiteral>         // for QStringLiteral
 #include <QUrl>                   // for QUrl
 #include <QWebChannel>            // for QWebChannel
@@ -89,13 +90,13 @@ Map::Map(QWidget* parent,
   connect(mclicker, &MarkerClicker::logTime, this, &Map::logTime);
 
   // We search the following locations:
-  // 1. In the file system in the same directory as the executable.
+  // 1. In the file system in the QStandardPaths::AppDataLocation directories.
   // 2. In the Qt resource system.  This is useful if the resource was compiled
   //    into the executable.
-  QString baseFile =  QApplication::applicationDirPath() + "/gmapbase.html";
+  QString baseFile = QStandardPaths::locate(QStandardPaths::AppDataLocation, "gmapbase.html");
   QString fileName;
   QUrl baseUrl;
-  if (QFile(baseFile).exists()) {
+  if (!baseFile.isEmpty()) {
     fileName = baseFile;
     baseUrl = QUrl::fromLocalFile(baseFile);
   } else if (QFile(":/gmapbase.html").exists()) {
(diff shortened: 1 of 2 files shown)


gui/map.cc
```diff
@@ -89,13 +90,13 @@ Map::Map(QWidget* parent,
...
-  // 1. In the file system in the same directory as the executable.
+  // 1. In the file system in the QStandardPaths::AppDataLocation directories.
   // 2. In the Qt resource system.  This is useful if the resource was compiled
   //    into the executable.
-  QString baseFile =  QApplication::applicationDirPath() + "/gmapbase.html";
+  QString baseFile = QStandardPaths::locate(QStandardPaths::AppDataLocation, "gmapbase.html");
   QString fileName;
   QUrl baseUrl;
-  if (QFile(baseFile).exists()) {
+  if (!baseFile.isEmpty()) {
```
tsteven4, line 99: why not check existence?

tsteven4 (commented):
Are you packaging gpsbabel for a distribution?

These used to be the most patched lines of our code.  The current search list was worked out with a number of packagers and allows them to avoid patching location information.  Is there a reason you don't want to embed the translations and map?

I believe your changes will break our mac distribution.

gui/map.cc
turboencabulator: QStandardPaths::locate() "Returns the absolute path to the first file or directory found, otherwise returns an empty string" according to its documentation.  Checking for existence would be redundant.

turboencabulator:
Yes I'm packaging (for Slackware), and yes I've patched those paths in the past.  I read through some of the PR discussions relating to where to put these files.  In my experience, when there's no guidance from upstream on a preferred location, usually a few distros will choose their preferred location and the rest follow suit.  (I copied from Arch.)  My assumption is that Qt didn't offer a good mechanism for you to choose a location (until QStandardPaths) so the easiest thing for you to do was to place data files relative to the executable.  That's probably fine/expected for Windows but not Linux, hence all the patching.

I don't know anything about Mac's packaging guidelines.  Can we determine where AppDataLocation resolves to on Mac and move the files there?

Some reasons for not embedding:  Users could tweak gmapbase.html, add/modify translations, etc. without rebuilding.  If building for multiarch, the data is in one location instead of in each executable.

tsteven4:
It looks like archlinux has embedded the gpsbabel translations and dropped support for map preview.  They have dropped their install and no-map patches.  See https://github.com/archlinux/svntogit-community/commit/9bb2b0d11cdcbf58b1f87452800d04cceb390a57

If you use the file scheme for gmapbase you may need https://github.com/GPSBabel/gpsbabel/commit/c4b5d6abed04b4c1248c9556c4358d75e0291af3

GPSBabelDeveloper:
Thankx, but this is an OS-local change that should be handled by the packaging systems on those OSes.

PR #882, opened 2022-05-29
add option to transform filter to delete creation times

tsteven4:
This can be useful when one has recorded a track, e.g. with strava, and wants to create a route, e.g with ridewithgps.

./gpsbabel -i gpx -f track.gpx -x transform,rte=trk,del,timeless -x simplify,error=2m -o gpx -F route.gpx

### DIFF
--- a/xmldoc/filters/options/transform-timeless.xml
+++ b/xmldoc/filters/options/transform-timeless.xml
@@ -1,4 +1,4 @@
 <para>
 This option tells
-GPSBabel to creating points without creation times instead of copying the creation time from the source point.
+GPSBabel to create points without creation times instead of copying the creation time from the source points.
 </para>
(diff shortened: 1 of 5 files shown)


xmldoc/filters/options/transform-timeless.xml
```diff
@@ -0,0 +1,4 @@
+<para>
+This option tells
+GPSBabel to creating points without creation times instead of copying the creation time from the source point.
```
GPSBabelDeveloper: * to create, 

I like this filter. Useful for KML/Earth, too,  since the Time handling there is confusing.

LGTM

GPSBabelDeveloper (approved):

robertlipe:
Good idea. Thanks.

PR #886, opened 2022-06-23
qt6_depr

robertlipe:
(If this isn't in Qt 5.whatever, we may have to do some ifdef gunk.)


GPSBabelDeveloper:
Wake me up with Linux has Qt 7...

PR #888, opened 2022-07-06
retire qsort in favor of std::sort, std::stable_sort

tsteven4:
Also, fix a bug in the simplify filter with the sort compare filter.  This bug didn't show up in our test cases on windows, linux, or macos, but did show with std::sort on macos, but not on windows or linux.  The sort compare function incorrectly compared two HUGEVAL entries.


garmin_txt.cc
```diff
@@ -834,17 +826,19 @@ garmin_txt_write()
 
   if (waypoints > 0) {
     wpt_a_ct = 0;
-    wpt_a = (const Waypoint**)xcalloc(waypoints, sizeof(*wpt_a));
+    wpt_a = new const Waypoint*[waypoints]{};
```
GPSBabelDeveloper, line 829: wpt_a has a super weird life cycle. It's allocated and destroyed here, but is used in enum_waypt_cb, but isn't passed as an argument. It's a global, but it's not clear taht it should be.

You didn't leave this aspect any worse than you found it, but that's a good cleanup tasks for anyone looking for a good first time PR.

garmin_txt.cc
```diff
@@ -834,17 +826,19 @@ garmin_txt_write()
...
     waypt_disp_all(enum_waypt_cb);
     route_disp_all(nullptr, nullptr, enum_waypt_cb);
-    qsort(wpt_a, waypoints, sizeof(*wpt_a), sort_waypt_cb);
+    auto sort_waypt_lambda = [](const Waypoint* wa, const Waypoint* wb)->bool {
+      return wa->shortname.compare(wb->shortname, Qt::CaseInsensitive) < 0;
```
GPSBabelDeveloper, line 833: This doesn't have to preserve the subsort on index because we're assuming that nobody actually cares about the order and that stable_sort() will keep the nuanced sort cases under wraps for cases like our tests, right?

It seems weird to have this in garmin_txt, but I'm not at all sure how garmin_txt is used in the real world or if it is at all.

GPSBabelDeveloper (approved):
I like it.

I started down this path recently on my regular "how can we kill xcalloc" scan of the code and got tangled up in something before losing my footing, so this was fresh (ish) on my mind.

This actually hits two cases (garmin_txt, exported date) that were pet projects of two earlier developers and I'm not at all sure are seeing any regular use. Cleaning up two cases of variable length arrays via xcalloc is helpful. (Most of our remaining ones are in places that require hardware to test, so they take more of a commitment.)

Thanx for tackling it.

garmin_txt.cc
tsteven4: duplicate (not garmin_txt) didn't preserve the final sort key of index, but the new sort should work exactly the same as the old sort.  We are NOT assuming nobody actually cares about the order, the behavior is equivalent and matches the preexisting comment in duplicate.cc L84-115.   In this case stable_sort was not used for test case portability.  Stable sorting algorithms preserve the relative order of equal elements.  Adding the index as the final sort key is a method to get a stable sort using qsort.  Stable behavior is part of the std::stable_sort contract: "The order of equivalent elements is guaranteed to be preserved."

This isn't the first conversion from qsort to std:stable_sort that has allowed us to simplify the sorted objects and comparison function by dropping the index.  However, there is at least one usage of stable_sort elsewhere that was done solely for portability of test cases.

tsteven4: I don't know why garmin_txt sorts waypoints, I just re-implemented the previous behavior.  I do wonder if garmin_txt, and several other formats are ever used.  So far you  have done a remarkable job of deprecating formats without complaint!

garmin_txt.cc
tsteven4: I agree wpt_a is a bit weird.  It's not just wpt_a, also wpt_a_ct, gtxt_flags.enum_waypoints, waypoints.  This pattern of using variables with internal, external or module linkage with a callback function was unfortunately common with waypt_disp_all and route_disp_all.  It often comes up when converting from a LegacyFormat to a Format.

I stopped before I fell down the slippery slope by implementing too many good ideas in one PR.  Some good ideas are
1) use a smart pointer with new
2) use a container instead of a plain array
3) use a lambda with capture as a callback instead of variables with internal, external or module linkage and a function.

robertlipe:
Jolly good. Thanks for the clarifications. Merge at will.

PR #889, opened 2022-07-14
string optimizations

tsteven4:
This PR contains two string optimizations.   The first commit uses QStringLiteral("\*").arg... instead of QString("\*").arg...  The second uses utf-16 strings with QDateTime, QDate, QTime literal string format arguments.


GPSBabelDeveloper:
Nicely done. Thanks.

PR #892, opened 2022-07-17
various clang tidy fixes

tsteven4:


GPSBabelDeveloper:
OK. 

I'm not certain everything here is clearly better, but I'm trying to quit arguing with robots. Delegate to our clang-tidy overlords. :-)

tsteven4:
Although found by the bot I hand picked these particular fixes.
duplicate includes - slam dunk
static member functions - besides readability promising and enforcing no access to non-static member data this allows us to avoid wrapping callbacks in functors or lambdas for *_disp_all.
const parm in declaration - Not universally accepted but made sense to me, https://abseil.io/tips/109
raw string literal - I accepted 1/3 instances.  I think regular expressions are easier to read without the escapes, and this is consistent with other regular expressions in the same file.

Issue #895, opened 2022-07-30
unmatched tags in geocache reference file


tsteven4:
1. https://github.com/GPSBabel/gpsbabel/blob/8dae11e100677217c6b4cd858a76f3bf6bc8d12c/reference/gc/GCGCA8.gpx#L46
The tags &lt;/a&gt;&lt;/p&gt; are unmatched.
This causes the output from gpsbabel -i gpx -f reference/gc/GCGCA8.gpx -o html -F GCGCA8.html to contain unmatched tags.
2. There is an additional problem.  The P end tag we add after gpsbabeldesclong is unmatched because of the UL tags in what we think is the paragraph content.

robertlipe:
That HTML was created by hand (mine) that was rendered inside an iframe of
browsers and using crudest tools almost 20 years ago. It's even possible
that it wasn't even really written in 'real' HTML as I think the cache page
editor of that time used sort of a bbcode markup that was transcribed on
the server.

As long as we can approximately round-trip what's inside
groundspeak:long_description,
what we do with that HTML is up to us. That cache placement has long since
been removed and out of circulation. There's no need to overthink keeping
it in sync with a master copy or whatever.

We're free to change that reference file and the ones that come from it as
we see fit. We're free to edit the past if it makes the present or future
better.

Signed,
Winston Smith

robertlipe:
Is there remaining action here? That embedded HTML snippet is from ancient browser days, when <p>s were considered self closing. (This was before even XHTML.) It's certainly in "quirks mode" even if it's not outright broken.

If you're OK with me deleting &lt;/a&gt;&lt;/p&gt at the end of long description and fixing any fallout reference files I'll do that. It's trying to close a link that's not here  and a paragraph that was never opened. That's just copy-paste error in the source material from the Mozilla 1.5 era or so...

We're not testing the validity of our embedded HTML, so I don't think we have to faithfully reproduce/handle/fix what was clearly broken HTML. 

I'm happy to do the fix if you agree that's OK.

robertlipe:
Sticking a note to https://github.com/GPSBabel/gpsbabel/pull/1196

tsteven4:
> 2\. There is an additional problem.  The P end tag we add after gpsbabeldesclong is unmatched because of the UL tags in what we think is the paragraph content.

Point 2. was fixed in dcdbb6a90353231497273d3062c2cd5d396e1d6e

tsteven4:
[...]

PR #905, opened 2022-08-03
convert html writer to html 5.

tsteven4:
this also fixes a bug where we could fail to entitize wpt->description.
this also fixes a bug where we could create illegal fragments and id
attributes.

There are some minor changes in rendering with firefox related to margins in p elements.

Every gpx reference file we have was converted to html, and the html was checked with tidy.  There are two files that have warnings, and no errors.
```
reference/tpo-sample3.gpx:
line 250 column 1 - Warning: <a> converting backslash in URI to slash
```
We are interpreting 'w:\photo.jpg' in tpo-sample3.tpo as a link.
```
reference/gc/GCGCA8.gpx:
line 45 column 154 - Warning: discarding unexpected </a>
line 45 column 158 - Warning: inserting implicit <p>
line 45 column 158 - Warning: trimming empty <p>
```
GCGCA8.gpx has two unpaired end tags (entitized) in a groundspeak:long_description element. Also see #895.

The html files produced by text.test were checked with https://validator.w3.org.  Related to GCGCA8.gpx we see these warnings/errors:
```
Warning: Consider adding a lang attribute to the html start tag to declare the language of this document.
Warning: The border attribute is obsolete. Consider specifying img { border: 0; } in CSS instead.
Error: Stray end tag a.
Error: No p element in scope but a p end tag seen.
```

### DIFF
--- a/html.cc
+++ b/html.cc
@@ -57,6 +57,13 @@ HtmlFormat::wr_deinit()
   mkshort_del_handle(&mkshort_handle);
 }
 
+QString HtmlFormat::create_id(int sequence_number)
+{
+  // It's easier to create a legal fragment and identifer from scratch than
+  // from user supplied text.
+  return QStringLiteral("WPT%1").arg(sequence_number, 3, 10, QChar('0'));
+}
+
 void
 HtmlFormat::html_disp(const Waypoint* wpt) const
 {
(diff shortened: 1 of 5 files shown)


robertlipe:
Well done.

The html inside PQs is strange and doubly so for the age of our examples.
They predate html5 by far and the html is meant to be embedded inside a
geocache listing page, but not inside an iframe. That's why there's not a
top level html tag, for example. Further the html is often rendered by a
handheld gps that doesn't have a full html browser so there are all kinds
of unwritten rules.

In short, don't stress about the html inside our GC reference files. The
html target was originally written to take those files and render them
viewable on PDAs before the GPSes could.

tsteven4:
QXmlStreamWriter is nice for XML, and QTextStream seems better than gbf*, but for HTML Qt doesn't seem to have a good solution.  QTextDocument in 6.3.2 supports a subset of HTML 4.  XHTML doesn't seem to have much share.

PR #910, opened 2022-08-08
delete obsolete and unused document sections.

tsteven4:
all these files correspond to formats, format options, filters and filter options that no longer exist.  They were not referenced when building the document.


robertlipe:
Before submitting, can you please make nuke_format clean up at least some
of these?

I  don't think I ever actually finished that big cleanup/deletion round I
started some months ago.

Thanx for your diligence in this stuff!

tsteven4:
Without analyzing nuke_format too thoroughly it assumes the source file roots match the xml tree document roots.  I think this only occurs by convention, and the convention is not always adhered to.  Lots of these deletions predate nuke_format.

I counted on xmldoc/makedoc creating the files it expects to see that don't exist (provided the directory exists). After generating the document with empty xmldoc/formats, xmldoc/formats/options, xmldoc/filters, xmldoc/filters/options I compared these to the files that exist.

We also have many orphaned reference files.  Conventions in naming reference files are even less consistent.

GPSBabelDeveloper:
All true. "even less consistent" is kind - it's a mess.

Sometimes I do builds and look at the access timestamps on things to look
for stragglers but they're not always easy to find.

Still, if the fruit isn't hanging low enough that a line or two in
nuke_formats would help prevent it from happening again, so be it.

Thanx for checking.

I think it's probably time for new rules within the orphanage, too. Maybe
one release after we've put a format in deprecated, we remove them. With
Git, it's not like anything is ever lost. It prevents us from having the
size on a checkout and from needing to exclude the directory from ack or
other tools that look at the entire directory structure. So far, we've not
really had enough screaming about any of our deletions that we've seriously
considered putting them back. Maybe a deader form of dead is in order.

Not in this PR, of course. Just a discussion.

PR #911, opened 2022-08-09
delete reference files not used by testo (or whitelisted)

tsteven4:
the script tools/obsolete_unused_reference_files does this,
although it might be delicate.
note any unused files in the reference tree that are not whitelisted will be deleted!

Indeed, alternate reference files, readme files, and other files files we derived our references from must be handled specially.  The ones I have found are whitelisted in the updated obsolete_unused_reference_files script.


tsteven4:
@robertlipe please glance at the long list of files and see if you can catch any more that we want to keep despite them not being used by testo.

tools/obsolete_unused_reference_files
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -ex
+# delay -t must be long enough for testo to complete
+inotifywait -r -m -t 120 reference > log&
+# wait to make sure inotifywait is monitoring
+sleep 10
+./testo
+# wait to make sure inotifywait is done
+sleep 120 
+grep ' OPEN ' log | sed 's/ OPEN //' | sort | uniq > used
+find reference/ -type f | sort | uniq > exists
```
robertlipe: Is 'sort -u' (here and above) a suitably universal thing? I think it's POSIX 1003.1.

PR #913, opened 2022-08-15
QStringize vecs a bit more.

tsteven4:
This fixes an ancient FIXME regarding the use of the option name
to indicate a option was given without a value.  This practice was unnecessary as we can (and could before) distinguish between an option not being given (nullptr or isNull), an option be given without a value(pointer valid and points to \0 or !isNull && isEmpty), and option being given with a value.

This was tested with a custom format and filter that echoed the received argument values.  However, this test apparatus was not included in this PR. 

This enhances the validity checking of default, minimum and maximum argument
values.

More rigorous input validation was attempted with std::stoi & std::stod.  This revealed that two trackfilter options had range errors.  Although the values of these options are a series of decimal digits they are interpreted as a string by the filter.  The GUI handles creates these values from dates and times.  This PR changes these argtypes from ARGTYPE_INT to ARGTYPE_STRING.

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -960,13 +960,13 @@ void setshort_defname(short_handle, const char* s);
 #define ARG_NOMINMAX nullptr, nullptr
 
 struct arglist_t {
-  const char* argstring{nullptr};
-  char** argval{nullptr};
-  const char* helpstring{nullptr};
-  const char* defaultvalue{nullptr};
+  const QString argstring;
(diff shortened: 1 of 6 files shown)


tsteven4:
Teaser (not in this PR): something like this to move conversions and QStrings into vecs a bit at a time?
using argval_t = std::variant<std::monostate, char**, int*, double*, bool*, QString*>;'

PR #916, opened 2022-08-23
fix GUI handling of track filter start/stop times.

### LINKED ISSUE #915: GUI trackfilter start and stop time zone conversion is incorrect.
When the "Local Time" checkbox is checked(unchecked) I think the time in the start/stop edit box should be interpreted as a local time(UTC).  The time passed to the CLI is always UTC according to our documentation.

Local Time checked.  But value passed has not been converted to  UTC. (3PM local = 15 hours local)
image

Local Time unchecked.  But value passed has been converted from local to UTC. (3PM local = 15 hours local = 21 hours Z)
image

tsteven4:
This resolves #915

The timeSpec of the QDateTimeEdit objects is maintained in
accordance with the TZCheck QCheckBox.  This solves additional
issues with DST transitions and edits.

As an aside, when the TZCheck box check status changes, the times shown in the start and stop edits will be converted between the two time zones.  This behavior is part of QDateTimeEdit.

There is an ambiguity in the fall with local times, we don't have a way to distinguish 1:30AM daylight time from 1:30AM standard time.

### DIFF
--- a/gui/filterdata.cc
+++ b/gui/filterdata.cc
@@ -55,14 +55,9 @@ QStringList WayPtsFilterData::makeOptionString()
 }
 
 //------------------------------------------------------------------------
-static QString optionDate(const QDateTime& dt, bool useLocal)
+static QString optionDate(const QDateTime& dt)
 {
-  QDateTime d;
-  if (useLocal) {
-    d = dt.toLocalTime();
-  } else {
-    d = dt.toUTC();
-  }
+  QDateTime d = dt.toUTC();
 
   QDate date = d.date();
   QTime time = d.time();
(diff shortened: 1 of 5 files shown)


GPSBabelDeveloper (approved):
Looks great. Thanks! 

How much longer is qt5 in play?

tsteven4:
I pushed a further change to use radio buttons for the LocalTime/UTC choice.  They make more sense as these options are exclusive, and the choice is clearer as opposed to the Local Time checkbox which doesn't make the alternative clear.

Unfortunately the latest ubuntu LTS (22.04.1) release ships with a defective Qt6 core that makes Qt webengine DOA.  Despite my fix being accepted in Debian (and available in stable back ports) Ubuntu seems disinclined to pick it up.
https://bugs.launchpad.net/ubuntu/+source/qt6-base/+bug/1970057
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1010575

robertlipe:
On Thu, Aug 25, 2022 at 8:27 AM tsteven4 ***@***.***> wrote:

> Unfortunately the latest ubuntu LTS (22.04.1) release ships with a
> defective Qt6 core that makes Qt webengine DOA. Despite my fix being
> accepted in Debian (and available in stable back ports) Ubuntu seems
> disinclined to pick it up.
>
> https://bugs.launchpad.net/ubuntu/+source/qt6-base/+bug/1970057
> https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1010575
>

Good grief.

They whine when we provide our own, then ship bad libraries for months even
after you fix their brokenness for them and/or they know a fix is available
upstream.

In the early 90's the thought of shared libraries seemed really nifty, but
now with decades of experience, I'm not so sure it's a win at all.  Sigh.

Thanx for the good fight.

RJL

PR #918, opened 2022-09-01
add test driver for GUI.

tsteven4:
This can be run under cmake. I haven't, and don't intend to, enhance our qmake support at this point.

mkdir bld
cd bld
cmake ..
cmake --build . --target GPSBabelFE_test
ctest -V -R test-gui

I haven't added coverage collection for the gui yet, nor merged the gui coverage and the cli coverage reports

This is working on macOS and Ubuntu and windows.


robertlipe:
Clever. I've started something like this a time or two through the years and never got anywhere with it. This looks really promising.

PR #920, opened 2022-09-05
prepare for deprecation of qmake.

tsteven4:
build a release level package in CI with cmake on windows & macOS.
this can be compared with the release package build in CI with qmake.

support cmake multi-config generators on macOS (e.g. Xcode) as well
as well as single-config generators (.e.g Unix Makefiles, Ninja).

support packaging of out of source builds on macOS when using cmake.
note that Qt6 support for packaging still has many features in
preview, thus we still use our script based packaging.

move some Ubuntu CI jobs to jammy.  Unfortunately Qt6WebEngine is
unusable on jammy due to a bug in qt6-base.  So we are still using
Qt5 :(
https://bugs.launchpad.net/ubuntu/+source/qt6-base/+bug/1970057

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -17,15 +17,18 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - QT_VERSION: '5.12.12'
-            XCODE_VERSION: '11.7'
-            RELEASE: false
           - QT_VERSION: '5.15.2'
             XCODE_VERSION: '12.4'
+            GENERATOR: qmake
             RELEASE: false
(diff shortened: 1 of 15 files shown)


tsteven4:
I am going to cut this off and leave for the future:
1) cleanup of package_app target and scripts.  It will be simpler once we drop qmake.
2) movement of linux CI jobs from qmake to cmake
3) dropping Windows CI qmake jobs
4) dropping macOS CI jobs, adding a cmake Xcode job.
5) make sure we have a Debug build type in CI to test assertions.

PR #923, opened 2022-10-02
Test binary and text separators

robertlipe:
- A quick test to see if all our compilers re c++14 enough to support separators in 0xab'cdef and separators and binary prefixes 0b0110'0011.

### DIFF
--- a/wbt-200.cc
+++ b/wbt-200.cc
@@ -43,7 +43,7 @@
 
 #define TK1_MAGIC       "WintecLogFormat"
 #define TK1_DATA_OFFSET 0x0400
-#define TK1_END_FLAG    0x04000000ul
+#define TK1_END_FLAG    0x0400'0000ul
 
 /* Used to sanity check data - from
  *   http://hypertextbook.com/facts/2001/DanaWollman.shtml
@@ -504,7 +504,7 @@ static double get_param_float(const char* cmd)
     int min    = (((tim) >>  6) & 0x3F); \
     int hour   = (((tim) >> 12) & 0x1F); \
     int mday   = (((tim) >> 17) & 0x1F); \
-    int mon    = (((tim) >> 22) & 0x0F); \
+    int mon    = (((tim) >> 22) & 0b0'1111); \
     int year   = (((tim) >> 26) & 0x3F);
 
 static time_t decode_date(uint32_t tim)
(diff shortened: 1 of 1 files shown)


tsteven4:
We require c++17 already.  Usually, we have had problems with old versions of Microsoft Visual Studio tools, but things have been pretty good since we set the floor at MSVC 2017.  main enforces >= MSVC 2017 at any version (_MSC_VER >= 1910).  We test with the final tools of MSVC 2017 (MSVC++ version 14.16, _MSC_VER 1916, version 15.9).  There may be a hole with an earlier MSVC 2017 version.  If there is I'm happy to move the main test up to _MSC_VER >= 1916.

robertlipe:
I'll probably dump this change, but the ruling on the field is that "new" delimiters and binary constants are allowed in GPSBabel.

long million = 1'000'000;
long register = 0xffff'f0f0;
long b31 = 0b01000'0000'0000'0000;

I don't particularly plan to change our existing uses, but I'll lean on this more in new code, particularly hardware-heavy code.

tsteven4:
Both those are c++14, they shouldn't be a problem.
https://en.cppreference.com/w/cpp/language/integer_literal (binary literals and optional single quotes.

robertlipe:
> We require c++17 already. 

Awesome. I misread the "gnu++1z" in the generated build. I should have looked it up. 

> If there is I'm happy to move the main test up to _MSC_VER >= 1916.

Given how rare external contributions have become, the floor can be whatever we're happy with our auto-builders running.

> Both those are c++14, 

I'd read about these becoming standard long ago and wasn't in a position to use them then (plus I now have the memory of a goldfish)  but they're nice to have in the bag of tricks..

In C, they're rejected in c17 and gnu17, but allowed in -std=c2x and gnu2x. [ /me visits Google ] They're really new in the standard, though I worked on/with a C compiler and assembler that supported this syntax in the '90's. Weird.
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2549.pdf
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2606.pdf
Looks like both features are mentioned in the upcoming (likely '23) C standard and ahve been common for a long while in C++. https://en.wikipedia.org/wiki/C2x

That version gets us (I have another project in C) much nicer parity between the languages. Jolly good. My days of tracking and even attending those meetings are over.

robertlipe:
Closed (at least partially to save time on build runners) but with the potential AI of changing autobuild to use and enforce 2017 Update 9,  _MSC_VER  1916, _MSC_FULL_VER  191627030, Visual Studio 2017 version 15.9.11.

Honestly, if we bump it, should we go to 2019 Update 11 or maybe even Visual Studio 2022 version 17.3.4 ? (assuming it's an autobuild option)

Note to self: decoder keyring at https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd


<br class="Apple-interchange-newline">

PR #927, opened 2022-10-06
build: Remove headers from sources

jtojnar:
CMake actually does not need them listed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -202,96 +202,8 @@ if(${QT_VERSION_MAJOR} EQUAL "6")
   set(SUPPORT ${SUPPORT} src/core/codecdevice.cc)
 endif()
 
-# HEADERS
-set(HEADERS
-  cet_util.h
-  csv_util.h
-  defs.h
-  dg-100.h
-  exif.h
-  explorist_ini.h
-  filter.h
-  filter_vecs.h
-  format.h
-  formspec.h
-  garmin_device_xml.h
-  garmin_fit.h
-  garmin_fs.h
-  garmin_gpi.h
-  garmin_icon_tables.h
-  garmin_tables.h
-  gbfile.h
-  gbser.h
-  gbser_private.h
-  gdb.h
-  geojson.h
-  globalsat_sport.h
-  gpx.h
-  grtcirc.h
-  gtrnctr.h
-  heightgrid.h
-  holux.h
-  humminbird.h
-  html.h
-  inifile.h
-  kml.h
-  legacyformat.h
-  lowranceusr.h
-  magellan.h
-  navilink.h
-  nmea.h
-  osm.h
-  qstarz_bl_1000.h
-  random.h
-  session.h
-  shape.h
-  skytraq.h
-  strptime.h
-  subrip.h
-  text.h
-  unicsv.h
-  units.h
-  vecs.h
-  xcsv.h
-  xmlgeneric.h
-  jeeps/garminusb.h
-  jeeps/gps.h
-  jeeps/gpsapp.h
-  jeeps/gpscom.h
-  jeeps/gpsdatum.h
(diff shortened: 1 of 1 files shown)


tsteven4:
Thanks.  This is attractive from a maintenance point of view, but may have undesirable side effects.

The list of sorted HEADERS may be used by various IDEs.  The sorting came in with https://github.com/GPSBabel/gpsbabel/pull/772.  I can't remember if it was an issue with the qmake generated xcode project, a cmake generated project, or both.  In any event, the cmake xdode generator, the cmake visual studio generator, and CLion and Qt Creator use of CMakeLists.txt need to be checked to see if this influences the project view.

At the same time it would be good to verify that we don't drop dependencies on the headers in the above use cases.

jtojnar:
I compared the build directories for default (Unix Makefiles) and Ninja generators, and there were no meaningful changes. In fact, neither of them mentions `.h` files at all, even if they are listed in sources, from which I assumed CMake does the same [thing as Meson](https://mesonbuild.com/FAQ.html#do-i-need-to-add-my-headers-to-the-sources-list-like-in-autotools) and including headers in sources is pointless, but I was not able to find any documentation for that.

I am unable to run the Xcode or VS generators, they are probably platform specific. Nor do I have access to the proprietary IDEs.

I tried Qt Creator and it does looks like it indeed does not show headers unless explicitly added as sources:
Project view without headers in sources

Cf. with master branch:

Project view with headers in sources

But that sounds more like a bug in Qt Creator, as mentioned in https://stackoverflow.com/questions/62429286/display-header-files-in-qt-creator-using-cmake#comment110424508_62437291.

tsteven4:
I checked Xcode using the cmake Xcode generator.  As is the header files are nicely included in alphabetical order in a folder "Header Files", alongside "Source Files".  When I test your PR the "Header Files" folder doesn't exist, and the header files aren't visible anywhere in the Project navigator.  The project does build under Xcode with your PR, but the navigator isn't as useful, e.g. to select a header file to edit.  At least Xcode does realize that it needs to recompile osm.cc when osm.h changes.

So while your PR offers simplified maintenance of CMakeLists.txt, it makes it harder to use Xcode.  I suspect there are similar problems with other IDEs.  Is including the HEADERS in the target source causing you some difficulty?

jtojnar:
I wanted to add support for a new GPS device (Transystem GL-770) but currently the source code is completely incomprehensible with everything in the top-level directory. So I wanted to split it out into separate directories (`src/filters/`, `src/fmts/`, `src/fmts/devices/`, `third-party/`) to make a sense of it. And rather than dealing with the headers, which are redundant I decided to just drop them from sources.

But looks like many IDEs are too stupid to discover sources on their own so until they are fixed, we might need to keep that redundant information (or maybe somehow query the compiler and add the headers to sources dynamically). I opened a [bug for Qt Creator](https://bugreports.qt.io/browse/QTCREATORBUG-28289).

tsteven4:
I am closing this PR without merging.  This is because the HEADERS allow various IDEs to present a better view of the project.

We have held off on reorganizing the tree because it makes examining history more difficult, and we spend a lot of effort doing that when analyzing issues.  We may revisit this after we get down to one build system.

We would like to see some general interest for a new format, in your case Transystem GL-770.  We are reluctant to take on the maintenance if the user base is too small.

format_skelteon.cc was the starting point for a new format, but maybe a bit out of date.  Subclass Format, do not subclass LegacyFormat.  create two files newformat.cc and newformat.h.  Please organize newformat.h along google style guidelines  (public section, private section, each with ordered sections (Types, Constants, Member Functions, Data Members).  modify vecs.cc, vecs.h to get it linked in.  At this moment we are supporting qmake(GPSBabel.pro) and cmake(CMakeLists.txt), so those files will need to know about a new format.

PR #928, opened 2022-10-06
build: Allow building xmldoc in CMake binary dir

jtojnar:
Also factor out generating xmldoc fragments into a separate CMake target. We need to symlink the source files into the build directory for the build to find them and having a single place to do that is cleaner.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -489,22 +489,36 @@ if(UNIX AND NOT APPLE)
                     USES_TERMINAL)
 endif()
 
-if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
-  set(WEB "../babelweb" CACHE PATH "Path where the documentation will be stored for www.gpsbabel.org.")
-  add_custom_target(gpsbabel.org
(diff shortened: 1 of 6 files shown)


tsteven4:
Thanks.  Some efforts have been made to allow documentation to be created in CMAKE_BINARY_DIR, but they hadn't landed yet.  It will be interesting to see how you solved this.

I think this will break our qmake flow.  We are preparing to deprecate and eventually remove support for qmake, but we aren't there yet.

jtojnar:
I have not even noticed there are qmake project files in the directory listing (I was verifying that configure has been removed). This will definitely break building docs from there. 

Maybe we can remove the docs building from qmake and build the website using CMake? If someone cannot use CMake for some reason, they can always just pick up the docs on the website.

tsteven4:
@jtojnar thanks for this.  I appreciate your ideas.  
I found a few issues
1) in-source builds
2) the build directories and links can be needed before makedoc is run.
3) the removal of the autogen subdirectory makes cleaning in-source builds more difficult, and doesn't play nice with the existing .gitignore.
4) universal support for symlinks.

These issues could be overcome, but as an alternative please see #936

jtojnar:
Closing in favour of #936.

PR #936, opened 2022-10-21
allow document creation for out of source builds

tsteven4:
This is an alternative to #928.  @jtojnar please see if this works for you.

This implementation works by using absolute paths for system identifiers.  These are setup by configure in CMakeLists.txt based on CMAKE_SOURCE_DIR and CMAKE_BUILD_DIR.

It works for in source builds and out of source builds.
It maintains the autogen directory which makes cleaning in-source builds easier.
It almost works with funny characters in the path, like a space character, except for a fop bug (https://bugs.launchpad.net/ubuntu/+source/fop/+bug/1993853, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1022197)

It does not work with qmake, but qmake is about gone (#935) .

The contents of the make_gpsbabel_*.sh scripts, that made things easier when we had multiple build systems, have been incorporated directly into CMakeLists.txt.

If the CMAKE_SOURCE_DIR or CMAKE_BINARY_DIR contains characters like spaces the uri encoding is necessary for the system identifiers.  This is taken care of in perl with tools/uriencode.

Another alternative is to add rewriteSystem rules for autogen, formats and filters that correct the relative paths in the system identifiers based on the build location.  This is a standard approach.  Unfortunately it is problematic because there isn't a good way to find the default catalog.  The default catalog normally is /etc/xml/catalog, but can be /usr/local/etc/xml/catalog for homebrew, or /opt/local/etc/xml/catalog for mac ports.

Even without adding more catalog rules, using homebrew requires care, it requires special modifications to the PATH so xmlwf, xmllint, xsltproc are all found, and setting XML_CATALOG_FILES so the docbook and docbook-xsl catalogs are found.  I'll try to capture the recipe in CI once qmake goes.


jtojnar:
Cannot test now but the diff looks fine at a glance.

By the way, you might also want to switch to XInclude to get rid of the weird DTD mess. Something like:

```patch
--- a/xmldoc/readme.xml
+++ b/xmldoc/readme.xml
@@ -5,16 +5,21 @@
 [
   <!ENTITY % parts SYSTEM "autogen/_parts.xml">
   %parts;
-  <!ENTITY % chaps SYSTEM "chapters/_chapters.xml">
-  %chaps;
 ]
 >
 
-<book id="GPSBabel_Documentation_Book">
+<book id="GPSBabel_Documentation_Book" xmlns:xi="http://www.w3.org/2001/XInclude">
   <bookinfo>
     <title>GPSBabel Documentation</title>
   </bookinfo>
 
-  &allchapters;
-
+  <xi:include href="chapters/preface.xml" />
+  <xi:include href="chapters/build.xml" />
+  <xi:include href="chapters/use.xml" />
+  <xi:include href="chapters/formats.xml" />
+  <xi:include href="chapters/filters.xml" />
+  <xi:include href="chapters/datums.xml" />
+  <xi:include href="chapters/garmin_icons.xml" />
+  <xi:include href="chapters/styles.xml" />
+  <xi:include href="chapters/glossary.xml" />
 </book>
```

tsteven4:
> By the way, you might also want to switch to XInclude to get rid of the weird DTD mess. Something like:

thanks.  and convert to docbook 5.

tsteven4:
xml and uri escaping arbitrary paths on arbitrary systems is non-trivial, and brings additional dependencies.  I killing this based on the KISS principle.  If we want to support out of source document builds (with single configuration generators) it is pretty easy to do so with cmake -E copy and copy_directory commands.

PR #940, opened 2022-10-30
update documentation to DocBook 5.0

tsteven4:
This PR updates the documentation from DocBook 4.5 (10/2006) to DocBook 5.0 (released 2/2008)
Methodology:
The conversion relied on the db4-update.xsl file provided by DocBook.  The exact process is captured in xmldoc/convertdoc.
Validation:
gpsbabel.pdf before and after this PR were compared with diffpdf.
gpsbabel.html before and after was saved as a pdf from firefox using "Microsoft Print to PDF", and those pdf files were compared with diffpdf.

A number of conversion issues were found and corrected using the above process.
1) the DocBook 4.5 productname element had a default class of trade, and rendered the elements with the trade mark sign.  DocBook 5.0 did not annotate productname elements with the trade mark sign.  The elements were changed to trademark elements to match the previous document.
2) the db4-update.xsl style sheet from the DocBook 5.0 release dropped the "Appendix A." from the titles "Appendix A. Supported Datums", etc. in gpsbabel.html.  This was fixed in the DocBook 5.1 style sheet.
3) the db4-update.xsl style sheet from the DocBook 5.0 releases dropped the index letters in the glossary. This was fixed in the DocBook 5.1 style sheet.
4) the DocBook 5.1 style sheet dropped the link portion of term/ulink elements.  We have two such elements, the link was adding back in using a patch.

There were issues using the entity replacement scheme for the individual bits of formats and filters.  The bits from formats/*.xml, formats/options/*.xml, filters/*xml and filters/options/*.xml are now copied by makedoc into autogen/fmt_*.xml and autogen/filter_*.xml.

This results in gpsbabel.html files that compare, and one miscompare for gpsbabel.pdf.  The miscompare is due to a different font that is used for uri elements as opposed to link elements, which are both derived from the original ulink elements.  We have two uri elements, and the one geocaching glossary entry is detected.  This difference was judged to be acceptable, and perhaps even desirable.

DocBook 5.0 uses the schema langauge RELAX NG, which is reportedly not fully implemented with xmllint.  We switched the validation tool to jing.

The documents are being prepared in CI on both ubuntu and macos.


tsteven4:
In the future I suggest we convert gpsbabel.html documentation to html 5.  We may also want to convert the gpsbabel.org version for smarty to html 5 as well, although I cannot test that result at all.

robertlipe:
Wow, you looked under a lot of refrigerators when tackling this. There are a whole lot of technologies I don't fully understand you had to touch. Thanx!

Approved.

robertlipe:
Probably, yes.

There was some ceiling I hit with Smarty where I couldn't trivially update
it past a certain version. I need to come back and tackle that.

PR #944, opened 2022-11-06
enhance faketime option to allow subsecond steps.

tsteven4:


robertlipe:
This is almost identical to what I scratched out. (I used a qint64 instead of a long long, but I can't imagine that ever mattering in the real world.)

Go for it, please.

PR #948, opened 2022-11-23
fix errors handling optional Waypoint fields.

tsteven4:
1. Hide wp_flags from formats by using WAYPT_SET.
2. Don't print values corresponding to Waypoint fields that aren't valid.
  a) if the field uses wp_flags, then the flag must be set.
  b) power, cadence heartrate must not be zero.

This is in anticipation of use of std::optional instead of wp_flags to indicate an optional value is present.
The use of invalid power, cadence, and heartrate values was found fortuitously.

Note this changes kml output, using empty gx:value elements to correspond to missing data.  gx:value elements are of type string, the empty string do not violate the schema (unlike empty when elements).  However, it is unclear who is using this extra SchemaData so we cannot check to make sure they acutally handle empty string values.


robertlipe:
In the early days, it was worth having the "have_foo" components in a
bitfielld, but std::optional is. a good direction for this these days.
Encouraged.

@Viet-tam Luu ***@***.***> , I think the ExtendedData specs say
this shouldn't matter. Do you agree?

We've danced around ExtendedData not being available widely (Google Maps,
ahem) since around Earth 4.2 or so, so about 12+ years ago when KML was
still congealing. Can we just stop it with creating the humungo
preformatted <html> <table> and have a reasonable expectation that anything
that reads KML can handle ExtendedData now? At the other pole, is that
supported in Earth >= 9 now/yet?

Changing our output style for KML is a different piece of business than
this CL, but is it safe to move that direction? Baking units and time zones
and formatting (oh, so much
<https://github.com/GPSBabel/gpsbabel/blob/c21589e041e89036824bce033cd59fa89a12224d/reference/track/gpx_garmin_extensions-kml_track.kml#L185>
formatting for each point) into those HTML balloons really needs to go. Can
we get that cleared?

RJL

viettaml:
> I think the ExtendedData specs say this shouldn't matter. Do you agree? We've danced around ExtendedData not being available widely (Google Maps, ahem) since around Earth 4.2 or so, so about 12+ years ago when KML was still congealing. Can we just stop it with creating the humungo preformatted `<html>` `<table>` and have a reasonable expectation that anything that reads KML can handle ExtendedData now?

I can't speak for "anything that reads KML" but certainly Google Earth handles KML ExtendedData just fine (granted, it's _displayed_ as an HTML table in placemark balloons, but in the underlying KML it's all `<Schema>` and `<ExtendedData>` etc.). In fact we rely on it extensively in our new GeoJSON import feature, and we recently [addressed a long-standing issue](https://github.com/libkml/libkml/pull/272) of KML Schemas not working with our Regionation tool.

robertlipe:
>
> I can't speak for "anything that reads KML" but certainly Google Earth
> handles KML ExtendedData just fine (granted, it's *displayed* as an HTML
> table in placemark
>
> OK, so for our first question, empty vs. null extendeddata sounds covered.
That's surely covered ground in any ExtendedData reader. How's the KML
compatibility in Earth 9 these days?

I have the (unbacked by data) hunch that Maps and friends reading KML just
matters less now than it used to.  Are you OK with us moving our big
<table> makers to ExtendedData in some reasonable future?

Are there still multiple slightly incompatible schemas in this area around
OGC KML vs. Earth KML? gx:Track was...contentious. Did that get coalesced
down? What set of 2.1, 2.2, and 2.3 matter in practice these days?

I think we (GPSBabel & GPS tracks) were one of the big proving grounds for
both ExtendedData and formatting data via HTML - for better and worse.
Still, that was a very long time ago. Some of that probably goes back to
'05 or '06.

balloons, but in the underlying KML it's all and etc.). In fact we rely on
> it extensively in our new GeoJSON import feature, and we recently addressed
> a long-standing issue <https://github.com/libkml/libkml/pull/272> of KML
> Schemas not working with our Regionation tool
>
>
Funny. I was the one that built Regionator - and also made the "better and
worse" decision to use libkml, (then "bactrian".  Regionator was a proving
ground for that tech, too. The plans to collapse those two KML compositors
(three if you count GPSBabel's - which has been completely replaced since
then) didn't exactly play out.

Surprised to see scoped_ptr and intrusive_ptr and other boost-isms have
survived, but it's not clear how "live" this code really is these days.

Thanx for the updates.

tsteven4:
thanks.  I now see GE uses the extended data as @viettaml mentioned.  And behavior with missing values is documented https://developers.google.com/kml/documentation/kmlreference#sparse-data, which notes that this applies to ExtendedData.  The 0 values we used to do aren't recognized as missing data, but the empty elements are.  The documentation states 
"Any element except <when> can be empty and will be interpolated between the nearest two well-specified elements."  This seems to be true for the GE generated balloons, thus the GE behavior improves with this change.

The balloons we produce are more complete than the GE generated balloons, e.g. they include the time.  I don't think the automatically generated GE balloons are a replacement for the gpsbabel generated balloons.

GE also uses the extended data in the elevation profile, which is quite a nice feature.  However, unlike the balloon elements where missing data seems to be interpolated as documented, when SimpleArrayData has missing points the array isn't included in the elevation profile at all.  This would seem to be a GE bug.  Previously, when we used 0 for missing data, all arrays were included in the elevation profile, and the 0 elements showed up as zeros.  @viettaml is this a known GE bug?  can it be fixed?  I'll attach a short test case.
[missing_data_elevation_profile_tests.zip](https://github.com/GPSBabel/gpsbabel/files/10096547/missing_data_elevation_profile_tests.zip)

PR #950, opened 2022-11-27
add power to kml track descriptions and balloons.

tsteven4:


robertlipe:
Nice. Thank you.

I'm not sure if realtime fitness is a thing, but it's so obvious and so
simple we may as well have it.

Power and the cadence sisters were added when we did terrain track
profile[1] back in Earth 5 (4.3?) or so and was done in a private branch of
GPSBabel while the trunk was evolving so we treaded pretty lightly on the
sweeping changes.

This was our clear intent. Thanks for delivering it.

I've long had some tension in whether the fitness (and geocaching) stuff
should be moved into format-specific data or subclasses, but it's been
years since we've had reports if people running out of memory, so maybe
counting bytes in that struct isn't necessary these days.  Having been
around for 20+ years, our typical systems have a thousand times the memory
(32MB->32GB), but our GPS data sizes haven't similarly grown.

I have kicked around an analogous format specific data structure for auto
track data, tracking individual wheel speeds, gforces, etc.. as collected
by Teslas in track mode or a Draggy or Racebox or such. It really needs an
auto enthusiast or sponsor interested in "driving" it and maybe some
interested users.  For now it's just kind of a pet project in an obscure
working directory.

[1] Years later, on one of my final trips to CA, my AirBNB was a few
thousand feet from where that sample track was collected. It was cool to
recognize it from the sample track. It's a harsh climb. The cyclist turned
around at Robert's Market. :-)

PR #956, opened 2022-12-07
Drop incomplete html_encrypt option in kml writer

robertlipe:


tsteven4 (approved):
This looks equivalent to what we have always done given that html_encrypted was always zero. 

The order of operations of logpart->cdata.toHtmlEscaped() is a bit cryptic, the member operators . and -> have equal precedence, and associativity is left to right so it does what we want (and what will compile).  I think this pattern is common in our code, I don't think parenthesis are called for.

Issue #957, opened 2022-12-07
Deprecate use of sprintf throughout


robertlipe:
Building on MacOS Monterey, we get a few dozen warnings of the form:

/Users/robertlipe/src/gpsbabel/jeeps/gpsprot.cc:397:10: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]

Upon review none of our uses were actually problematic writes into buffers of unknown size, but it's indeed good taste to just not do that.

As many of these warnings are in icky intersections of null-terminated C strings and (worse) fixed-length character arrays or are in code that's hard to test without wiring up ancient equipment, I took a mixed approach to fixing these. When I felt good about code coverage, I tried to move things to QStrings and use QString::asprintf() which gives a nice backward-compatible bridge that's not terribly disruptive. When the code was in "very C" files that otherwise didn't really know about QStrings, I just changed sprintf(foo, ...) to snprintf(foo, sizeof(foo).

duplicate.cc was writing the buffers twice. Removed the first stores.
garmin was reusing tbuf several times to copy into qstring. Moved to ::asprintf except in scary code with fixed, non-terminated arrays.
gpsapp.cc, gpscom.cc, gpsprot.cc - path of least resistance
mkshort.cc - fragie code
mtklogger . movign speed is covered in tests.  Satellites in sight is not, but this code sure feels better.
tpo.cc - whitespace. Those debug printfs should probably be deleted.
wbt-200.cc - poor coverage. Mechanical fixes.

robertlipe:
See candidate PR https://github.com/GPSBabel/gpsbabel/pull/958

PR #958, opened 2022-12-07
Remove all uses of sprintf

robertlipe:
See https://github.com/GPSBabel/gpsbabel/issues/957


wbt-200.cc
```diff
@@ -551,10 +551,9 @@ static int check_date(uint32_t tim)
 
 static Waypoint* make_point(double lat, double lon, double alt, time_t tim, const char* fmt, int index)
 {
-  char     wp_name[20];
   auto* wpt = new Waypoint;
 
-  sprintf(wp_name, fmt, index);
+  auto wp_name = QString::asprintf(fmt, index);
```
tsteven4: better than auto, move the QString::asprintf to the assignment to wpt->shortname on L562.

I note that some of these additions of QString::asprintf do not involve an extra round trip through codecs, and this is one of them.  We do the conversion to QString in QString::asprintf instead of implicitly when assigning to shortname.  These cases are all goodness compared to the base.  Cases involving converting from some encoding to UTF16 and then back to UTF-8 (by default) are a bit more questionable.

tpo.cc
```diff
@@ -687,7 +690,10 @@ static void tpo_process_tracks()
 
     if constexpr(debug) {
       printf("Track Name: %s, ?Type?: %u, Style Name: %s, Width: %d, Dashed: %d, Color: #%s\n",
-             qPrintable(track_name), line_type, qPrintable(styles[track_style].name), styles[track_style].wide, styles[track_style].dash,rgb);
+             qPrintable(track_name), line_type,
+             qPrintable(styles[track_style].name),
+             styles[track_style].wide, styles[track_style].dash,
+             qPrintable(rgb));
```
tsteven4, line 696: double conversion, but with the knowledge about what we are converting it should be safe.

mtk_logger.cc
```diff
@@ -801,16 +801,17 @@ static int add_trackpoint(int idx, unsigned long bmask, struct data_item* itm)
...
     if (mtk_info.speed > 0) {
-      sprintf(spds, " when moving above %.0f km/h", mtk_info.speed/10.);
+      spds = QString::asprintf(" when moving above %.0f km/h", mtk_info.speed/10.);
     }
-    trk_head->rte_desc = QString::asprintf("Log every %.0f sec, %.0f m%s"
-                                           , mtk_info.period/10., mtk_info.distance/10., spds);
+    trk_head->rte_desc = QString::asprintf("Log every %.0f sec, %.0f m",
+                                           mtk_info.period/10.,
+                                           mtk_info.distance/10.);
+    trk_head->rte_desc += spds;
```
tsteven4, line 814: LGTM.  no extra conversion, just doing it in a different place.

garmin.cc
```diff
@@ -666,12 +666,10 @@ lap_read_as_track(void)
       wpt->creation_time = array[i]->start_time;
       wpt->microseconds = 0;
 
-      sprintf(tbuf, "#%d-0", index);
-      wpt->shortname = tbuf;
-      sprintf(tbuf, "D:%f Cal:%d MS:%f AH:%d MH:%d AC:%d I:%d T:%d",
+      wpt->shortname = QString::asprintf("#%d-0", index);;
```
tsteven4: I think there is only one use of tbuf left.  If so can we move the declaration to just before its use at L655, and change the size appropriately?

mtk_logger.cc
```diff
@@ -1073,23 +1074,21 @@ static int csv_line(gbfile* csvFile, int idx, unsigned long bmask, struct data_i
...
-        slen += sprintf(&sstr[slen], "-%.2d", itm->sat_data[l].azimut);
+        s += QString::asprintf("-%.2d", itm->sat_data[l].azimut);
       }
       if (bmask & (1U<<SNR)) {
-        slen += sprintf(&sstr[slen], "-%.2d", itm->sat_data[l].snr);
+        s += QString::asprintf("-%.2d", itm->sat_data[l].snr);
       }
 
-      gbfprintf(csvFile, "%s%s", do_sc?";":"", sstr);
+      gbfprintf(csvFile, "%s%s", do_sc ? ";" : "", CSTR(s));
```
tsteven4: unfortunately this adds a roundtrip to/from QString, albeit with data that should be safe for most encodings.

tsteven4 (commented):
I had a few suggestions for improvements.  When your happy go for it.

wbt-200.cc
robertlipe: Done.

Agreed. This whole patch was another archealogical dig trying to put a Starbucks into a 15th century shrine at times.

tpo.cc
robertlipe: It was a debug print that's not even compiled in most of the time. If I touch this line again, it becomes a Debug() as all the format specifiers are the "easy" ones anyway. 

tpo is hanging on by a thread because we have one interested user.

garmin.cc
robertlipe: I took it a step further. Hard to test, as is most of this file.

mtk_logger.cc
robertlipe: How do you like this instead? gbfputs so it handled the utf8 conversion explictly. There's no need for do_sc. It's all really heavy for a glorified ascii 7-bit stream, but it's at least internally consistent.

Streaming gbfile operators that just accepted objects would be nice, but that's a project we'd never earn back our development time and we'd end up with both vaarg and streaming variants, which seems the worst of both worlds.

Over our years, we've kind of cycled between strong isolation of tasks and monster helper functions that handle (formatting, I/O, compression, memory mapping, piping, etc.) all in one.

PR #960, opened 2022-12-26
move xml tag utilities to xml_tag.cc

tsteven4:
the utilities are already declared in src/core/xmltag.h, but are presently defined in util.cc.
move xml_tag.cc into src/core next to it's h file.


robertlipe (approved):
Looks Good To Me. Thanx.

I like that we're slowly whittling away at util and defs.h

tsteven4:
> I like that we're slowly whittling away at util and defs.h

I have a bunch of work going on where I have tried to get things out of defs.h for file xxxx.cc and into xxxx.h.  But
1) it is hard to avoid circular dependencies
2) it is easy to become sensitive to the include order.

PR #965, opened 2022-12-29
Use QString::compare(QStringView, Qt::CaseSensitivity) 

tsteven4:
instead of

- case_ignore_strcmp with a QString and a character
literal.
- QString::compare(QLatin1String, Qt::CaseSensitivity)

With a raised Qt floor QString::compare(QLatin1StringView, Qt::CaseSensitivity) would probably be better.

### DIFF
--- a/garmin_device_xml.cc
+++ b/garmin_device_xml.cc
@@ -42,7 +42,7 @@ jmp_buf gdx_jmp_buf;
 
 static void type_s(xg_string args, const QXmlStreamAttributes*)
 {
-  type = args.compare(QLatin1String("GPSData"));
+  type = args.compare(u"GPSData");
 }
 
 void device_s(xg_string args, const QXmlStreamAttributes*)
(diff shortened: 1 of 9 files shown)


GPSBabelDeveloper:
It's OK to go ahead and #if it in if you like while it's hot in your mental
cache.

Raising the floor is your call. Qt 5.15 is EOL'ed in May so we're running
out the clock on it anyway.

tsteven4:
Unfortunately some major Unix distributions, e.g. Fedora, still don't have Qt6WebEngine.  In related news my fix for Qt6Core on Ubuntu from May finally made it into the Pending Ubuntu Stable Release Updates queue (hopefully it will move along soon.)

I'm going to pass on the conditional code even though your suggestion is probably the right thing to do.  Even without QLatin1StringView I think this is considerably better than converting to QStrings as case_ignore_strcmp does.  I also think it requires an explicit cast to QStringViewLiteral, instead of the implicit cast this PR uses from char16_t* to QStringView.  I'm not sure Qt6 view classes and related methods are all settled out yet.

PR #967, opened 2022-12-30
fix clang diagnostic cast-qual

tsteven4:
All but 2 of these didn't need to drop the const qualifier at all.  These two, in gui/gpx.h, now use const_cast to cast away const.

gpsusbcommon.h L156, gpsusbwin L113 didn't need a cast at all.  They also don't need to force array decay, but the implementation in this PR allows the assembly listing of gpsusbcommon, gpsusbwin and jgpsutil to all match the base.
gpsusbwin had a number of (GUID*) casts were unnecessary.


robertlipe:
I know the promise is "you shouldn't need to" but, man, C++ casts are just
so ugly.

LGTM. Thanx.

PR #968, opened 2022-12-31
encapsulate geocache handling in class Geocache

tsteven4:


GPSBabelDeveloper:
This has so many ideas (and more) that I started and never finished. Thanx
for the major cleanup!

Geocaches are almost a "format specific" data for us (ignoring the name of
that concept briefly) but they don't seem to be a great fit, plus that
linear search on each access is a drag.

I've thought before about ways to remove those linear searches in type_map,
container_map  and friends. You used QVector which has indexOf() which
might concievably build an index behind our back and use it (it doesn't
<https://github.com/radekp/qt/blob/b881d8fb99972f1bd04ab4c84843cc8d43ddbeed/src/corelib/tools/qvector.h#L673>)
but that got me thinking about other container types. These are kind of
funky because we'd like fast access for indexing from both the key and the
value. Other than maintaining two separate, inverted QMaps (which, I
suppose, isn't OUT of the question, though I'll admit I *just* thought of
it...) do you know of a better container pattern for that? QMap at least
builds an internal binary tree of the structures by keeping the keys sorted
at insert time. (Not asking for any change here. Just volleying ideas for
now...) The QMap interface would look pretty good at the callers (container
= containers[container_type]) but gets really ugly when you're doing that
inverse operation and trying to get the container_type enum from the
string. How do other apps build and maintain such mappings?  Then again, I
guess a sequential trip through a dozen entries, where everything will
quickly land in L1 cache, isn't exactly totally terrible these days. This
is probably the low boundary where excessive cleverness is warranted anyway.

Now that you've implemented namespaced enums for everything (thank you!)
would you be in favor of me removing the C-style prefixes where geocache
container things are prefixed "gc_" and types are "gt_" and such? That's a
quick exercise in CLion (or Sublime) for me, I think, if you agree that's
an improvement.

I'm pleased you finally organized this so. Thanx!

RJL

...

tsteven4:
I thought about the mappings but in the end, I moved on reasoning that if the searches had been good enough until know they probably still are.  Be aware both maps are not 1:1, i.e., there are repeated type and container values.  The good news is the maps are fairly short.

I wanted to remove the gc_, gt_, gs_ under prefixes, but resharper failed to do the refactoring, leaving unchanged names behind in various files.  Also a few of the names would be reserved, e.g., true, false, virtual.  I didn't try clion or xcode.  It sure is nice to be using an IDE to make these changes, with all the annotations and links to related bits.  I have been unable to get visual studio to sign a commit though.

PR #971, opened 2023-01-04
remove remaining shadows of cet

tsteven4:
cet was reduced to setting up global_opts.codec and verifying that codec existed (cet_convert_init).  The only users of global_opts.codec were garmin, which was the only user of garmin_fs_garmin_after_read and garmin_fs_garmin_before_write, and garmin_gpi.  The uses were from the macros STRTOUNICODE and STRFROMUNICODE.

I moved garmin_fs_garmin_after_read and ...before_write from garmin_fs.cc to garmin.cc.  There are subtle changes in the encoding/decoding of the strings in those routines.  I believe they function equivalently, but we have no test coverage.  I believe our assumption was and is that these strings in the way are fixed length and zero filled, and may not have a null terminator.  When converting to unicode we used to create a fixed length byte array possibly containing nulls and then use constData to force QTextCodec::toUnicode to find either an included NULL or the NULL QByteArray automatically provides.  Now we use qstrnlen to created a byte array either of the characters before the first null or the length of the source, and pass that QByteArray to QTextCodec::toUnicode.

There were some extra byte array/c string conversions in GarminGPIFormat::write_string.  From the conversion we had a byte array, but we passed it as const char*, forcing write_string to count again.

I believe we do have encoding/decoding issues in garmin, as noted in the original and updated comment about receiver_charset, and nmea, and igc. I think the nmea and igc writers can use illegal characters, and the readers fail to detect illegal characters.  For IGC I think the spec might be https://www.fai.org/sites/default/files/igc_fr_specification_2022_with_al7_2022-1-31.pdf (see A6 which defines a subset of ASCII printable characters).  For nmea I believe only printable ASCII characters are allowed.  I think gdb and garmin_gpi might have been and remain free of encoding/decoding issues.

This means we can get rid of Format::get_fixed_encode and Format::get_encode, and their legacy equivalents.
I confess I don't understand why Format::get_fixed_encode (or ff_vecs_t.fixed_encode) existed, but I don't think it has been doing anything for years.  If this survives review we can chop them.


tsteven4:
As an aside, I noticed there is an implicit conversion from QByteArray to const char* and const void*.  This isn't mentioned in the list of QByteArray public functions, static public members, or related non-members.  It is mentioned that you can turn this off in the macro QT_NO_CAST_FROM_BYTEARRAY.  When we define that macro it shows two uses unrelated to this PR.  In general we have explicitly used constData().

tsteven4:
I have the removals of Format::get_fixed_encode and Format::get_encode, and their legacy equivalents queued, which touches 45 files, but I don't want to dilute any review.  @GPSBabelDeveloper let me know if you have any comments.

PR #973, opened 2023-01-06
remove encoding, fixed encoding data from formats.

tsteven4:


robertlipe:
This patch makes me happy. There was a lot of yarn to pull at for us over a long time to make this go away. Thanks!

PR #974, opened 2023-01-06
another use of potentially invalid waypt field.

tsteven4:


robertlipe:
Should those become optionals or at least some wrapper class that's explicit and doesn't show reads as a unary bool?

tsteven4:
Implementation of public member functions that mimic std::optional, and making the related fields private, is how I found this.  std::optional is more straightforward, but significantly increases the size of Waypoint.  Stay tuned, it is almost there.

GPSBabelDeveloper:
Awesome!

Too many cppcon 2022 videos last night...


We have many of these flagged members, butt we rarely have them all at
once. Maybe we could carry a map of type+char (shorts would cost another
byte of padding, but things like heading need more than a byte) std:: pairs
of something. We'd pay a little in runtime lookups, but it's likely our net
space cost would be lower.

Lets both kick that around a bit.

PR #975, opened 2023-01-06
mimic std::optional for flagged Waypoint fields.

tsteven4:
This implementation preserves the use of wpt_flags, which saves considerable space for every Waypoint instance compared to std::optional.

The Waypoint fields that use wpt_flags are made private to force usage of the member functions that access them.  This should guarantee that the wpt_flags are kept in sync with the data fields, and unset data fields are not used.


tsteven4:
sorry about unassigning @robertlipe, I thought I assigned @robertlipe instead of adding @GPSBabelDeveloper as a reviewer.

defs.h
```diff
@@ -405,27 +406,27 @@ class Waypoint
   float hdop;
   float vdop;
   float pdop;
-  float course;	/* Optional: degrees true */
-  float speed;   	/* Optional: meters per second. */
   fix_type fix;	/* Optional: 3d, 2d, etc. */
   int  sat;	/* Optional: number of sats used for fix */
```
robertlipe, line 480: Can't most of these (dop, cadence,  ⁠ಠpower, heart rate, etc).use the same scheme?

robertlipe:
It's fine.. Obviously, I'm both, just depending on what I'm doing. Here I use them interchangeably.

I like this. It extends the syntax and checking of what we already had, where my map idea would have lots of ugly fallout.


Looks good to me. Thanks!

defs.h
tsteven4: we aren't consistent.  some things assume zero indicates invalid.  I think heart rate, cadence and power all use this convention.  It looks like the *dops used >0 or ==0.  we could convert them to use the new macros for consistency, it shouldn't cost too much as the macros are all inline (as they are defined in class).  perhaps this is best left to another PR, I didn't even review the changes in this one myself yet, I was rushing to get a draft out you could look at.

garmin_txt.cc
```diff
@@ -711,15 +711,15 @@ track_disp_wpt_cb(const Waypoint* wpt)
...
   }
 
   if (prev != nullptr) {
     *fout << "\t";
     delta = wpt->GetCreationTime().toTime_t() - prev->GetCreationTime().toTime_t();
-    float temp = WAYPT_GET(wpt, temperature, -999);
+    float temp = (wpt->temperature_value_or(-999));
     if (temp != -999) {
       print_temperature(temp);
     }
```
tsteven4: shouldn't the printing of temperature, and it's tab, be outside the if statement?

tsteven4:
One issue is wp_flags is still public.  There are other bit fields that need to be public, but it would be best if the has_value bit fields were private.  We could split it into a private and public member, or leave it alone.

There are some questionable things going on in various garmin routines.
garmin_txt: uses value_or and then tests against the or value.  This mimic what we have always done, but it seems we should just use has_value.  Additionally, the printing of temperature as noted above seems wrong.
gdb: same thing, but more disguised.

tsteven4:
I split the optional bitfields into another class so they can be private.
I moved a wp_flags to manual tune packing.  This results in the sizeof(Waypoint) staying at 280 bytes on ubuntu x86_64.

PR #978, opened 2023-01-11
fix synthetic shortname option handling

tsteven4:
and make related options parameters bool.

the ozi and xcsv writers had ARGTYPE_BOOL options without defaults
for must upper, must unique and whitespace ok, but only used them
if they were not nullptr.
For ARGTYPE_BOOL without defaults false is signified by a value of nullptr.
But mkshort defaults whitespace ok and must unique to true, so they were
impossible to turn off.

add test for the above writers and the related options.


tsteven4:
This was fixed in #1351

PR #979, opened 2023-01-12
Update kml-prec.xml

jidanni:
I am guessing, based on the same sentence in gpx-elevprec.xml, what you meant to say.


tsteven4:
Thanks for pointing out the problem.  The option actually refers to latitude and longitude values, not altitude values.  I have made the correction to the correction.

PR #986, opened 2023-01-16
macOS brew install fix

tsteven4:
skipping the brew update/upgrade speeds up the brew step about 10x, and avoids the link issues.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -58,22 +58,16 @@ jobs:
       run: |
         ./tools/travis_install_osx ${{ matrix.QT_VERSION }} aqt
 
-    - name: Fixup brew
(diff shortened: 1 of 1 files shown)


robertlipe:
Wow. When I committed this, I chased it through scores of Github issues,
referencing other issues and the one I chose was the hot solution at the
time. Looks like it's ping-ponged all over dozens of projects, each playing
hot-potato with hacks and "ultimate fixes".

Will continue to monitor. You have the right call here for now. It's not
like not tracking the daily updates of these packages is going to cost us
any real development coverage.

Watching those builds trigger was where I noticed that every proposed CL
was triggering a build for each of our 35 (?) targets for every outstanding
PR we had open. That was what motivated me to integrate or drop all my
ancient pending PRs. That saved several dozen calls to the build farm on
every PR so that could only free up some system time.

Issue #981, opened 2023-01-12
"All track points have been dropped!" but end up in file anyway


jidanni:
```
gpsbabel -t -i kml -f /tmp/k.kml -x track,start=20221220204931,stop=20221220204931 -o csv -F /tmp/l.csv
trackfilter-range: All 36755 track points have been dropped!
wc /tmp/l.csv
  36757  110271 9449368 /tmp/l.csv
```
If I didn't use -F then they really would have been dropped. I am just following the example on
https://www.gpsbabel.org/htmldoc-1.8.0/filter_track.html#ex_track_startstop

robertlipe:
Does kml have place marks in addition to track points? That would explain
it. The lines in your example between the two are, well, fuzzy.

jidanni:
I see. A previous command had turned them into `<placemark>`s. Closing.

jidanni:
OK, it calls waypoints "trackpoints" perhaps.
Anyway no combination of -t or -w or -x track,start=1,stop=2023111111 will get rid of that message,
"trackfilter-range: All 36755 track points have been dropped!"

jidanni:
[Makefile.txt](https://github.com/GPSBabel/gpsbabel/files/10398307/Makefile.txt)
[FlightAware_JAL57_KSFO_RJAA_20221220.kml.gz](https://github.com/GPSBabel/gpsbabel/files/10398310/FlightAware_JAL57_KSFO_RJAA_20221220.kml.gz)
I was just trying to extract one point (one second) from the KML.

tsteven4:
What version of gpsbabel are you using? This works for me.  If you want to get rid of the waypoints leaving just the one trackpoint use the nuke filter.

./bld/gpsbabel -i kml -f FlightAware_JAL57_KSFO_RJAA_20221220.kml  -x track,start=20221220204931,stop=20221220204931 -x validate,debug -o kml,units=a,extrude=1 -F Flight.kml

Processing waypts
point ct: 2, waypt_count: 2

Processing routes
route head ct: 0, route_count: 0
total route point ct: 0, route_waypt_count: 0

Processing tracks
track 0 ct: 1, waypt_count: 1
track head ct: 1, track_count: 1
total track point ct: 1, track_waypt_count: 1

jidanni:
I also notice that "if we put the clothes back in the washing machine a second time" things change. (But no changes the  third time):
```
[...]

PR #987, opened 2023-01-17
Update trackfilter.h

jidanni:
Also mention exact matching timestamps. Proven by
https://github.com/GPSBabel/gpsbabel/issues/981#issuecomment-1380515460

### DIFF
--- a/reference/help.txt
+++ b/reference/help.txt
@@ -104,7 +104,7 @@ File Types (-i and -o options):
 	  ver                   Version of gdb file to generate (1..3)
 	  via                   (0/1) Drop route points that do not have an equivalent w
 	  dropwpt               (0/1) Don't create waypoints for non-user points
-	  roadbook              (0/1) Include major turn points (with description) from
+	  roadbook              (0/1) Include major turn points (with description) from 
 	garmin_txt            Garmin MapSource - txt (tab delimited)
 	  date                  Read/Write date format (i.e. yyyy/mm/dd)
 	  datum                 GPS datum (def. WGS 84)
@@ -180,8 +180,8 @@ File Types (-i and -o options):
 	  trackdata             (0/1) Include extended data for trackpoints (default = 1
 	  trackdirection        (0/1) Indicate direction of travel in track icons (defau
 	  units                 Units used when writing comments ('s'tatute, 'm'et
-	  labels                (0/1) Display labels on track and routepoints  (default
-	  max_position_point    Retain at most this number of position points  (0
+	  labels                (0/1) Display labels on track and routepoints  (default 
+	  max_position_point    Retain at most this number of position points  (0 
 	  rotate_colors         Rotate colors for tracks and routes (default autom
 	  prec                  Precision of coordinates, number of decimals
 	land_air_sea          GPS Tracking Key Pro text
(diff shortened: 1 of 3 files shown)


tsteven4:
Thanks. You are correct
```
        bool after_start = !start.isValid() || (wpt->GetCreationTime() >= start);
        bool before_stop = !stop.isValid() || (wpt->GetCreationTime() <= stop);
```


 I tweaked the wording a bit, and updated the impacted reference file.

reference/help.txt
```diff
@@ -104,7 +104,7 @@ File Types (-i and -o options):
 	  ver                   Version of gdb file to generate (1..3)
 	  via                   (0/1) Drop route points that do not have an equivalent w
 	  dropwpt               (0/1) Don't create waypoints for non-user points
-	  roadbook              (0/1) Include major turn points (with description) from
```
robertlipe, line 107: How did all these trailing whitespace instances come back? Can we please not do that?

tsteven4: They come back because disp_formats, disp_filters, disp_vecs, and disp_filter_vecs produce them.  When I update the reference files I copy the exact file produced from TMPDIR to REFERENCE, and then make sure the changes are appropriate.

GPSBabelDeveloper:
Oh. They're actually trailing tabs because we generate, say, 9 fields that
may be separated by tabs, but the fields may be blank. So the tradeoff is
that if you split on tabs and look at the 8th field, you'll always get a
"nothing" but if we wrote only four fields and split on the 8th field, we'd
have a sad.

0000d100  73 69 6f 6e 28 73 29 09  62 6f 6f 6c 65 61 6e 09
 |sion(s).boolean.|
0000d110  *09 09 09* 68 74 74 70 73  3a 2f 2f 77 77 77 2e 67  |...
https://www.g|
0000d120  70 73 62 61 62 65 6c 2e  6f 72 67 2f 68 74 6d 6c  |
psbabel.org/html|

Looking at the code, it's clearly "mine", no matter what git blame says, so
that's either my fault or my brilliant design and I just forgot/didn't
consider this case when looking at those diffs.

I'm so glad you're here. :-)

Carry on.

Should I make something in tools/ that regenerates all those files that
serialization.test likes to fondle, including de-fiddling the version
number, so that integrating changes like Dan's become a little easier?

Fixing upgrade_check is on my list for this week. To my (un)amusement,
something changed on my hosting company on the 8th, causing catastrophic
failure, but worse still, the error isn't returning the language's
construct of a null (e.g. (void*) 0 or __nullptr__) but is instead
returning an array of five characters: 'N', 'U', 'L', 'L', '\0' - which is
the dumbest behaviour I can imagine. I think the underlying language got a
security update, so i need to find and fix both the root cause and the
error path.  I have bailing wire in it now, returning an essentially
constant XML. I also need to handle the case of old doc directories
redirecting to new doc pages that no longer support that format. I've been
battling Webmaster Tools on several other long-standing fronts, too.

RJL

tsteven4:
> Should I make something in tools/ that regenerates all those files that serialization.test likes to fondle, including de-fiddling the version number, so that integrating changes like Dan's become a little easier?

I don't think it's worth it, or worth the maintenance burden of the tool.  It's pretty easy to comment the trap in testo, runt the tests and copy and paste from the ERROR message to cp the file in TMPDIR to REFERENCE, and then the usual review tools show the comparison.  I suppose if one was ambitious this could all be integrated into the github and automatically make a commit with changes to the PR, leaving them for manual review with the other changes.  That may confuse people as something else is changing "their" branch.

robertlipe:
Fair. It's not like we have hundreds of Dan-like contributors.

"Not all things worth doing are worth doing well."

PR #988, opened 2023-01-17
Enable the gpx reader to use the declared namespace prefixes for recognized namespace URIs.

tsteven4:
We have long assumed that a particular namespace URI has a known namespace prefix associated with it.  The conventional association is strong and generally allowed things to work, but it is not promised.

This assumption failed today reading a gpx file from connect.garmin.com.

This PR rewrites the hash table used by the gpx reader to recognize alternate prefixes for known namespace URIs.  Because multiple prefixes may be associated with the same URI the original hash table entries with the assumed prefixes are left in.  One of our reference files, basecamp.gpx, actually does this.  To be pedantic we should delete any hash table entries corresponding to undeclared prefixes, but I didn't bother.  I also didn't bother with humminbird extensions, although it would be easy to do so.


tsteven4:
I think this covers us for namespace prefixes declared in the gpx element, but it doesn't cover us for other cases where they are declared later, e.g. unknowntag.gpx.  I think the PR is an improvement, but there is still a hole.

tsteven4:
I'm looking at an alternative that should cover namespaces that are declared in or after the gpx element.

tsteven4:
An alternative is #990.  It covers the case where the xmlns attribute is on any element, not just the gpx element.

robertlipe:
Thanx. This is an aspect of GPX (well, XML) that's never made sense to me.
I've known we have achilles heels in this area. XML people like to say that
an XML file is self-describing, but apps still have to KNOW what's between
the angle brackets, right? Referring to a(n always broken) URL via a handle
doesn't change that, but I've known that we never allowed that binding
between handle and URL to be changed, so thank you for that. I know that a
we theoretically could have a foo:type that referred to either a geocache
type or a (fictional) fish finder type and that we SHOULD be matching on
that URL instead of of a combination of tag hierarchy and handle, but we
just haven't. In the expat days, that was all just too hard.

I suppose that gpx.h:GEOTAG has related opportunities that I should look
into, but the reality is that there's one (maybe two) producers of those
files that matter, so we just don't see a huge competing diversity in them.

Thank you for updating this.

tsteven4:
> I suppose that gpx.h:GEOTAG has related opportunities that I should look into

https://www.w3.org/TR/xml-names/#ns-qualnames:

> The [Prefix](https://www.w3.org/TR/xml-names/#NT-Prefix) provides the [namespace prefix](https://www.w3.org/TR/xml-names/#dt-prefix) part of the qualified name, and MUST be associated with a namespace URI reference in a [namespace declaration](https://www.w3.org/TR/xml-names/#dt-NSDecl).

I see a few samples in reference/gc with uri http://www.groundspeak.com/cache/1/0, and we have some files using the uri https://humminbird.com.  It would be quite simple to add these to #990 (which I prefer instead of this PR).  basecamp.gpx references a few others but they are unused/unknown to our reader.

tsteven4:
geocaching.com is still using xmlns:groundspeak="http://www.groundspeak.com/cache/1/0", so at least that is live.  I'm not sure any creator ever references this URI with a different namespace prefix, but it's easy to add to the hash table.

It is a bit harder to see to if xmlns:h="http://humminbird.com" is still live, perhaps humminbirdPC creates files using it.

GPSBabelDeveloper:
Yes, Groundspeak were early adopters of GPX 1.0 (which actually grew from
the format we call "geo") and had an early following of developer tools
(and Garmin GPSes) that knew how to read their 1.0 extensions, so they
remain GPX 1.0 extensions instead of the (annoyingly different) GPX 1.1
extensions. They're probably one of the few companies still producing and
distributing hundreds of GB of GPX 1.0 these days. As there's hardware that
is long unsupported by Garmin that reads those files directly, they're
quite change-adverse.

I'm happy with #990. Go for it when you're happy.

tsteven4:
closed in favor of #990

PR #989, opened 2023-01-18
Update interpolate.xml

jidanni:
Add warnings for #981.
Note please somebody make this sound better.
Also link KML to fmt_kml.html .
Also I forgot a closing ")" and now it is too late to learn how to add it.
Thanks.


tsteven4:
I think vague warnings about the kml writer are not appropriate in the interpolate section of the filters chapter.

PR #992, opened 2023-01-20
use bool type

tsteven4:


robertlipe:
LGTM

In new code, I'm increasingly using enums for everything when there's more
than one bool in a signature, but here the mechanical solution still
improves things.

I don't feel it's worth doing more than the mechanical thing on existing
code I'm not really dirtying up.

This is the way.

PR #994, opened 2023-01-20
eliminate some time_t uses.

tsteven4:
This is suppose to be the easy ones.

- text, html: if creation_time is invalid (according to QDateTime, not gpsbabel::DateTime) then the printed string will be empty.
- kml real time position: if the source isn't provided creation times then the icon might be different than before, but I am not sure we care.
- gtrnctr: element is required, continue to use 0 if no data available.
- smplroute: add explicit checks for valid, use msecs instead of secs.  There are subtle differences as (m)secsTo is computed before rounding, while before toTime_t rounds before the difference was computed.
- gpx.cc:  last_found is of class gpsbabel::DateTime.  For default init toTime_T will return 0, and isValid will return false.  There is no need to use toTime_t to see if last_found has been assigned.

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -184,7 +184,6 @@ struct global_options {
 
 extern global_options global_opts;
 extern const char gpsbabel_version[];
-extern time_t gpsbabel_now;	/* gpsbabel startup-time; initialized in main.c with time() */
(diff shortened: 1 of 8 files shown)


GPSBabelDeveloper:
Yes, please.

PR #995, opened 2023-01-23
eliminate use of time_t etc in igc.

tsteven4:


robertlipe:
Fifteen years from this week, 32-bit timet's roll over.

Any remaining Garmin protocol devices, for example, are going to get very
weird.

PR #997, opened 2023-01-26
de-duplicate format setup in find_vec

tsteven4:
This de-duplication isolates the code that prepares a format for execution.  Previously it was duplicated once for normal formats and once for style based formats.  In the future it is anticipated this will be used with dynamic formats, likely with a modified set of parameters.

### DIFF
--- a/format.h
+++ b/format.h
@@ -169,29 +169,6 @@ public:
   virtual ff_type get_type() const = 0;
   virtual QVector<ff_cap> get_cap() const = 0;
 
-  QString get_name() const
-  {
-    return name;
-  }
-
-  void set_name(const QString& nm)
-  {
-    name = nm;
-  }
-
-  QString get_argstring() const
-  {
-    return argstring;
-  }
-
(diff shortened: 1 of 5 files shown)


tsteven4:
I have preliminary work that let's formats either by statically allocated as they are now, or dynamically allocated before rd_init/read/rd_deinit and wr_init/write/wr_deinit, based on a modified vecs_t entry in Vecs::Impl::vec_list for each format.  My test case only involved one format, but passes testo/vtesto.  Dynamic formats should allow simpler lifecycle management of Format class data members.  This PR paves the way by making it easier for main to do the preparation necessary to run a format, some of which is now encapsulated in Vecs::prepare_format.  This primarily consist of argument handling.  The init_vecs and exit_vecs portion is easy to split out, and the added entry in vecs_t is a factory to create a Format instance.

robertlipe:
Exciting!

This was something I was thinking of during the original move to C++, so
I'm happy that you're running with it. Thanx.

PR #999, opened 2023-01-27
create wpts from kml placemark/multigeometry/points.

tsteven4:
amazingly enough earth offered up a kml from a search and we couldn't recover the point.


robertlipe:
Wow. We got the hard cases and missed the easy one.

Yes, please.

PR #1001, opened 2023-02-08
Remove remaining callers and definition of rtrim.

GPSBabelDeveloper:
Minor CV-qualifier battle in jeeps for const correctness. Implement a white-space trail in mkshort where it's low risk and due to be replaced anyway.

### DIFF
--- a/mkshort.cc
+++ b/mkshort.cc
@@ -536,14 +536,12 @@ mkshort(short_handle h, const char* istring, bool is_utf8)
     memmove(dp, np, nlen);
     dp[nlen] = 0;
     // Essentially ostring.rtrim() from here down.
-    if (ostring[0] == 0) { // Empty output string? Bail.
-      return;
+    if (istring && ostring[0] == 0) { // Empty output string? Bail.
+      char *end = ostring + strlen(ostring) - 1;
+      while (end > ostring && isspace(*end)) {
+        *end = 0;
+      }
     }
-    char *end = ostring + strlen(ostring) - 1;
-    while (end > ostring && isspace(*end)) {
-      *end = 0;
-    }
-
   }
 
   /*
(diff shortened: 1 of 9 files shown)


mkshort.cc
```diff
@@ -535,14 +535,20 @@ mkshort(short_handle h, const char* istring, bool is_utf8)
     }
     memmove(dp, np, nlen);
     dp[nlen] = 0;
-    rtrim(ostring);
+    // Essentially ostring.rtrim() from here down.
+    if (ostring && ostring[0] != 0) { // Empty output string? Bail.
+      char *end = ostring + strlen(ostring) - 1;
+      while (end > ostring && isspace(*end)) {
+        *end = 0;
+      }
```
tsteven4: loop may terminate prematurely, end isn't modified in the loop.

GPSBabelDeveloper: Thank you. It's actually a different bug (loop execute once, making it an un-loop) but it's definitely a bug. end is now decremented inside the(intended) loop body.

The process of proving that worked led me to dust off the test case for this and exposed another dormant bug involving trailing whitespace ("foo ." and "foo                  ." didn't reduce correctly) but all is now well.

Someday, I'll replace mkshort.cc with a "real" QString version and it'll become way less fragile. I'll probably design a real test suite for it, too. This is just a tourniquet until then. Beyond building a hashlist to sit on top of a QList (it should just BE a QList...) all that byte-wise string bashing is head-hurting. It was also designed kind of PIMPL and OO-ish, so changing most of the callers should be nearly mechanical. It seems I've started that a couple of times before. Anyway, that's a different project...

This is code that was super important to me in the days of GPSR's using 6 and 8 character names and dropping them if they were duplicates. Now that most can handle 20 or more and there mostly isn't a requirement that they be unique, I'm not really sure how useful any of this is in practice. We're not often talking to hand-coded Dragonball assembly code with fixed, small field and table sizes on the other end of the wire these days.

PR #1003, opened 2023-02-13
Update gpsbabel_man.xml

jidanni:
Fix critical mistake, overlooked for years.
(Sure hope I am editing the real source file.
There is no warning at the top of this file that it is machine generated.)


tsteven4:
Thanks.

This has never been correct, starting with the initial version from Jun 8 16:37:51 2006.  This file is used by a limited number of linux distributions, but is not used by installers provided by gpsbabel.org.

PR #1006, opened 2023-02-18
Change QLatin1String to Qlatin1StringView as recommended in Qt6.4 docs

robertlipe:


GPSBabelDeveloper:
Womp Womp Womp. Come back to this when Qt 6.4 is our floor everywhere, Spaceman.

PR #1008, opened 2023-02-20
add support for dynamic filters

tsteven4:


main.cc
```diff
@@ -451,10 +451,26 @@ run(const char* prog_name)
...
-        filter->init();
-        filter->process();
-        filter->deinit();
-        FilterVecs::free_filter_vec(filter);
+        if (filter.isDynamic()) {
+          filter.flt = filter.factory();
+          FilterVecs::init_filter_vec(filter.flt);
+          FilterVecs::prepare_filter(filter);
+
+          filter->init();
```
GPSBabelDeveloper, line 459: Would it be easier to have init and deinit right called from the ctor/dtor of these or is that just busy work in each of the filter bodies for now?

filter_vecs.cc
```diff
@@ -193,63 +199,67 @@ FilterVecs& FilterVecs::Instance()
...
-    if (svecname.compare(vec.name, Qt::CaseInsensitive) != 0) {
-      continue;
-    }
+  QVector<arglist_t>* args = fltdata->get_args();
 
-    QVector<arglist_t>* args = vec.vec->get_args();
+  Vecs::validate_options(fltdata.options, args, fltdata.fltname);
 
-    Vecs::validate_options(options, args, vec.name);
+  /* step 1: initialize by inifile or default values */
```
GPSBabelDeveloper, line 208: I really wonder if anyone uses the inifile stuff... (This isn't the place for tackling that. Just musing.)

GPSBabelDeveloper (approved):
I like it a lot!

PR #1010, opened 2023-03-01
Make exif debugging a constexpr compile-time conditional instead of #ifdefs

robertlipe:


exif.cc
```diff
@@ -129,6 +129,9 @@
 #define GPS_IFD_TAG_DATUM     0x0012
 #define GPS_IFD_TAG_DATESTAMP 0x001D
 
+// Internal flag for debugging.
+const bool validate_tags = false;
```
tsteven4, line 133: constexpr bool validate_tags ...

exif.cc
```diff
@@ -337,10 +340,11 @@ ExifFormat::exif_load_apps()
   return exif_app_;
 }
 
-#ifndef NDEBUG
 void
 ExifFormat::exif_validate_tag_structure(const ExifTag* tag)
 {
+  if (!validate_tags) return;
```
tsteven4, line 346: if constexpr (...

exif.cc
```diff
@@ -563,9 +566,9 @@ ExifFormat::exif_read_ifd(ExifApp* app, const uint16_t ifd_nr, const gbsize_t of
       }
       printf("\n");
     }
-#ifndef NDEBUG
-    exif_validate_tag_structure(tag);
-#endif
+    if (validate_tags) {
```
tsteven4, line 569: if constexpr (...

exif.cc
```diff
@@ -1214,9 +1217,9 @@ ExifFormat::exif_write_ifd(ExifIfd* ifd, const char next, gbfile* fout)
   for (auto& tag_instance : ifd->tags) {
     ExifTag* tag = &tag_instance;
 
-#ifndef NDEBUG
-    exif_validate_tag_structure(tag);
-#endif
+    if (validate_tags) {
```
tsteven4, line 1220: if constexpr (...

tsteven4 (commented):
I note NDBUG is not defined for CMAKE_BUILD_TYPE=Debug.  build_extra_tests.sh is a CI test case that used Debug, so the code was exercised in CI.  Note the macro assert also depends on NDEBUG.

I added comments related to constexpr if we are going to hard code it.  However, maybe it is better to base validate_tags off global_opts.debug_level

In any event I think it is desirable to exercise the assertions during CI.  I suspect adding a non-zero debug level would cause some tests to fail. e.g. those that test to verify standard output works.

robertlipe:
I don't have deep feelings about how it's triggered. I'm just trying to have less #if FOO_DEBUG code that's rarely compiled. Putting it under global_opts.debug seems a bit of a misfit, but I don't totally hate that idea. Certainly that means we're going to ship the blocks that would have normally been compiled away in a non-debug build.

I can make #if NDEBUG constexpr blah = \#if defined (NDEBUG) \ true \# elsefalse ; but that's the kind of code we're trying to avoid. Still, it reduces it by far.

This doesn't change assert. If you like assert, you can keep your asserts.

In my testing (not of this file specifically), a global that defined to true and 'if (foo)' andn 'if constexpr (foo) always acted the same when optimized. When not optimized, the code was emitted but there wasn't enough constant propagation to allow emit to elide the blocks. I just want the bodies of the code to be built in CI.

I can drop all the constexpr nonsense in this file and move it to a flag guarded by global_opts.debug if you think that's better. That sidesteps most everything, I think.

tsteven4:
There are a bunch of files that got checked in here that should not be in version control.  This made codacy unhappy.  Everything but exif.cc, exif.h and xmlgeneric.cc should not have been checked in.

PR #1012, opened 2023-03-01
Make track filter debugging an internal compile-time consant.

robertlipe:


robertlipe:
This code isn't large or even clever. Should this just be moved to a runtime test and printed if debug level >=2 or something?

Issue #1015, opened 2023-03-01
fmt_unicsv.html: clarify "Add feet" 


jidanni:
On https://www.gpsbabel.org/htmldoc-development/fmt_unicsv.html
please clarify
```
  alt =      Elevation (in meters) of the point. Add "ft" or "feet" for feet.
```
Say if you mean we should use
- alt ft, or
- altft, or
- ft

The first two give errors. So maybe you mean the last.
But then say "use instead" instead of "Add".

But saying just "ft" (or "feet") would then conflict with
```
prox =     Proximity (in meters).  Add "ft" or "feet" for feet.
```

So please add examples of what you mean to that page.

Also one notes the `-o unicsv` example has
```
No,Latitude,Longitude,Name,Date,Time,Filename
```
which means one cannot just go round
```
$ ... -o unicsv | ... -i unicsv
```
without shortening the header names back into their abbreviated form. Interesting.

tsteven4:
***alt ft***, ***altft*** both work, we try to be flexible.  In the example below an input altitude of 10 feet is converted to 3.048m in the gpx file (gpx ele element values are always in meters).

:~/work/vecs$ cat x.csv
No,Latitude,Longitude,Altitude,Date,Time,***alt ft***
1,51.599380,-177.425000,9754.0,2022/12/20,20:25:38,10

```
:~/work/vecs$ ./bld/gpsbabel -i unicsv -f x.csv  -o gpx -F -
<?xml version="1.0" encoding="UTF-8"?>
<gpx version="1.0" creator="GPSBabel - https://www.gpsbabel.org" xmlns="http://www.topografix.com/GPX/1/0">
  <time>2023-03-02T17:11:35.362Z</time>
  <bounds minlat="51.599380000" minlon="-177.425000000" maxlat="51.599380000" maxlon="-177.425000000"/>
  <wpt lat="51.599380000" lon="-177.425000000">
    <ele>3.048</ele>
    <time>2022-12-21T03:25:38Z</time>
    <name>WPT001</name>
    <cmt>WPT001</cmt>
    <desc>WPT001</desc>
  </wpt>
</gpx>
```

tsteven4:
This was resolved by #1018

tsteven4:
I mistakenly thought #1018 was closed.

tsteven4:
Closed, #1018 is merged.

PR #1018, opened 2023-03-03
Clarify how to write feet in fmt_unicsv

jidanni:
For #1015.
By the way e.g., "height" oddly only mentions meters so I do the same.

### DIFF
--- a/xmldoc/formats/unicsv.xml
+++ b/xmldoc/formats/unicsv.xml
@@ -11,7 +11,7 @@
 <para>
    The list of keywords include:
    <screen>
-      alt =      Elevation (in meters) of the point. Add "ft" or "feet" for feet.
+      alt =      Elevation (in meters) of the point. For feet use "alt ft", "altft", "alt feet", or "altfeet".
       arch =     <link linkend="style_def_isarchived">Geocache archived flag</link>
       avail =    <link linkend="style_def_isavailable">Geocache available flag</link>
       bng_e =    British National Grid's easting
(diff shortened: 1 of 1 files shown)


robertlipe:
Great. Thank you.

tsteven4:
The unicsv format documentation has bigger issues.  The use of screen elements is not working out for the pdf conversion.

GPSBabelDeveloper:
I haven't looked, but screen was one of the docvook5 changes.

tsteven4:
babelpdf.xsl seems to cause the problem.

tsteven4:
problem came in here https://github.com/GPSBabel/gpsbabel/commit/63b6e0121 
```
Fix build on Catalina, HTML doc formatting improvements (https://github.com/GPSBabel/gpsbabel/pull/759)
* PDF maker: format screen and input same. Shrink padding slightly.
* ```

This change lost the verbatim nature of screen elements, i.e. newlines are lost and the screen content is all jumbled together.

tsteven4:
[screen](https://tdg.docbook.org/tdg/5.0/screen.html) is a verbatim environment.  [userinput](https://tdg.docbook.org/tdg/5.0/userinput.html) is not a verbatim environment, but an inline.  We are applying a [customization](http://www.sagehill.net/docbookxsl/CustomInlines.html) for inline text to a verbatim environment (screen).  

The result is we lose the line breaks and the list of unicsv keywords is all mushed together and thus unreadable.

tsteven4:
We can regain respect for line endings by not matching screen in babelpdf.xsl.  But, when the lines in the screen are too long they get cutoff.

tsteven4:
A fix is described in http://www.sagehill.net/docbookxsl/FittingText.html, although the hyphenation character part doesn't seem to work.

[patch.txt](https://github.com/GPSBabel/gpsbabel/files/10902310/patch.txt)

tsteven4:
For the[ bigger issue](https://github.com/GPSBabel/gpsbabel/pull/1018#issuecomment-1456607119) a potential fix is #1027.
Let's get that settled before we deal with this PR as this PR increases line lengths in a verbatim element and is thus related.

robertlipe:
Approved.

I can regenerate and re-push -development if needed.

tsteven4:
BTW "alt", "ele", "height" are all equivalent and work in meters by default, or in feet with the extension.  "z" also means the same thing, but it is always in meters.

tsteven4:
This resolves #1015

PR #1022, opened 2023-03-06
Update interpolate-distance.xml

jidanni:
In #1021 we gave sub-second time examples. So we had better give sub-mile distance examples too!


GPSBabelDeveloper:
I don't think we got fraxtuonal args accepted until tsteven4 added them in
August in
https://github.com/GPSBabel/gpsbabel/commit/1df589daa16d15475990048f88cd3722d7f21aba

For sub-second work in filters, you really need to build from source or use
the preview binsl.

PR #1026, opened 2023-03-06
support subsecond resolution for trackfilter start/stop options.

tsteven4:


robertlipe:
LGTM
That code. Is icky.

tsteven4:
Hard to argue.  In general we are not very consistent in the format of similar options, although we do have parse_distance.  In many cases each option was developed in isolation, seemingly according to the whims of the creator.  In this case, making things more complicate isn't making it prettier.  It is too bad QDateTime::fromString doesn't handle subseconds in a useful way, perhaps a better solution is to hide that in a derived class.  This tripped up nmea for a while.

GPSBabelDeveloper:
As usual, we agree.

Throughout the project, M being Meters or Miles, depending on the home
country of the author, remains a regret.

tsteven4:
> That code. Is icky.

Ok, you shamed me into a cleaner implementation.

robertlipe:
I'm not into shame-driven development - the sale of GPSBabel World HQ
to Musk Industries is on hold - but that is a LOT better, so thank you.

PR #1027, opened 2023-03-06
wrap long lines in verbatim elements, such as screen.

tsteven4:
The use of inline.boldmonoseq on verbatim elements such as screen results in a lose of the line endings. This makes it impossible to differentiate items in what might have better been a list.  However, if the line in the screen element is too long it will be trucated. Use the monospace.verbatim.property wrap-option to wrap any such lines.

An example that illustrates the problems is the list of keywords in the unicsv format.

### DIFF
--- a/xmldoc/babelpdf.xsl
+++ b/xmldoc/babelpdf.xsl
@@ -21,13 +21,14 @@
     <xsl:call-template name="inline.boldmonoseq"/>
   </fo:block>
 </xsl:template>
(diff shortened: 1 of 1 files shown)


robertlipe:
Yes, please.

I remember that something (FOP?) was crashing the build after an upgrade of (??? who can tell any more) and break-after="auto" got it building, but it clearly, after your investigation, the wrong fix.

Thank you.

PR #1029, opened 2023-03-07
remove some orphan gpssim reference files.

tsteven4:


tsteven4:
These were missed in #1028

Issue #1019, opened 2023-03-05
First node of a track shouldn't have Heading. Or if it was a middle node before truncation, more could be mentioned


jidanni:
[We see](https://www.jidanni.org/comm/air/routes/programs/gps/tips.html) the Heading field is definitely bogus and should not be output on the first node of a trace.

But that first node could actually be fully filled out, as earlier in the same command line its heading and speed were indeed known.

Data file used:
[FlightAware_UAL871_KSFO_RCTP_20221220.kml.gz](https://github.com/GPSBabel/gpsbabel/files/10870668/FlightAware_UAL871_KSFO_RCTP_20221220.kml.gz)

tsteven4:
This is an ancient issue with track_recompute(), which is used by the kml, gpssim and gtrnctr **writers**.  I almost wonder if this was done on purpose for some unknown reason that is hard for us to imagine now.  A fix touches 9 reference files.  While it is fair game for a writer to manipulate output data as it sees fit, it is questionable practice for a writer to change the internal route and track data as track_recompute does.  This has side effects that will be visible to any subsequent writer.  An alternative is to use the track filter course and speed options to synthesize those fields.

To fill out the your selected first node add the course option when using the track filter as shown below.  This uses a different algorithm than track_recompute with it's own peculiarities related to the initial speed and course.  However since you aren't selecting the first point the anomalies aren't visible in your output.
```
$ t=20221221032538 f=FlightAware_UAL871_KSFO_RCTP_20221220.kml
$ [gpsbabel](https://www.gpsbabel.org/) -i kml -f $f \
-x interpolate,time=1 \
-x track,course,start=$(($t-1)),stop=$t_ \
-o kml,units=a,points=1,lines=0,track=0 -F out.kml
```

robertlipe:
That's clearly a wart. It was done in the early KML writer to give users
pretty graph data without adding a filter invocation as it built without
filters. I think gx:Track may make that less important.

It was kind of OK in the (sigh) character-converter days because the
writers were secretly given their own copy of the blob in flight because
all the data was copied to new wpt's in order to be iconv'ed and then
destroyed.  Yes, it was dueling warts that canceled each other out.

I thought I deleted GPSSim. Maybe I intended to. Maybe I just should. Do we
have any proof of users?They contracted for that years and years ago, but
they be maintaining their own local copy.

Should gtrnctr suffer that same fate?

I'm OK with a little fix that touches refs and I think  I'm OK with
[...]

PR #1031, opened 2023-03-07
don't create fictitious course and speed values.

tsteven4:
We compute course and speed in track_recompute, which is used
by the kml and gtrnctr writers.  We also compute these in the track
filter with the course speed options.

Both of these algorithms compute the course and speed for a point
based on the time and position of that point and a previous point.
Given that no information from the future is used, the course and
speed of the first point cannot be determined.  This PR introduces
a change so fictitious values for the first point are not generated.


tsteven4:
This fixes #1019

robertlipe:
Good fix. Thanks.

tsteven4:
Some of the changes in realtime.kml aren't obvious.  They are the result of

1. the random realtime reader generates course base on consecutive points.
2. the kml realtime writer skips points that were within 50 meters of the previous point.
3. the old recompute_track always re-generated course based on the points that weren't skipped.  the new recompute track doesn't recompute course if the original point had a course.  The previous point used in the course calculations by the random realtime reader and the kml realtime writer will be different if we skip a point.

GPSBabelDeveloper:
This is all fine. We have a lot of latitude, IMO, in what's in those
balloons in realtime KML...  The intersection of users of realtime and
those doing computations based on the text in those kml balloons is surely
the null set. Surely. (Google is free to come to me with requirements and a
plan to support it otherwise....)

For #2, this is because Earth of the 3.0 era when this was introduced,
didn't need an indefinite number of drawables plopped on the map while it
was trying to scroll, probably while tethered to an Edge connection. We
were targeting motorists and marine, so 50M steps wasn't unreasonable.

It's pretty rare for non-marine units to provide course that's any more
meaningful than we we get from our own calcs. Wheel and wind sensors aren't
exactly common these days. The difference in course computed from a
previous point or a dropped point is just not THAT significant.



It's good to call these sorts of things out, but I think it's the right
call here to shrug it off.

RJL

PR #1034, opened 2023-03-11
document fixes

tsteven4:
fix fop warnings "... exceed the available area in the inline-progression direction ...".  This PR resolves all these warnings in CI on macos and ubuntu jammy, but we still get some on ubuntu focal.

spelling corrections

in cmake set USES_TERMINAL on document jobs so we get output as they run.


tsteven4:
The problem in focal is related to an empty fop-hyph.jar file in that distribution.  Debian fixed in in 2.4-3, but Ubuntu focal still has 2.4-2.

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=932734

PR #1055, opened 2023-03-30
Update kml-deficon.xml

jidanni:
It's not a name. It's a URL!

P.S., I will probably spend the afternoon hunting this down, scattered through the source code.

(P.S., here's some undocumented ones that even work offline

```
$ perl -nwle '/http.*pin.png/ && undef $h{$&};
  END{print for sort keys %h;}' \
  /opt/google/earth/pro/resources/userpalette.kml
http://maps.google.com/mapfiles/kml/pushpin/blue-pushpin.png
http://maps.google.com/mapfiles/kml/pushpin/grn-pushpin.png
http://maps.google.com/mapfiles/kml/pushpin/ltblu-pushpin.png
...)
```

### DIFF
--- a/xmldoc/formats/options/kml-deficon.xml
+++ b/xmldoc/formats/options/kml-deficon.xml
@@ -1,3 +1,3 @@
 <para>
-This option specifies the default name for waypoint icons
+This option specifies the default URL for waypoint icons
 </para>


jidanni:
Maybe gpsbabel/xmldoc/formats/options
/geo-deficon.xml needs to be fixed too.
And add an example there, so people know what you are talking about.

https://github.com/search?q=repo%3AGPSBabel%2Fgpsbabel%20deficon&type=code has more.

robertlipe:
They are names from a palette provided by Google (and probably poorly documented) OR they can be URLs of images that meet specific criteria.

We'll stick with our parlance of names.

jidanni:
Well do add some examples, as there is no way a user can guess
- names from a palette provided by Google
- OR they can be URLs

Nope, no way to imagine.

GPSBabelDeveloper:
We're a conversion tool, not an authoring tool.

If you convert a KML with a placemark with an attached image of a house,
it's not like can convert that to a "home" icon on your Garmin. We're as
blocked by this process as you are. Yet in practice, is primarily an output
for us and our millions of users haven't found this problematic. (Which is
a good thing.) It's not like we can exactly harvest the textual
descriptions of http://kml4earth.appspot.com/icons.html#shapes and convert
those to anything useful in other programs. Should we convert all boats to
the picture of a boat or only sail boats? Only blue boats?

Don't answer any of that. It's rhetorical. The mapping of icons in and out
of Google's mapping products is not at all trivial. People that need them
know how to get them passed through.


Instead of these academic nitpicks, I'd really like to focus my remaining
development time on supporting the product now....

jidanni:
All I know is I had to give a value to deficon and then examine the KML to see what it did. Apparently it adds 'href' tags to whatever you give it. Okay well at least document that. Thanks.

PR #1056, opened 2023-03-30
Update help.txt

jidanni:
I am just guessing that all three of these should be changed.


robertlipe:
None of these are strictly URLs.

PR #1060, opened 2023-04-07
Integrate OAT and SIU with Waypoints

PacketFiend:
As always, there just might be a better way to do this.

Overloads the igc_fsdata.set_value() method so if supplied a Waypoint, it will merge OAT with Waypoint::temperature and SIU with Waypoint::sat. Plus a few other changes because Waypoint::sat wasn't fully fleshed out.

If set_value() is not supplied a Waypoint object, all data will be stored as igc_fsdata with no integration.

\# of satellites is not included in trackpoint descriptions, that's already getting a bit bloated...


robertlipe:
If it's not useful to graph (and it's not), why is this useful?

We've had a couple of people ask about tracking the satellites in view from
NEMA, in particular, but when we remember we're a conversion program, even
if you do find two programs that have a way to represent this, how is it
useful to know that space vehicles, A, B, C, and D were visible above the
horizon and delivering a signal that was useful to the correlators?

If this really IS usefu, the duplication of set_Value() bothers me. We
should implement one in terms of the other. But let's answer the "why"
before we deep dive on the 'how'.

PacketFiend:
The only reason I can think of that SIU would be useful to graph would be as a way to verify fix accuracy (FXA), which is mandatory for badge and record claims, and all FAI competitions. The two should be almost perfectly inversely proportional. Are there other reasons people have requested it?

Either way it would be better to put SIU behind an option of some kind when generating the KML. It's something I'd like to do anyway as the number of extensions can become cumbersome, but I'm not there yet.

My basic thinking is, if we *have* the data, and we *can* graph it with existing constructs, why not at least provide the *option*?

More philosophically, should a translation program be inserting itself into the debate over which data is useful and which is not? My personal opinion is no, but I'm new around here.

GPSBabelDeveloper:
Are your folks *really* checking claims that your device said it was
receiving  sat s.y.z, but y is marked offline for maintenance on that day
and the almanac says that SV klllz would have been below the horizon at
time/date DT for that lat/long and therefore unusable for generation

That's possible, of course, but seems pretty intense.

I think we already have an 'nsats' and [HV] ]pdop, but in the past we've
declined to store the array of SVs and  their respective health at all

tsteven4:
I don't think we are talking about storing the array of satellites that are in view anywhere.  We are talking about storing data from the SIU extension which is the number of satellites in use.

`SIU I, B Satellites in use. A two-character field from the NMEA GGA or GNS sentences, or equivalent data agreed by GFAC`

GGA field 7 = Number of satellites in use, 00 - 12
GNS field 7 = Total number of satellites in use, 00-99

Note the GGA NMEA field is stored in Waypoint::sat, we don't parse NMEA GNS sentences.

Storing the SIU data into Waypoint::sat is appropriate.  Plotting the number of satellites in use seems of questionable utility, especially since estimated position error data is available from the required FXA extension.

An overload for set_value is not necessary, only one is used.  But if both were needed it could be covered by adding a default to the wp parameter:
bool set_value(IgcFormat::igc_ext_type_t type, double value, Waypoint* wp = nullptr)
and checking wp for nullptr before use.

Dead code, or code that isn't useful does increase the maintenance burden.  Dead code can negatively impact our test coverage, and can lead to "false" positives detected by quality tools.

PacketFiend:
Ah, that makes more sense, I wasn't following Robert. Steve is correct,
it's not an array of satellites, just a simple count. The only case I could
see it being useful to graph would be to casually check that one's reported
position error (FXA), is , in fact, correct, and matches the number of
satellites in use (SIU). It's been #ifdef'd out in kml.cc until now as
yeah, it's not very useful, especially considering FXA is mandatory. That
said, the goal is to make it *optional* to plot somehow - which I haven't
gotten to yet.

Re: overloading set_value(), I also considered the *nullptr case. It was a
mental coin toss, I'm happy to do that either which way. Both methods
provide the opportunity to bypass storing data in Waypoints as a way to
test during development, if necessary.

kml.cc
```diff
@@ -1935,6 +1948,9 @@ void KmlFormat::write()
     if (traits->trait_depth) {
       kml_mt_array_schema(kmt_depth, "Depth", "float");
     }
+    if (traits->trait_sat) {
+      kml_mt_array_schema(kmt_sat, "Satellites", "float");
```
tsteven4: If we do this you should use type "int" instead of "float".

kml.cc
```diff
@@ -1496,6 +1496,10 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
       writer->writeTextElement(QStringLiteral("gx:value"), wpt->temperature_has_value()?
                                QString::number(wpt->temperature_value(), 'f', 1) : QString());
       break;
+    case wp_field::sat:
+      writer->writeTextElement(QStringLiteral("gx:value"), wpt->sat?
+                                QString::number(wpt->sat, 'f', 1) : QString());
```
tsteven4: If we do this you should use QString::number(wpt->sat).  Don't implicitly convert it to a double.

PacketFiend:
All better.

For some reason I thought SIU was being stored with an implied decimal and that some recorders were interpolating missing values. It's a two character field though which **must** record up to at least 12 satellites ([3.3](https://www.fai.org/sites/default/files/igc_fr_specification_with_al8_2023-2-1_0.pdf)) so definitely an int. It was Earth interpolating some data.

Waypoint->temperature [is actually a float](https://github.com/GPSBabel/gpsbabel/blob/cad16e14c34f3ee16fde021aa4df82b2461f5bff/defs.h#L359), not a double. Any issues there?

And looking a bit more closely, Waypoint->sat [is an int](https://github.com/GPSBabel/gpsbabel/blob/cad16e14c34f3ee16fde021aa4df82b2461f5bff/defs.h#L478), so `wp->sat = value` in set_value() is also an implied cast. Would `wp->sat = (int)value` work instead?

tsteven4:
1. There is code that is not used.  set_value is always called with a waypoint that is not nullptr.  Restricting operation of get_value to this condition allows simplification.
2. There is a subtle bug in kml.cc regarding deciding if Waypoint::sat is valid when printing.  It was possible to merge and print -1 for the number of satellites.

These are both fixed in this patch.  Please consider it.
[unused.patch](https://github.com/GPSBabel/gpsbabel/files/11217299/unused.patch)

PacketFiend:
> 1. set_value is always called with a waypoint that is not nullptr.  Restricting operation of get_value to this condition allows simplification.

Currently, yes. My idea is to leave it like this so if some future development requires bypassing the integration of extension data with Waypoints (there's a lot of pointing and dereferencing and casting going on there), the *nullptr case provides that. Whether or not to include that is a stylistic choice methinks, so I'll leave that up to you folks. Personally, I would like to keep the ability to keep all the IGC extension data separate, as a way to make future development easier. I am blissfully unaware of whether or not that's testable and what broader implications it has.

> 2. There is a subtle bug in kml.cc regarding deciding if Waypoint::sat is valid when printing.  It was possible to merge and print -1 for the number of satellites.

Ah yes, I see that now. Will fix.

PacketFiend:
Coming back to this after a week or so, and I surely merged something into the wrong place. My `master` branch wound up with code that's supposed to be on my `develop_igc_options` branch alone. My apologies for any confusion.

My intent is to keep all the IGC optional extension stuff on a seperate branch. This PR should contain no code that includes individual boolean options for IGC extension ingestion. It should be limited to the integration of IGC data with Waypoint members.

As per above, my preference is to keep the *nullptr case for set_value, so it's possible to easily separate all the IGC extension stuff in the future, or to extricate it down the line if that becomes necessary. Whether or not you guys think that's a good idea, I don't know.

waypt.cc
```diff
@@ -74,6 +74,7 @@ void update_common_traits(const Waypoint* wpt)
   traits.trait_power |= wpt->power > 0;
   traits.trait_depth |= wpt->depth_has_value();
   traits.trait_temperature |= wpt->temperature_has_value();
+  traits.trait_sat |= wpt->sat > 0;
```
tsteven4: sorry for the delay - this is just a start:

Since -1 is the default, and we may have 0 satellites, shouldn't this be "wpt->sat >= 0"?

tsteven4:
In general macros are discouraged in c++, and we are slowly working through our legacy converting them.
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-macros2

Can we make
// #define INCLUDE_IGC_TRT // Generally not very useful to graph on Google Earth
#define INCLUDE_IGC_SIU // Satellites in use, not entirely useful to graph
into constexpr bools in the kml.h "Constants" section, and use constexpr if statements in kml.cc?  I think you will need to declare has_igc_siu and has_igc_trt all the time, but the rest of the #ifdef/#endif can be replaced with "if constexpr (kIncludeIGCSIU) {...}" etc.  This will give better visibility of the conditional code to analysis tools.

kml.cc
```diff
@@ -70,7 +70,7 @@
 
 #define MYNAME "kml"
 // #define INCLUDE_IGC_TRT // Generally not very useful to graph on Google Earth
-// #define INCLUDE_IGC_SIU // Satellites in use, not entirely useful to graph
+#define INCLUDE_IGC_SIU // Satellites in use, not entirely useful to graph
```
tsteven4: please delete the define for INCLUDE_IGC_TRT, INCLUDE_IGC_SIU.

kml.cc
```diff
@@ -1702,6 +1707,10 @@ void KmlFormat::kml_mt_hdr(const route_head* header)
       kml_mt_simple_array(header, kmt_power, wp_field::power);
     }
 
+    if (has_sat) {
```
tsteven4: This is nearly a dupe of L1741:1744.  Since we haven't put sats in an array up till now, maybe all we need is the kIncludeIGCSIU block.

tsteven4: The more I think about this I think we should delete these added lines L1710:1713, and leave in the igc specific ones below.  There hasn't been any cry to have sats in an an array, but there have been repeated whines about the kml file sizes.  Also, in these days with multiband GPS, I think there is less interest now than ever.

robertlipe:
Agreed. People used to try to post-process GPS data by knowing there were
three birds in sight at azimuth XYZ and spherical altitude ABC and try to
compute "better" fixes, but the correlators are so good these days that
this is rarely done this way now in consumer-ish systems and the people
that ARE doing it are with professional gear, using ROVER files from
Trimble gear that just aren't in our space.  Sure, it's in some NMEA
streams but mostly those fields are turned off for space and timing reasons
in things like drones.

I know this is in an extension, but unless there's a real reason for that
data, it's just more numbers to obsess about and things like GLONASS,
Galileo, and unifyign all the European-specialized networks sounds icky.

We're fundamentally a location conversion utility. Does this data help us
convert to or from any other formats and convey any necessary information
for them?

PacketFiend:
I can only comment on this with my perspective as a competitive glider pilot.

While the specific array of satellites is not useful to us, the *number* of satellites, and therefore the fix accuracy, is. Recording the number of satellites or fix accuracy is necessary for us.

I will need to dig through the code a bit more, but I do think that we should provide a facility by which to include this metric in the Waypoint object.

robertlipe:
We have |sat| to hold "the number of sats used for fix" as well as fix (2d,
3d) and the three DOP brothers which we carry around, though they're rarely
useful in a conversion.

This is all on the common struct Waypoint, not one of the extensions. This
is one of a few things (like heart rate and all the geocaching stuff)
that's in the common part that probably SHOULD BE in an extension, but here
we are...and it's been years since anyone has legitimately reported running
out of memory inside GPSBabel so shrinking it isn't much of a priority,
though we try to hold it down since it can be multiplied by tens or
hundreds of thousands in a run for us.

RJL

tsteven4:
> I will need to dig through the code a bit more, but I do think that we should provide a facility by which to include this metric in the Waypoint object.

Waypoint::sat is "number of sats used for fix" so recording the " the number of satellites" is already provided for in Waypoint.

I now realize the two blocks:
kml_mt_simple_array(header, kmt_sat, wp_field::sat);
kml_mt_simple_array(header, kmt_igc_siu, wp_field::igc_siu);
are mutually exclusive because IgcFormat::setvalue sets one or the other (as implement always the former).
The same is true of 
kml_mt_simple_array(header, kmt_temperature, wp_field::temperature);
kml_mt_simple_array(header, kmt_igc_oat, wp_field::igc_oat);

So I understand your reluctance to delete the block with kml_mt_simple_array(header, kmt_sat, wp_field::sat);.
Perhaps a compromise is to have setvalue set both the wp sat field and igc_siu and delete the kml_mt_simple_array(header, kmt_sat, wp_field::sat); block.  Then IGC users can see sat data in their kml, but other apparently satisfied users who do complain about kml sizes won't be bothered with it.  This will also leave the data in the Waypoint for other writers to make use of.

Now temperature has always been output as a kml_mt_simple array and seems of broader interest.  So IGCFormat::setvalue could store both the wp temperature field and igc_oat, and delete the kml_mt_simple_array(header, kmt_igc_oat, wp_field::igc_oat); block so we don't print twice.  This would also leave the data in the Waypoint for other writers to make use of.  It does mean igc_oat is a red herring left to confuse us as nobody would consume it.

I do realize that those two items would be handled differently by the above suggestion, sats only goes to a mt array for IGC, while temp goes to mt array irregardless of where it came from.

tsteven4:
Another possibility that avoids spamming non-igc users with the number of satellites is to
1) have IgcFormat::setvalue set both the appropriate field in Waypoint and the field in igc extended data for siu and oat.
2) have a kml option to use igc extension data.  When using that option kml outputs multitrack arrays using igc extended data.  If the option is false then kml outputs multitrack arrays using the Waypoint structure and ignoring igc extended data.  revert a few lines so kml doesn't output the number of sattellites when not using igc extensions.

This preserves symmetry.

@PacketFiend  I'd like to resolve this before an upcoming release if we can.

PacketFiend:
I'm reading and keeping up to date with all the comments here. I'm just a bit swamped at the moment. If not this week, I think I should be able to address all the concerns here by next. Is there a scheduled release coming?

I think a good number of the concerns here will be addressed in another branch I'm working on - https://github.com/GPSBabel/gpsbabel/compare/master...PacketFiend:gpsbabel:develop_igc_options. It's been a while since I got back to that. The general idea is to have a way to specify which IGC extensions to include, with a somewhat sane set of defaults.

I think the big question for this PR is whether we should use the existing constructs for temperature and satellites that already exist in the Waypoint object, or keep those separate as IGC extension data. The former may simplify things for an end user, but the latter will make it easier to just "switch off" all the IGC stuff.

PacketFiend:
This commit stores SIU and OAT in both igc_fsdata and a waypoint object if one is passed to igc_fsdata::set_value().

It will use IGC extension data when writing mt arrays if it's available, for both satellites in use and outside air temperature. If that data is /not/ present (as would be the case with any non-IGC files, or if those extensions were omitted by options), kml.cc will default to using the values in waypoints.

It first writes IGC mt arrays specific to extension data, setting a `include_kmt_sats` or `include_kmt_temperature` flag to false if it finds corresponding IGC data. The existing mt array writer then checks that flag before writing out the array.

This keeps IGC specific code out of the general purpose kmt writer; the flag could be set by any number of formats that produce data which would be useful to override the waypoint with.

Edit to add: I see there's some interest in removing the satellite count from waypoint objects entirely. I can certainly put that in another commit. I do feel it should be handled separately though. I make a concerted effort to keep as many of my changes restricted to IGC specific code, at least until I understand this beast known as GPSBabel better.

Removing Waypoint->sats may cause breakage in several places (although you fellows would have a much better idea of its impact than I would). Wouldn't the removal be better left to another pull request, in case it turns out to be more involved than originally anticipated?

tsteven4:
I merged up and pushed a commit to fix my comment Jun 16 "please delete the define for INCLUDE_IGC_TRT, INCLUDE_IGC_SIU."

However, in reviewing the reference files I found an issue.  In the [Google KML Reference on Extended Data](https://developers.google.com/kml/documentation/kmlreference#extendeddata) they a method to add extended data is:

> Declaring new typed fields using the ```<Schema>``` element and then instancing them using the ```<SchemaData>``` element (advanced)

However, your ```<Schema>``` element and the ```<SchemaData>``` element are out of sync.  Specifically, you have gx:SimpleArrayData with name attributes that don't correspond to any SimpleArrayField name attributes.

For example:
```
    <Schema id="schema">
      <gx:SimpleArrayField name="temperature" type="float">
        <displayName>Temperature</displayName>
      </gx:SimpleArrayField>
    </Schema>
```
...
```
          <ExtendedData>
            <SchemaData schemaUrl="#schema">
              <gx:SimpleArrayData name="Engine Noise">
```
...
```
              </gx:SimpleArrayData>
              <gx:SimpleArrayData name="True Airspd">
```
...

Also see https://developers.google.com/kml/documentation/extendeddata

It seems our assumption is that we can define extra SimpleArrayFields in the Schema, but we try to avoid it using traits as opposed to making an additional pass through all the points.  On the other hand, referring to a Schema element that doesn't have that name defined seems wrong.  

This seems to be more clearly required in kml 2.3, See https://docs.ogc.org/ts/14-068r2/14-068r2.html#atc-127 which says the following about SimpleData (not SimpleArrayData but ...)

> all kml:SimpleData child elements have a 'name' attribute that matches the name of a declared kml:SimpleField element in the associated kml:Schema element (see [ATC-126](https://docs.ogc.org/ts/14-068r2/14-068r2.html#atc-126));

Also see https://docs.ogc.org/is/12-007r2/12-007r2.html#177 which says in 9.7.4 refering to kml:SimpleArrayData
>A name value  reference to the corresponding kml:SimpleArrayField. See 9.12.4.2 name.

And while KML 2.3 never seemed to get off the ground, in my mind it more clearly defines what Google intended with gx:

@robertlipe any comments?

tsteven4:
Maybe the way out of this is to switch from using SchemaData to Data, which would allow us to drop the Schema element and likely the whole trait tracking code.  I would consider that solution out of scope of this PR.  I also note that the current difficulties are not introduced by this PR, but already made it into our master branch.

tsteven4:
Unfortunately Data doesn't seem to support arrays, just key value pairs.

tsteven4:
Another simplifying option may be to output all the possible SimpleArrayFields in the Schema element regardless of whether they are used or not.  This might allows us to drop all support for traits.  This is a couple dozen lines in the kml file and doesn't grow with the number of points, just with the number of possible SimpleArrrayData types.  As noted in waypt.cc, the trait code is "a bit tacky".  The only user of traits I can find is the kml writer.

tsteven4:
Or just make a pass through all the waypoints in kml_write to collect the actual usage.  This would amount to moving some code from kml_mt_hdr (has_*), and wrapping it a new track_disp_all.  And a little extra for fields that can be in igc data or other data (oat = temperature, siu = sat).  Note while this adds a new loop through all the points, the contents are just moving from one loop to another.  And it also let's us drop support for traits.

tsteven4:
I have something working along the lines of my last comment.  As I think we discussed at one point we need to scan all the tracks to see what schema to emit, and scan each track to decide which arrays to output with that track.

This is a significant change, i.e. likely a fit for an additional PR.

tsteven4:
My swipe at the changes is at https://github.com/tsteven4/gpsbabel/tree/kmlaccumulate
It's still a bit rough, but it could eventually be a subsequent PR to resolve the issue I pointed out today.

PacketFiend:
I really like this. I was trying to do it at the top of kml_write with another track_disp_all, but the callbacks and lambdas are unfortunately a bit beyond me currently. This works quite well.

The only change I would make is rather than listing them all individually, enumerate all the IGC extensions at/around: https://github.com/GPSBabel/gpsbabel/blob/5013df57604316b1dc6f9f9f0840c55755a7b148/kml.cc#L1611

Likewise at https://github.com/GPSBabel/gpsbabel/blob/5013df57604316b1dc6f9f9f0840c55755a7b148/kml.cc#L74

Which I simply haven't managed to finish doing yet. The general idea is that changes to IGC code should not, as a rule, require parallel changes to KML code - changing kml.cc can cause problems with other data formats. (And I do try to keep safely within my IGC sandbox wherever possible.)

PacketFiend:
In the meantime though, what is the proposed solution to the schema mismatch issue?

Here's my swipe at it: https://github.com/PacketFiend/gpsbabel/tree/kmlschemafix

I *mostly* works, I just can't yet figure out how to pull `has_*` out of `fs_igc` after that initial run through track_disp_all at the top of KmlFormat::write. Once I've cracked that, it should be a simple matter to write the schemas correctly.

tsteven4:
> In the meantime though, what is the proposed solution to the schema mismatch issue?

I beleive it is fixed in https://github.com/tsteven4/gpsbabel/tree/kmlaccumulate.  They test cases bear this out, the missing SimpleArrayFields have been added to your reference files.

I haven't had a chance to look at your kmlschemafix.

PacketFiend:
Actually, it looks like you did exactly what I was attempting to do, you just did it faster and better than I could and threw in a bit extra.

tsteven4:
> The only change I would make is rather than listing them all individually, enumerate all the IGC extensions at/around:

I don't understand what you mean by that.  If you mean to automate that somehow based only on igc.* perhaps that can wait until later.  This PR has been sitting here way too long.  If we can accept it, and follow it with my kmlaccumulate branch to fix the SchemaArrayField/SchemaArrayData mismatch, and follow that with any further automation I would rather do it in pieces.

PacketFiend:
Absolutely, leave it until later. I'm eager to get this merged too. I was just bouncing ideas.

If your suggestion then is to merge this PR, then merge in your kmlaccumulate branch, and *then* work on other improvements, I'm all for it. I've got another branch that I can't get much further on until this is merged as well.

tsteven4:
I'm merging this with the understanding I will submit a PR to fix the SchemaArrayField/SchemaArrayData mismatch, which is a problem that existed before this PR.

PR #1063, opened 2023-04-11
#1062 Fix core dump when merging tracks with disparate IGC extensions

### LINKED ISSUE #1062: igc exensions can cause core dump
kml.cc line 1509
` double value = fs_igc->get_value(member).value();`
can cause a core dump.

It isn't safe to assume that just because one waypoint had igc extension data available all waypoints will have the same extension data available.  Also the arrays need to have the same number of elements.  If data isn't available for a waypoint print an empty element.
For example
` writer->writeTextElement(QStringLiteral("gx:value"), wpt->temperature_has_value()?
                               QString::number(wpt->temperature_value(), 'f', 1) : QString());`

This is a bit contrived, but it demonstrates the failure:
```
tsteven4@PEDALDAMNIT:~/work/packetfiend$ cp reference/track/92HV66G1.igc x.igc
tsteven4@PEDALDAMNIT:~/work/packetfiend$ cat reference/track/92GV66G1.igc | sed 's/ACZ/YYY/' >y.igc
tsteven4@PEDALDAMNIT:~/work/packetfiend$ ./bld6/gpsbabel  -i igc -f x.igc -f y.igc -x track,merge -x validate,debug -o kml,floating=1,track=1,points=0 -F xy.kml

Processing waypts
point ct: 0, waypt_count: 0

Processing routes
route head ct: 0, route_count: 0
total route point ct: 0, route_waypt_count: 0

Processing tracks
track 0 ct: 3194, waypt_count: 3194
track head ct: 1, track_count: 1
total track point ct: 3194, track_waypt_count: 3194
terminate called after throwing an instance of 'std::bad_optional_access'
  what():  bad optional access
Aborted (core dumped)

Core was generated by `./bld6/gpsbabel -i igc -f x.igc -f y.igc -x track,merge -x validate,debug -o km'.
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=139773484832768) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=139773484832768) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=139773484832768) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=139773484832768, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f1f8d05c476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f1f8d0427f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007f1f8d3edbbe in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007f1f8d3f924c in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#7  0x00007f1f8d3f92b7 in std::terminate() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#8  0x00007f1f8d3f9518 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#9  0x00005630d89be482 in std::__throw_bad_optional_access () at /usr/include/c++/11/optional:102
#10 0x00005630d89bf527 in std::optional<double>::value() && (this=0x7fff74221420) at /usr/include/c++/11/optional:968
#11 0x00005630d89b7b0b in KmlFormat::kml_mt_simple_array (this=0x5630d8b99600 <Vecs::Instance()::impl+1440>, header=0x5630da455bf0, name=0x5630d8ae8687 "Z Accel",
    member=KmlFormat::wp_field::igc_acz) at /home/tsteven4/work/packetfiend/kml.cc:1509
#12 0x00005630d89b9188 in KmlFormat::kml_mt_hdr (this=0x5630d8b99600 <Vecs::Instance()::impl+1440>, header=0x5630da455bf0) at /home/tsteven4/work/packetfiend/kml.cc:1722
#13 0x00005630d89ba4a4 in operator() (__closure=0x7fff74221730, rte=0x5630da455bf0) at /home/tsteven4/work/packetfiend/kml.cc:1967
#14 0x00005630d89bd591 in RouteList::disp_all<KmlFormat::write()::<lambda(const route_head*)>, KmlFormat::write()::<lambda(const route_head*)>, std::nullptr_t>(struct {...}, struct {...}, decltype(nullptr)) (this=0x5630da45e850, rh=..., rt=..., wc=<error reading variable: Attempt to dereference a generic pointer.>)
    at /home/tsteven4/work/packetfiend/defs.h:786
#15 0x00005630d89bcf13 in track_disp_all<KmlFormat::write()::<lambda(const route_head*)>, KmlFormat::write()::<lambda(const route_head*)>, std::nullptr_t>(struct {...}, struct {...}, decltype(nullptr)) (rh=..., rt=..., wc=<error reading variable: Attempt to dereference a generic pointer.>) at /home/tsteven4/work/packetfiend/defs.h:843
#16 0x00005630d89bbcd1 in KmlFormat::write (this=0x5630d8b99600 <Vecs::Instance()::impl+1440>) at /home/tsteven4/work/packetfiend/kml.cc:1972
#17 0x00005630d89d28d5 in run_writer (ovecs=..., ofname=...) at /home/tsteven4/work/packetfiend/main.cc:293
#18 0x00005630d89d367b in run (prog_name=0x7fff74223462 "./bld6/gpsbabel") at /home/tsteven4/work/packetfiend/main.cc:412
#19 0x00005630d89d5163 in main (argc=15, argv=0x7fff74221e68) at /home/tsteven4/work/packetfiend/main.cc:777
```

PacketFiend:
You're correct, I hadn't considered the case when merging multiple IGC files - I didn't even realize this was possible, so assumed any extension data present will be present for all Waypoints. This issue occurs when merging tracks from recorders using non-identical sets of extensions.

This fix is an igc_fsdata::has_value() function that still allows for IGC extensions to be processed in a single block. I *have* noticed some function templates (not sure what they're called) in [defs.h starting at line 385](https://github.com/GPSBabel/gpsbabel/blob/cad16e14c34f3ee16fde021aa4df82b2461f5bff/defs.h#L385), but that implementation is beyond me currently.

Another `QMap<IgcFormat::ext_rec_type_t, igc_fsdata::std::optional*> igc_opt_fsdata_map` could also work, but I think this way is more readable.

Should we add a test that covers this?

Thanks for bearing with me. At least in this case I think it was my unfamiliarity with GPSBabel and not my out of date C++.

Fixes #1062

### DIFF
--- a/kml.cc
+++ b/kml.cc
@@ -1505,20 +1505,14 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_fxa:
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
-    case wp_field::igc_acz: {
-      if (fs_igc) {
-        if (fs_igc->has_value(member)) {
-          double value = fs_igc->get_value(member).value();
-          if (global_opts.debug_level >= 6) {
-            printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
-          }
-          writer->writeTextElement(QStringLiteral("gx:value"), QString::number(value));
-        } else {
-          if (global_opts.debug_level >= 7) {
-            printf(MYNAME ": Writing empty KML SimpleArray data for %s\n", name);
-          }
-          writer->writeTextElement(QStringLiteral("gx:value"), QString());
+    case wp_field::igc_acz:
+      if (fs_igc && fs_igc->has_value(member)) {
+        double value = fs_igc->get_value(member).value();
+        if (global_opts.debug_level >= 6) {
+          printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
         }
+        writer->writeTextElement(QStringLiteral("gx:value"), QString::number(value));
+        writer->writeTextElement(QStringLiteral("gx:value"), QString());
       // No igc_fsdata present, but we still need to write out the SimpleArray
       } else {
         if (global_opts.debug_level >= 7) {
@@ -1527,7 +1521,6 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
         writer->writeTextElement(QStringLiteral("gx:value"), QString());
       }
       break;
-      }
     default:
       fatal("Bad member type");
     }
--- a/kml.cc
+++ b/kml.cc
@@ -1512,7 +1512,6 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
           printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
         }
         writer->writeTextElement(QStringLiteral("gx:value"), QString::number(value));
-        writer->writeTextElement(QStringLiteral("gx:value"), QString());
       // No igc_fsdata present, but we still need to write out the SimpleArray
       } else {
         if (global_opts.debug_level >= 7) {
--- a/kml.cc
+++ b/kml.cc
@@ -1506,18 +1506,28 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
     case wp_field::igc_acz: {
-      if (fs_igc->has_value(member)) {
-        double value = fs_igc->get_value(member).value();
-        if (global_opts.debug_level >= 6) {
-          printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
+      if (fs_igc) {
+        if (fs_igc->has_value(member)) {
+          double value = fs_igc->get_value(member).value();
+          if (global_opts.debug_level >= 6) {
+            printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
+          }
+          writer->writeTextElement(QStringLiteral("gx:value"), QString::number(value));
+        } else {
+          if (global_opts.debug_level >= 7) {
+            printf(MYNAME ": Writing empty KML SimpleArray data for %s\n", name);
+          }
+          writer->writeTextElement(QStringLiteral("gx:value"), QString());
         }
-        writer->writeTextElement(QStringLiteral("gx:value"), QString::number(value));
-      } else if (global_opts.debug_level >= 7) {
-        printf(MYNAME ": Writing empty KML SimpleArray data for %s\n", name);
+      // No igc_fsdata present, but we still need to write out the SimpleArray
+      } else {
+        if (global_opts.debug_level >= 7) {
+          printf(MYNAME ": Writing empty KML SimpleArray data for %s\n", name);
+        }
         writer->writeTextElement(QStringLiteral("gx:value"), QString());
       }
       break;
-    }
+      }
     default:
       fatal("Bad member type");
     }
--- a/kml.cc
+++ b/kml.cc
@@ -1506,13 +1506,14 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
     case wp_field::igc_acz:
-      if (fs_igc && fs_igc->has_value(member)) {
+      if (fs_igc && fs_igc->get_value(member).has_value()) {
         double value = fs_igc->get_value(member).value();
         if (global_opts.debug_level >= 6) {
           printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
         }
         writer->writeTextElement(QStringLiteral("gx:value"), QString::number(value));
-      // No igc_fsdata present, but we still need to write out the SimpleArray
+      // No igc_fsdata present, but we still need to write out the SimpleArray.
+      // This can happen when merging tracks with different sets of IGC extensions.
       } else {
         if (global_opts.debug_level >= 7) {
           printf(MYNAME ": Writing empty KML SimpleArray data for %s\n", name);
--- a/kml.cc
+++ b/kml.cc
@@ -1506,9 +1506,8 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
     case wp_field::igc_acz: {
-      double value;
       if (fs_igc->has_value(member)) {
-        value = fs_igc->get_value(member).value();
+        double value = fs_igc->get_value(member).value();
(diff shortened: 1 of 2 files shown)


PacketFiend:
Further testing showed that this issue was still present when merging an IGC file **with** extensions with one **with no extensions** at all, due to the way igc_fsdata is stapled to the Waypoint with FsChainAdd(). I have a few IGC files from homebrew flight recorders that will cause this, if merged with another one with extensions. https://github.com/GPSBabel/gpsbabel/pull/1063/commits/7416399f9237d8107cb5cdb6394e8efc9d1c0c23 fixes this.

I'm seeing some weirdness when merging these tracks, but I'm chalking it up to this homebrew recorder's weirdness - notably, it only keeps timestamps with minute resolution (discarding the seconds), and it lacks extension data of any kind.

The issue would only present either when (and I'm not sure exactly which one of these is causing it)
- Merging a track *with* extension data to a track *without* it, or
- Merging a track with second resolution with one without

It's such an edge case though, I think it can be safely ignored. All commercially available flight recorders record seconds in their timestamps, and have at least one extension (FXA).

kml.cc
```diff
@@ -1506,18 +1506,28 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
     case wp_field::igc_acz: {
-      if (fs_igc->has_value(member)) {
-        double value = fs_igc->get_value(member).value();
-        if (global_opts.debug_level >= 6) {
-          printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
+      if (fs_igc) {
+        if (fs_igc->has_value(member)) {
```
tsteven4: ~~I don't think we need to check fs_igc here.  I think has_igc_exts implies fs_igc != nullptr, and kml_mt_simple_array is only called with these members when has_igc_exts is true.  But it would be good to add an assertion:
assert(fs_igc != nullptr);
Assertions express the programmers intention and in non-debug code are compiled out.  In debug code they will stop execution with a message if violated.~~

If we have to check fs_igc then can't we collapse this to
if (fs_igc && fs_igc->has_value(member)) {
  ...
} else {
  ...
}

tsteven4:
OK, I think I see why you tested fs_igc.

I think the homebrew recorder violates the spec, at least for a " IGC FRs".  In any event, we should try to avoid crashing even with malformed input.

What you noticed in defs.h at 385 is a set of [function-like macros](https://en.cppreference.com/w/cpp/preprocessor/replace).  This [isn't preferred in modern c++](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es31-dont-use-macros-for-constants-or-functions), it's an artifact of our history from c.  I recently did fiddle with it making it look more like the std::optional interface.

I don't think we need to add a test case.  I went hunting because of similar issues in other pre-exisiting members that I recently fixed.

kml.cc
```diff
@@ -1506,13 +1506,28 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
     case wp_field::igc_acz: {
-      double value = fs_igc->get_value(member).value();
-      if (global_opts.debug_level >= 6) {
-        printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
+      if (fs_igc) {
+        if (fs_igc->has_value(member)) {
```
tsteven4: Instead of implementing has_value(KmlFormat::wp_field type), or a QHash/QMap (which will fail with Qt5 as we discovered previously),
if (fs_igc && fs_igc->get_value(member).has_value()) {
  double value = fs_igc->get_value(member).value();
  ...
} else {
...
}

PacketFiend:
> If we have to check fs_igc then can't we collapse this to
if (fs_igc && fs_igc->has_value(member)) {

~~Apparently not. If I collapse that `if`, then no extension data gets turned into SimpleArrays at all in the final KML. I'm still trying to figure out why.~~

Nevermind. That was a brain fart. I left [this](https://github.com/GPSBabel/gpsbabel/blob/d4c32a2769bb6830b0a9551d17a5a072d85fd5a6/kml.cc#L1515) line in there.

PacketFiend:
It's not that this homebrew recorder violates the spec, per se. It's more that any logs it produces will not be "IGC/FAI certified", and therefore can't be used as evidence for record claims, badge attempts, or any kind of competition scoring. AFAICT the logs it produces are still valid on a technical level.

In my club's case, they will be useful for billing purposes and training flight analysis. The cheapest available "certified" recorders *start* at about $750 USD, hence why we're experimenting with rolling our own, which we can do for under $100.

tsteven4:
Can we kill the change to igc.h in its entirety, and just change your kml.cc as follows?  fs_igc->get_value(member) exist and returns a std::optional<double>.  We can test this optional to see if it has a value irrespective of which member we are talking about.
```
diff --git a/kml.cc b/kml.cc
index 95b2d0e9..09d4dc33 100644
--- a/kml.cc
+++ b/kml.cc
@@ -1506,7 +1506,7 @@ void KmlFormat::kml_mt_simple_array(const route_head* header,
     case wp_field::igc_gfo:
     case wp_field::igc_siu:
     case wp_field::igc_acz:
-      if (fs_igc && fs_igc->has_value(member)) {
+      if (fs_igc && fs_igc->get_value(member).has_value()) {
         double value = fs_igc->get_value(member).value();
         if (global_opts.debug_level >= 6) {
           printf(MYNAME ": Writing KML SimpleArray data: %s of %f\n", name, value);
```

PacketFiend:
Indeed we can, I just overlooked that possibility. https://github.com/GPSBabel/gpsbabel/pull/1063/commits/cc6f3c43becf40a6dcab799cb21d778b292bd050 tests fine for me on whatever combination of files I throw at it.

Aside, merging files presents the possibility of plotting a glider's path over an entire year. That could prove interesting.

robertlipe:
TBC, the writer can also see more or fewer points from even a single file
read in the presence of filters in the pipeline.

For example, simpilfy can toss points that don't materially change the
shape of your line and interpolate can add points that won't have your
extensions.

tsteven4:
This resolves #1062 .

PR #1065, opened 2023-04-12
Remove holux gm-100 format.

GPSBabelDeveloper:
Following through with plan announced Feb 28, 2022.
This was a serial unit from the turn of the century. Holux is out of business.
We've done nothing but mechanical maintenance on this format for years.


tsteven4:
We should delete reference/paris.wpo as well.  This was only referred to in the deleted file testo.d/holux.test

robertlipe:
Good call. Got it.

PR #1066, opened 2023-04-12
Removing baroiq and cambridge formats.

GPSBabelDeveloper:
Following through with plan announced Feb 28, 2022.
This was a format seemingly unused by anyone but the creator.
We've done nothing but mechanical maintenance on this format for years.

WIP: This PR is pending approval from PacketFiend


PacketFiend:
It looks like these are/were used by hang gliders and paragliders. I've certainly never seen one. They record height as a function of time, and that's it (a few other bells and whistles but that's the meat of it). I wouldn't even classify this as geospatial data.

I occasionally use something similar but without any kind of recording capability, just something that emits a tone that audibly tells me if I'm gaining or losing energy. While this barograph may serve a similar purpose, I highly doubt enough people would want to massage the data it produces - they'll just use something more capable.

So I would say, safe to mothball this because A) I can't see any demand for running its output through GPSBabel, and B) it does not record geospatial data, only height = f(time), and C) its functionality has been entirely replaced by [XCSoar](https://xcsoar.org/), which can run on a mobile phone.

Search for "Brauniger IQ Series Barograph" and the first two hits are the GPSBabel documentation...

robertlipe:
>
> Search for "Brauniger IQ Series Barograph" and the first two hits are the
> GPSBabel documentation...
>
That's indeed one of the criteria I use in determining how popular a
device/format is. 😉 If our doc is more popular that the maker's own,
reviews, or blogs from owners, we're usually the last one standing and it's
time to turn out the lights.

Thanx,
RJL

PR #1067, opened 2023-04-13
use constexpr if for trackfilter debug

tsteven4:


robertlipe (approved):

trackfilter.cc
```diff
@@ -456,7 +456,7 @@ void TrackFilter::trackfilter_split()
           fatal(MYNAME ": invalid time interval unit specified.\n");
         }
 
-        if (TRACKF_DBG) {
+        if constexpr(TRACKF_DBG) {
```
robertlipe, line 459: I checked compiler output and in every case, it was able to still parse the contents while optimizing away the body of that block.

Because TRACKF_DBG is static and so simple, I believed them equivalent outside of contexts like template substitution. Isn't this just redundant decoration? It's not like the defn of TRACK_EXPR is exactly obtuse for the code generator to see.

tsteven4: I believe "constexpr if" requires the body of the block to be discarded, and your implementation left it up to compiler optimization to do so as it saw fit.  Resharper will flag:

>       trackfilter.cc:23 Variable 'TRACKF_DBG' can be made constexpr.

It will also flag "If statement with constant condition can be replaced with 'if constexpr'', although it didn't find these which is a bit of a mystery.  I have seen that warning in the past.

tsteven4:
Resharper will flag
`if (false) {`
      trackfilter.cc:459 If statement with constant condition can be replaced with 'if constexpr'
But it misses
`        if (TRACKF_DBG) {`
when 
`static constexpr bool TRACKF_DBG = false;` or
`static const bool TRACKF_DBG = false;`

robertlipe:
So it can recognize

if (false) blah
but not
static cont bool foo = false
if (foo) blah

?

Interesting. It's probably "just" looking at a stream of tokens coming out
of a parser and not doing lifetime analysis on symbols the way a
self-respecting compiler would.

The optimizer knows the life of that symbol and it's never changing. (Well,
even then, it could be an extern, so it wouldn't know at link time. I guess
it's up to the tools to decide whether to leave that block in or out. Even
now (I first saw it on MIPS in the early 90's) optimizing linkers are rare
without a profiling cycle.

Based on all this and some additional reading, I think I've got the beat
now.

if constexpr(foo)

means you're PROMISING the compiler that foo is a compile-time const and
not leaving it to how resourceful the compiler is to figure out how const
it is - even if you've left strong bread crumbs for it, like I did.

Thanx for the improvement in the code-review and the resulting education.

PR #1074, opened 2023-04-20
introduce route_head::empty()

tsteven4:


robertlipe:
Maybe too late but how about isEmpty so we don't confuse it with clear

tsteven4:
I sort of agree that using verbs/adectives for modifing/reporting make sense, but I followed STL:confounded:

In general, WaypointList and RouteList are using the STL versions, and not the QT versions.  Qt isEmpty is STL empty.
STL uses empty to check whether the container is empty and clear to clear the contents.  WaypointList and RouteList aren't using a clear, they use flush which I think is our history.  They are derived from PRIVATE QList<T>, so only the methods that are explicitly included with using are available.

Anway, empty/isEmpty shows some real advantages with all the unsigned/signed count/size variability.  >0, <= 0, ==0, != 0 all reduce to a bool independent of the containers use of a signed or unsigned type for count/size.

tsteven4:
P.S. my motivation for using STL Container method names was they are supported by Qt containers, but not vice-versa.  I was trying to make a switch to STL Containers less painful.

GPSBabelDeveloper:
Good reasoning. My other, bigger, qt project had that same tension in
naming and style. It's just awkward.

PR #1085, opened 2023-04-25
hide inifile_t from users

tsteven4:


GPSBabelDeveloper:
I wonder if we should hide inifile in the deprecation bin.

PR #1089, opened 2023-04-27
use Qt datetime when writing gdb header.

tsteven4:


robertlipe:
LGTM.

But fwiw, I consider GDB to be hanging on by a thread and maybe not worthy
of maintenance. Better alternatives exist.

PR #1099, opened 2023-05-02
use QRegularExpression for xml generic tag matching.

### LINKED ISSUE #1098: Problem converting kml with large number of coordinates to osm
I have a kml file that has a bit of coordinates ([sq.zip](https://github.com/GPSBabel/gpsbabel/files/11374511/sq.zip)) and I would like to convert it to osm.
I tried following command line:
```
gpsbabel -i kml -f sq.kml -o osm,tag=highway:primary -F tiles.osm
```

But all I get is an almost empty tiles.osm file with just the main tag.
I tried with `-D 5` option (the docs miss what values can be put there so I tried with `5`), but didn't get any errors, I'm not sure what I'm doing wrong, or such file is maybe unsupported?

When I upload that kml file to online tools it shows the correct data on maps.

tsteven4:
The kml reader now creates tracks from innerBoundryIs elements in addition to the historic creation from outerBoundryIs elements.

This resolves #1098.


GPSBabelDeveloper:
Looks good.. strmatch is a string I've tugged at for a while, so Thanx for killing it AND making the kml. Reader better in useful ways.

PR #1100, opened 2023-05-06
Various modernizations in humminbird.

robertlipe:
- Make humminbird_icons[] a QStringList.
 - Various attempts to reduce extra allocations and hope that Qt template magic can
 - reduce some of those pointer/len buffer move when deailing with those fixed (not zero-terminated) buffers.
 - Try to more clearly mark when arrays represent fixed buffers on device files. I don't think the icon finder ever actually worked the way it was described in the comments.


humminbird.h
```diff
@@ -48,7 +49,7 @@ class HumminbirdBase
 
   /* Constants */
 
-  static constexpr const char* humminbird_icons[] = {
+  QStringList humminbird_icons = {
```
tsteven4: I would recommend this be a const QStringList.  That avoids any checks for detachment as well as alleviates the need to use qAsConst with this list, so you save the two uses you added in humminbird.cc as well.

humminbird.cc
```diff
@@ -245,25 +247,22 @@ HumminbirdBase::humminbird_read_wpt(gbfile* fin)
 
   auto* wpt = new Waypoint;
 
-  // Could probably find a way to eliminate the alloc/copy.
-  char* s = xstrndup(w.name, sizeof(w.name));
-  wpt->shortname = s;
-  xfree(s);
+  QByteArray ba = QByteArray(w.name, sizeof(w.name));
+  wpt->shortname = QString(ba).remove(QChar::Null);
```
tsteven4: Can this be
auto ba = QByteArray(w.name, qstrnlen(w.name, sizeof(w.name));
?
If so there are 2 more occurrences of this pattern below.

tsteven4: or better
wpt->shortname = QByteArray(w.name, qstrnlen(w.name, sizeof(w.name));

robertlipe: Yes! That case of having NO terminator kept haunting me. I wasn't happen with what I landed on (too many type conversions, too many many iterations over the buffer) but this is better.
 
I actually looked for a qnstrlen but not a qstrnlen. I see now you used this basic approach in garmin.cc where we have that same edge between fixed length, not terminated, buffers and QStrings. Experience!

robertlipe: This was better handled in the trunk, so I've accepted the trunk's version.

This PR has had a hard life. (owing to my short attention span) I think you and I have independtly recreated the meat of this a couple of times.

PR #1102, opened 2023-05-08
datetime purity.

tsteven4:
eliminate some QDateTime -> gpsbabel:DateTime conversions.  These can also help performance as DateTime defaults to UTC.

eliminate some unecessary involvment of the UNIX epoch when computing time differences.


GPSBabelDeveloper:
Nice.

Those are some really fine-grained nits...

PR #1103, opened 2023-05-09
refactor unicsv date time handling.

tsteven4:
move unicsv time handling from c-style legacy time support to Qt.

Traditionally this format has tried to handle times without a date.  This is problematic.  Without a date conversion between UTC and localtime is not possible.  With a fake date (1970/01/01) it is easy for other writers to not realize it is fake.  This module used to interpret datetimes from the first 24 hours of the unix epoch as only containing a time, but other modules did not.  I changed that to the first 48 hours to cover time zone conversions (using the assumed date, which could lead to DST errors).  However, perhaps it is better just to ignore time data when when don't also have date data.

It is more reasonable to assign time to the start of day when we receive a date without a time.  I continue to do so, although I have extended the utc option to apply to this case as well.

I also was careful not to mix utc date/times with local date/times.


GPSBabelDeveloper:
Nice. This has been a tarpit for us for a while. I picked at it a bit, but
ran away.

tsteven4:
Should we ignore times when we don't get a date instead of inventing a date (1970/01/01)?

I have similar changes to xcsv running.

robertlipe:
In old times, we never really had a concept of a time without a date.  We
now have a sensible way to represent that, so I'm OK with that.

There are a lot of formats where I don't really know how to represent that
concept, though. Maybe for some formats we do still have to make up a date
and 1-1-1970 is a pretty widely recognized signpost for that. We can cross
that bridge if we have to figure out validating a GPX file that comes from
a unicsv "2pm" or something. Let's not deadlock on it now.

tsteven4:
This now brings similar changes to the xcsv reader.  A difference is the unicsv concept of needs_time, needs_date is not used with xcsv.  Thus generally unicsv prefers the first source, while xcsv prefers the last.

robertlipe:
Killing those two functions makes me happy.

Inconsistency between those two doesn't bother me. One we control tightly.
One we have no control over at all. Different parsing problems can have
different optimal solutions, IMO.

tsteven4:
Please see https://github.com/GPSBabel/gpsbabel/pull/1114 which is favored over this PR.

tsteven4:
Closed in favor of #1114, which included this.

PR #1106, opened 2023-05-12
xcsv date/time handling corrections, test, and documenation

tsteven4:
For the xcsv reader & writer:

- avoid printing date/time fields when there isn't a valid date/time.
- avoid warning on reading date/time fields when the input is empty.
- when printing  with HMSG_TIME, HMSL_TIME use times from a 12 hour clock.  If a fourth field is provided we print ampm.  It doesn't make sense to print times from a 24 clock and then print ampm.  Switching between 12 and 24 hour clock would require scanning the format string.  Printing with a 24 hour clock can be done with GMT_TIME and LOCAL_TIME by using "%H%M%S".  **Do we define HMSG_TIME, HMSL_TIME to always use a 12 hour clock? For a 12 hour clock how do we allow for printing hours only, hours and minutes, or hours and minutes and seconds with the required ampm?** (needs further documentation/work)
- fix addhms to read from a 12 or 24 hour clock based on the existence of the fourth field, ampm.  This means that for a 12 hour clocks hours, minutes and seconds are required.  For a 24 hour clock minutes and seconds are optional, and ampm is prohibited. Reading from a 24 hour clock could be done with GMT_TIME and LOCAL_TIME by using "%H%M%S".  D**o we define HMSG_TIME, HMSL_TIME to always use a 12 hour clock? For a 12 hour clock how do we allow for reading hours only, hours and minutes,  or hours and minutes and seconds with the required ampm?** (needs further documentation/work)
- add support for printing dotnet time.
- use QDateTime::toString to print ISO time.
- use QDateTime::fromString instead of xml_parse_time when reading ISO datetimes.  This avoids xml_parse_times intentional odd behavior of treating non-timezoned times as UTC.
- allow GMT_TIME to accumulate before or after HMSG_TIME like LOCAL_TIME and HMSL_TIME.  This worked long ago, but has been broken for some time.
- restore YYYYYMMDD to use UTC as mentioned in 2012 in https://sourceforge.net/p/gpsbabel/mailman/message/29544538/
- enhance excel time conversions to use any fractional second information.

For the unicsv readers:

- use QDateTime::fromString instead of xml_parse_time when reading ISO datetimes.  This avoids xml_parse_times intentional odd behavior of treating non-timezoned times as UTC.

test:

- add a test of the xcsv reader and writer time related fields.  This only runs in America/Denver time zone.  As setting the time zone is system dependent the test only runs if tzselect is available.  If so it assumes America/Denver is available.

documentation:

- correct documentation to give a sensible format for xcsv field GMT_TIME.
- improve description of HMSL_TIME, HMSG_TIME.
- clarify that YYYYMMDD is UTC.


tsteven4:
Please see #1114 which is favored over this PR.

tsteven4:
Closed in favor of #1114, which included this.

PR #1111, opened 2023-05-14
workaound curl bug in lastest windows github runner.

tsteven4:
workaround  https://github.com/actions/runner-images/issues/7329 which was causing windows build to fail to upload the installer to the continuous release.

### DIFF
--- a/tools/uploadtool/upload_github.sh
+++ b/tools/uploadtool/upload_github.sh
@@ -132,9 +132,11 @@ echo "Upload binaries to the release..."
 for FILE in "$@" ; do
   FULLNAME="${FILE}"
   BASENAME="$(basename "${FILE}")"
+  # use -http1.1 to avoid https://github.com/actions/runner-images/issues/7329
   curl -H "Authorization: token ${GITHUB_TOKEN}" \
        -H "Accept: application/vnd.github.manifold-preview" \
        -H "Content-Type: application/octet-stream" \
+       --http1.1 \
        --data-binary @$FULLNAME \
        "$upload_url?name=$BASENAME"
   echo ""


GPSBabelDeveloper:
LGTM

Nice find.

tsteven4:
Found while reviewing #1109.  I will have comments.

tsteven4:
How about the continuous release for Linux of a snap?  It's already been used by a problem reporter to verify his fix.  We seem to be getting a rash of osm format issues, and the reporters seem to be on ubuntu.

robertlipe:
I'm clammy on binary releases for Linux, but if you think the ABI is
actually stable now, go for it. The system qt being too broken for the GUI
is certainly not a comforting sign.

I just really don't want to be disassembling libc and finding that some
security update inserted a field in the middle of a binary structure or
something. We left that horror before.

Let's set clear rules in what's expected to be supported because I also
don't want to be in the distro installing/testing biz.

tsteven4:
This is my third try or so with snapcraft, over many years.  The first two flamed out, but this time the CLI came together almost at once.  I think having a Linux installer is useful.  A number of recent problems were reported on Linux, and it is a burden that may not be overcome to ask the reporter to build gpsbabel from source to verify a fix.  In one instance the reporter built from source, in the other I got the snap online before they were able to build from source, and they used the snap to verify the fix.  If the reporter isn't regularly building gpsbabel then it is much easier to use the snap built by CI.

The GUI is more complicated than our CLI as it uses plugins.  I eventually learned enough to get by this.  But QWebEngine adds another layer that I haven't conquered yet.  In any event, for testing a pre-release the CLI is likely what all we need.  Let's see how that flies.

BTW the snap should portable, a single snap should be usable on not just multiple versions of Ubuntu, but Debian, Fedora, Arch, and more.

robertlipe:
OK, we'll see how it goes. I've just been burned by this so hard in the
past that I'm gun-shy.

I'll admit to long being skeptical that 3,400 people build and run from
source with rarely a question asked, even when our instructions have been
wrong.

PR #1068, opened 2023-04-13
de-duplicate format and filter vector display code.

tsteven4:

PR #1116, opened 2023-05-15
restore tab when display vector formats.

tsteven4:
this was lost in #1068
update reference files to be exactly what the test produces, including exact whitespace, trailing tabs and no newline at end of file.


GPSBabelDeveloper:
Sorry. My editor configs (both vi and Sublime) - at least on one computer -
"fix" those before I even see them.

PR #1117, opened 2023-05-15
fix garmin serial/usb device character encoding/decoding.

### LINKED ISSUE #1115: Converting encoding of waypoints names
Hello. 

I'm trying to use GPSBabel for importing and exporting tracks and waypoints from/to Garmin GPSmap 60cx. As other older units it doesn't support unicode. 
Is it possible to make on-the-fly conversion of names and comments from 8-bit encoding to UTF-8 and vice-versa? E. g. from Win-1251 to UTF8?

So far I'm getting waypoints like these in .gpx:
```
  <wpt lat="57.142150495" lon="33.117158515">
    <name>5 ÊÀËÈÍÈÍÀ</name>
    <cmt>5 ÊÀËÈÍÈÍÀ</cmt>
    <desc>5 ÊÀËÈÍÈÍÀ</desc>
    <sym>Lodging</sym>
  </wpt>
```

tsteven4:
We got away with breaking this for a long time, but we finally got caught.

This resolves #1115

@svintuss has tested both read and write on his Garmin GPSmap 60cx using windows-1251.


GPSBabelDeveloper:
On phone, but I think I like it.
Is there a way to connect this to the switch on device id that sets all the things we "know" about eath model like like waypoint length and lower case capability (a loose proxy for CP -125x if we have nothing better?
I think if we send a gpx with international characters to something like a gps II or a. 12 or a 3 or something, the device will crash,. So we should be conservative in doing that, but the USB -capable devices are less broken.


... And most of my hyperactive heart attack in this code is in the layer below. It's mostly the jeeps level of stuff that can break a quarter of the world. A streetpioli and a 60 are both USB, but very different in the wire. Heck, a 60c and 60cx are very different.

tsteven4:
By default the receiver charset is US-ASCII.  The switch on gps_waypt_type and gps_save_id may update this to windows-1252.  After all that if the codec option was used we will override the automatically selected codec with whatever was asked for in that option.

At this point on all platforms, with Qt5 or Qt6, we should have access to at least the minimal set of "supported encodings" promised by [Qt5](https://doc.qt.io/qt-5/qtextcodec.html).  This works on Qt6 because we are using the Qt 5 Core Compat module.  Included in the minimal set of supported encodings is Windows-1250 to 1258.

If you feel like it checking the lengths of the converted strings in the writer might be helpful.  Also looking for any implicit conversions in the reader I missed (I did catch a couple).

robertlipe:
Is his ID missed in the switch that would have auto selected 1252? Maybe
it's a localized device with a different ID. Garmin does that. We've had to
special cases devices in the market for Japanese and such before. They
would get the default, disallowing spaces and lowercase and such.

We'll lose qt5cimoat with qt7 ... or before. Without that do we have at
least minimal common codecs? (I don't much care about the full list of 200
or whatever that you just posted. 125x, latin1, utf8/16, and ...)

tsteven4:
I don't think it is missed.  Before I added the codec option he was seeing the line
 fprintf(stdout, "reciever charset detected as %s.\r\n", receiver_charset);
print windows-1252.

I think were good with Qt5Compat until Qt7, then I think we will only have codecs if Qt is built with ICU support.  And I don't think the Qt supplied builds don't have ICU support enabled on all platforms.

garmin.cc
```diff
@@ -347,7 +368,26 @@ rw_init(const QString& fname)
...
+
+  /*
+   * Beware, valid_waypt_chars shouldn't contain any character class metacharacters,
+   * i.e. '\', '^', '-', '[', or ']'
+   */
+  assert(!QString(valid_waypt_chars).contains('\\'));
+  assert(!QString(valid_waypt_chars).contains('^'));
+  assert(!QString(valid_waypt_chars).contains('-'));
+  assert(!QString(valid_waypt_chars).contains('['));
+  assert(!QString(valid_waypt_chars).contains(']'));
```
robertlipe, line 388: These could be a single QString::contains(const [QRegularExpression](https://doc.qt.io/qt-6/qregularexpression.html) &re, but it would admittedly be a quoting nightmare and wouldn't give you a single error message like this will. Your call.

garmin.cc
```diff
@@ -347,7 +368,26 @@ rw_init(const QString& fname)
    * However, this is still used for garmin_fs_garmin_after_read,
    * garmin_fs_garmin_before_write.
    */
+  if (opt_codec != nullptr) {
+    // override expected codec with user supplied choice.
+    receiver_charset = opt_codec;
```
robertlipe, line 373: This is a good option - though we should set low expectations for it in the doc - but if he has an ID 292 receiver, why does he need to specify this option?

Or was that just for testing? I'll admit I didn't follow the play-by-play.

garmin.cc
```diff
@@ -347,7 +368,26 @@ rw_init(const QString& fname)
...
+   * Beware, valid_waypt_chars shouldn't contain any character class metacharacters,
+   * i.e. '\', '^', '-', '[', or ']'
+   */
+  assert(!QString(valid_waypt_chars).contains('\\'));
+  assert(!QString(valid_waypt_chars).contains('^'));
+  assert(!QString(valid_waypt_chars).contains('-'));
+  assert(!QString(valid_waypt_chars).contains('['));
+  assert(!QString(valid_waypt_chars).contains(']'));
+  invalid_char_re = QRegularExpression(QStringLiteral("[^%1]").arg(valid_waypt_chars));
+  assert(invalid_char_re.isValid());
```
robertlipe, line 390: I'll grant we should never ever see these asserts failing, but triggering a failing assert inside rw_init - or after - can leave the device in a dangling state that can require power cycling the receiver to recover from. This same issue is shared with the calls to fatal(), so I won't argue with it strongly.

This is one of those rare textbook cases where exceptions might actually make sense. Our own _deinit() doesn't really do any meaningful cleanup, for much the same reason dtors aren't called on failed ctors.

Then again, we're not flagrantly in the middle of a _Write_Packet/Get_Ack cycle (which I KNOW will hang many devices).

Meh. It's an exceptional case and I don't really know that we can do anything else better.

No action required, but this may be educational, so I'm not deleting it...

garmin.cc
```diff
@@ -418,8 +458,8 @@ waypt_read()
   for (int i = 0; i < n; i++) {
     auto* wpt_tmp = new Waypoint;
 
-    wpt_tmp->shortname = QString::fromLatin1(way[i]->ident);
-    wpt_tmp->description = QString::fromLatin1(way[i]->cmnt);
+    wpt_tmp->shortname = str_to_unicode(way[i]->ident);
```
robertlipe, line 461: Should we give those foo[NNN]'s in jeeps/gps.h a better type than 'char' so we could add some meaningful overrides here and not live in constant fear of failing to zero pad them and/or character convert them?

Optional.

reference/help.txt
```diff
@@ -137,6 +137,7 @@ File Types (-i and -o options):
 	  category              Category number to use for written waypoints
 	  bitscategory          Bitmap of categories
 	  baud                  Speed in bits per second of serial port (baud=9600
+	  codec                 override codec to use for device
```
robertlipe, line 146: Uppercase, please.
Needs a corresponding option_codec.xml in docbook.

garmin.cc
```diff
@@ -134,8 +142,22 @@ static int d103_icon_number_from_symbol(const QString& s);
...
+{
+  return codec->fromUnicode(qstr);
+}
+static QString str_to_unicode(const QByteArray& cstr)
+{
+  return codec->toUnicode(cstr);
+}
+
+static void
+write_char_string(char* dest, const char* source, size_t destsize)
```
robertlipe, line 155: Good idea. I should have done that 22 years ago. :-)
Should it have a sibling that does the character set conversion from the source?
I don't know if you got them all, but almost every occurrence of CSTRc in this source file is a candidate for replacement thusly, especially those as arguments to memcpy or strncpy.

robertlipe (approved):
I quite like it. Merge it down whenever you're ready.

garmin.cc
tsteven4: Yes, and it would save a bunch of QString construction.  But, the code is only run on init, and only in debug builds, so I thought clarity was better.

garmin.cc
tsteven4: Because he is using windows-1251 to get Cyrillic characters instead of the default windows-1252 Latin.

garmin.cc
tsteven4: If we were starting over certainly. We do treat some as zero padded and require zero termination, and others (in fs data) as zero padded but not necessarily zero terminated.  To me it isn't really clear what the requirements are, or even if the requirements were constant across all the devices.  The fs data ones generate Wstringop-truncation warnings.  But I'm not messing with jeeps today.

garmin.cc
tsteven4: If we assume the source is zero terminated, which we have, then the sibling isn't necessary.

tsteven4:
@robertlipe,
This seems fishy to me.  If we must upper, we do.  Then we throw away everything but A-Z 0-9, and maybe space, + and -.  If we don't force to uppercase we throw away a-z.  Was that the intent or should a case insensitive compare have been done?
https://github.com/GPSBabel/gpsbabel/blob/488aa41264c3c94c8c6e1d3de41ebd0c54d0c0d9/garmin.cc#L1012-L1021

garmin.cc
tsteven4: Its kind of tacky to report a code problem as a warning, but that would avoid leaving the device dangling.

reference/help.txt
tsteven4: done

garmin.cc
robertlipe: It would have just been a single place to wrap the toLatin/fromUtf style converters. That's all.

garmin.cc
robertlipe: Good call.
The original JEEPS API used zero-padded buffers everywhere and we had all kinds of fits when initially using that with C strings. It's very likely that the newer code (e.g. D108, D109) uses C strings instead of zero-padded ones like D103 originally would have used. It's likely currently a mixture and that's indeed messy.
If we were starting over with jeeps, it would be without jeeps and the Garmin protocol implementations would look very very different. Running away is a a good decision. The payoff in picking at this scab isn't worth it.

garmin.cc
robertlipe: I *think* this is early enough that it won't, in practice, hang. Since the triggered case would be 100% here, hanging the device or not isn't a big deal if it even DOES hang the device.

robertlipe:
You're right. That loop at 1012 is underachieving. We toss those those instead of "fixing" them, but such a receiver would have surely had 'must_upper' set, so it's presumably a very minority case.

I'd expect the right thing to do here is replacing this with a call to write_char_string after a conditional toUpper on the string if mustupper is set.

I think you're done here. Nice work!

tsteven4:
@robertlipe are you good with this resolution of rtept names?  For wpt names we defeat the cleaning function of mkshort when receiver_must_upper is false, see the similar comment in the code.
https://github.com/GPSBabel/gpsbabel/blob/911ab99766b7eb84336be8863f857ffd872ae4f6/garmin.cc#L1058-L1065

robertlipe:
Yes. It's definitely less wrong than what we already have, so commit this,
please.

There may be other dragons lurking, but these are dark, dark corners so
let's not deadlock on "perfect".

Thanx!

PR #1118, opened 2023-05-21
Cosmetic cleanups in defs.h

robertlipe:
Eliminate use of tabs in defs.h
Try to make comments into sentences when I had a reason to touch them.
More liberal use of // comments so editors can reflow.


robertlipe:
Nothing here is worth the merge. Rescinding.

PR #1120, opened 2023-06-02
Demand format before filename

jidanni:
If the user doesn't adhere to this he will get an error.


robertlipe:
Thanx, but this is already covered in the doc.

https://www.gpsbabel.org/htmldoc-1.8.0/Invocation.html#:~:text=Options%20are%20always%20processed%20in%20order%20from%20left%20to%20right

PR #1125, opened 2023-06-10
add fedora 38 to CI

tsteven4:


tsteven4:
g++ 13.1.1 does have a new false positive -Wuse-after-free warning in util.cc related to the use of xrealloc in xvasprintf.

tsteven4:
xrealloc can't return a nullptr because of the conditional call to fatal, thus the warning is a false positive.
 ```
In function ‘void xfree(const void*)’,
    inlined from ‘int xvasprintf(char**, const char*, __va_list_tag*)’ at /__w/gpsbabel/gpsbabel/util.cc:261:12:
/__w/gpsbabel/gpsbabel/util.cc:88:7: warning: pointer ‘buf’ may be used after ‘void* realloc(void*, size_t)’ [-Wuse-after-free]
   88 |   free(const_cast<void*>(mem));
      |   ~~~~^~~~~~~~~~~~~~~~~~~~~~~~
In function ‘void* xrealloc(void*, size_t)’,
    inlined from ‘int xvasprintf(char**, const char*, __va_list_tag*)’ at /__w/gpsbabel/gpsbabel/util.cc:257:42:
/__w/gpsbabel/gpsbabel/util.cc:132:28: note: call to ‘void* realloc(void*, size_t)’ here
  132 |   char* o = (char*) realloc(p, s);
      |  
```

robertlipe:
Will experiment with nonull and not_null features here.

our xvasprintf is excessively clever. It was a pain to perfect because of
ambiguities in the return code where POSIX went one way and MSVC went
another (of course...).

If it can't see through our xFOO wrappers, maybe it's time to drop them
inside xvasprintf. A failed malloc is now caught as an exception and blows
up the program, just like xmalloc used to do.

Oh, and FIRSTSIZE should absolutely be set higher than 1.

I can look at it later this weekend if you don't beat me to it.

RJL

tsteven4:
were close to being able to delete xasprintf entirely, which would let us delete xvasprintf.  do you want to kill all that gdx_jmp_buf stuff?

GPSBabelDeveloper:
I'm AFK, but I think gdx was the stuff that let a Nufi (for example) be
treated like a serial device...or maybe a 60 to be treated like a file
device. I'm hazy.

Protocol devices haven't been made in many years.If that's the correct user
of that code, i'm fine with it dying.

tsteven4:
#1126 eliminates xasprintf, xvasprintf entirely.

robertlipe:
Excellent. That was painful code to write.

PR #1126, opened 2023-06-10
fix humminbird icon matching,

tsteven4:
eliminating one use of xasprintf.
This was broken in 6029c8267 over a decade ago.


tsteven4:
This linearly adds to the inefficiencies in the garmin_txt read side, but maybe that is not worth bothering with.

robertlipe:
It's not.

Im not sure that the person that wanted to keep it is even actually using
it or if they were just voting for "a garmin text file".

RJL

PR #1127, opened 2023-06-10
drop support for filesystem access via garmin usb.

tsteven4:
History from a private discussion 1/28/2023 between @tsteven4 and @robertlipe : 

    Can we kill the garmin setjmp/longjmp hijack of the gpx format and have
    users reference the found gpx files directly?

Yes. That bit was excessively clever. It was meant to give apps (like earth) that had the ability to call gpsbabel without a file picker the ability to still read and write magellans and garmins. Now that GPX devices have, by far, taken over the market, we can nuke that.


GPSBabelDeveloper:
Thanx. That code was painful to write. Now that Earth has file pickers, I
think we can safely do without that code.

I imagine anyone that called us handled the mass storage/vendor protocol
thing many years ago by now.

RJL

PR #1133, opened 2023-06-25
Strptime update

tsteven4:
This update strptime to glibc 2.37.

The license for strptime is now LGPL 2.1 instead of LGPL 2.0.  If this is not acceptable we may be able to license it under GPL 2.0.

See strptime/README.GPSBabel.

The modifications required are in strptime/strptime_l.c.patch:
1. langinfo.h may not exist, don't include it.
2. __strncasecmp_l may not exist.  we aren't using passed locales anyway, so just use strncasecmp.
3. sntrncasecmp may not exist.  add a function to replace it.  this is the same as we have done previously.
4. tm->tm_gmtoff may not exist.  don't use it.
5. era handling.  we don't have information on the era structure.   remove this code as we did previously (although the code is slightly different).

This solves warnings from clang 15 about -Wdeprecated-non-prototype related to upcoming changes in C23.  With the MSVC compiler a number of other warnings crop up and are disabled in strptime.cmake.


strptime/strptime_l.c.patch
```diff
@@ -0,0 +1,80 @@
+--- ../../glibc/time/strptime_l.c	2023-06-25 08:08:49.327979261 -0600
++++ strptime_l.c	2023-06-25 09:20:10.503668377 -0600
+@@ -21,7 +21,7 @@
+ 
+ #include <assert.h>
+ #include <ctype.h>
+-#include <langinfo.h>
```
robertlipe, line 7: c++17 has 
#if has_include <langinfo.h>
#include <langinfo.h.>
#endif
...but if we don't need it, we don't need it. {shrug}

strptime/strptime_l.c.patch
```diff
@@ -0,0 +1,80 @@
...
+      result; })
+ #else
+ /* Oh come on.  Get a reasonable compiler.  */
+ # define match_string(cs1, s2) \
+-  (strncasecmp ((cs1), (s2), strlen (cs1)) ? 0 : ((s2) += strlen (cs1), 1))
++  (case_ignore_strncmp ((cs1), (s2), strlen (cs1)) ? 0 : ((s2) += strlen (cs1), 1))
++/* now now, don't need to be rude .... */
++static
++int
++case_ignore_strncmp(const char *s1, const char *s2, int n)
```
robertlipe: I don't *think* I wrote the above (it does sound like me...) but in case this is for MS (and it always is) SO says that the current wisdom to #define a passthrough to _strnicmp

It really is pretty embarrassing how crude the string functions are in C 50 years later. Just sayin'

robertlipe (approved):
LGTM++ 
Thank you.
Was this really easier than flushing out our shim to QDateTime's toStrting/fromString?

strptime/strptime_l.c.patch
tsteven4: nice idea, but this is c.

strptime/strptime_l.c.patch
tsteven4: now using _strnicmp

PR #1135, opened 2023-07-17
eliminate ppa for qlibraryinfo bug

tsteven4:


tsteven4:
The fix was released in qt6-base-dev (6.2.4+dfsg-2ubuntu1.1) which is available in jammy-updates.

robertlipe:
Reviewing that thread on the launchpad Ubuntutracker just makes me sad.
Thank you for sticking with it so long. You gave them several different
fixes, tested, confirmed, and wrapped in a bow and despite the pleading of
two package owners, well, it was just unfortunate.

Since Jammy's support window is closing anyway, does this help unblock the
turndown of the Qt5 builds?

tsteven4:
The end of standard support for jammy is April 2027.  But this does help us move towards Qt6, as the latest LTS release (jammy) from Ubuntu now works with the Ubuntu supplied Qt6.  Let's release after merging #1060 and move the Qt bar up to 6.2.

PR #1090, opened 2023-04-28
C-style date and time -> QDateTime in wbt format.

tsteven4:
This also enhances the wbt tests so they look at the timestamps!

PR #1136, opened 2023-07-18
Deprecate wbt

tsteven4:
This was discussed in #1090.


robertlipe (approved):

tsteven4:
which leaves the discussion in #1090 about navilink, sbp, sbn

PR #1137, opened 2023-07-18
Deprecate navilink, sbp, sbn formats

tsteven4:
The potential for this was discussed in #1090.  When and if we are ready to deprecate these formats we should merge this PR.


robertlipe (approved):

tsteven4:
This was done in #1138

PR #1139, opened 2023-07-19
reduce complexity of mkshort from O(n^2) to O(n).

tsteven4:
Here is some test data with and without this PR.  time(ref) is without changes to mkshort, time(dut) is with the changes to mkshort in this PR.
The test is 
./gpsbabel -i gpx -f Auckland-numbers.gpx -x validate,debug -o gdb -F Auckland-numbers.gdb
However, in both cases the other O(n^2) issue in the gdb writer is bypassed:
```
@@ -1513,6 +1552,7 @@ GdbFormat::write_waypoint_cb(const Waypoint* refpt)
 // but, but, casting away the const here is wrong...
   (const_cast<Waypoint*>(refpt))->shortname = refpt->shortname.trimmed();
   Waypoint* test = gdb_find_wayptq(&wayptq_out, refpt, 1);
+  //Waypoint* test = nullptr;
```
mkshort_opt
mkshort_opt_zoom

### DIFF
--- a/mkshort.cc
+++ b/mkshort.cc
@@ -25,8 +25,9 @@
 
 #include <QByteArray>  // for QByteArray
 #include <QChar>       // for QChar, QChar::ReplacementCharacter
-#include <QList>       // for QList
+#include <QHash>       // for QHash, QHash<>::iterator, qHash, QHash<>::size_type
 #include <QString>     // for QString
+#include <Qt>          // for CaseInsensitive
 #include <QtGlobal>    // for foreach
 
 #include "defs.h"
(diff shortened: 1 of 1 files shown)


tsteven4:
Here is a fit over more iterations.  It looks like mkshort costs ~22 minutes for the full data set.  It isn't clear how much gdb_find_wayptq is costing, presumably much more.

mkshort_opt_long

tsteven4:
Above test data had an O(n) hack for gdb_find_wayptq. Still, the difference between the plots is all mkshort.  But taking out the gdb_find_wayptq hack will be O(n^2) and dominate.

tsteven4:
I assessed the cost of the gdb_find_waypt issue, on my machine it is ~101 minutes for the entire data set.  R^2=0.9996 with a polynomial of degree 2, O(n^2) confirmed.  So the mkshort fix alone is about 18% of the issue at n=650,000, corresponding to the ratios of the coefficients of the n^2 terms for the two drivers.

robertlipe (approved):
Excellent research and solution. Thank you!

Your findings and proposed solutions very much match my own. I'm still trying to find a happy place with the the other one. Your qHash use is very clever. I like that and it may become part of that mentioned happy place.

I'm a little smug that this code, which is probably some of the oldest in the project, had such a close mapping to the Qt/C++ replacements, but I see now why for this data set my own hash_string() performed worse than strcmp() would have. For long keys, it just has to walk the entire string and that's just brutal on L3 cache misses for data like this and a fixed bucket size of "37" just doesn't scale to 650,000 items even if they were perfectly distributed agains collisions. (Probably 'just' realloc-ing the bucket list by doubling it a few times as needed  My little (in 2002) baby grew up.  

OTOH, with the benefit of 2023 web search for string hashes and trying to find the origin of "((hash<<5) ^ (hash>>27)) ^ *key;" and the best-ranked article is a good description of hashing that uses this very function as an example of one that sucks. "allows a function to slip through sanity checks, like hashing an English dictionary uniformly, while still frequently bombing on user data." - I remember using the dictionary as a test case. :-/

Anyway, this is good stuff. Thank you.

tsteven4:
I have a fix for gdb_find_wayptq in hand, for read/write/waypoints/routes.

PR #1141, opened 2023-07-20
reduce complexity of gdb waypt searchs from O(n^2) to O(n)

tsteven4:
For the reader two parallel hashes are maintained with different
keys but identical values.  This is necessary because sometimes
we want to match the name and position, and other times we just want
to match the name.

### DIFF
--- a/gdb.cc
+++ b/gdb.cc
@@ -29,7 +29,6 @@
 #include <QByteArray>               // for QByteArray, operator==
 #include <QDate>                    // for QDate
 #include <QDateTime>                // for QDateTime
-#include <QList>                    // for QList<>::const_iterator, QList
 #include <QString>                  // for QString, operator!=, operator==
 #include <QTime>                    // for QTime
 #include <Qt>                       // for CaseInsensitive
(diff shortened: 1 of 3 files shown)


gdb.cc
```diff
@@ -70,19 +69,34 @@
...
+inline bool operator==(const GdbFormat::WptNameKey& lhs, const GdbFormat::WptNameKey &rhs) noexcept
+{
+  return lhs.shortname.compare(rhs.shortname, Qt::CaseInsensitive) == 0;
+}
+
+inline bool operator==(const GdbFormat::WptNamePosnKey& lhs, const GdbFormat::WptNamePosnKey &rhs) noexcept
+{
+  return
+    (lhs.shortname.compare(rhs.shortname, Qt::CaseInsensitive) == 0) &&
+    lhs.lat == rhs.lat &&
```
robertlipe, line 82: Aren't we more likely to have duplicates of names (especially names that are the same at the beginning) than by location? If so, consider changing the order of this test so that the numeric test is front of the short-circuit order.
Also, I know this is a behavioural change over The Way It's Always Been, but is a numeric comparison of double equality OK here? Im inclined to say that it is, as LHS and RHS would have the same incoming jitter when converted to doubles, but are they candidates for the abs(a-b) < kEpsilon trick?
...which may very well negate the first point.

OTOH, these numbers are hard to argue with and it's the way it's always been. {shrug}

gdb.cc
```diff
@@ -194,39 +208,45 @@ GdbFormat::gdb_fread_strlist() const
...
 Waypoint*
-GdbFormat::gdb_reader_find_waypt(const Waypoint* wpt, const char exact) const
+GdbFormat::gdb_reader_find_waypt(const Waypoint* wpt, bool exact) const
 {
-  Waypoint* res = gdb_find_wayptq(&wayptq_in, wpt, exact);
-  if (res == nullptr) {
-    res = gdb_find_wayptq(&wayptq_in_hidden, wpt, exact);
+  Waypoint* res;
+  if (exact) {
+    res = gdb_find_wayptq(waypt_nameposn_in_hash, wpt);
```
robertlipe, line 233: This is much better than what I started and hadn't yet really come back to.

When pitching this, be sure that e knows that this is all possible because of the hundreds of hours we put into moving this into C++ a long time ago. Having ready access to indisputably "better" hashes than ones we created (I created) ourselves is just so much more powerful.

gdb.h
```diff
@@ -63,6 +63,40 @@ class GdbFormat : public Format
   void write() override;
   void wr_deinit() override;
 
+  /* Types */
+
+  class WptNamePosnKey;
+  using WptNamePosnHash = QHash<WptNamePosnKey, Waypoint*>;
+  class WptNamePosnKey {
```
robertlipe, line 72: Nice.

robertlipe (approved):
Nice. I started down a similar path but didn't get it finished. I like this a lot.

Please send those guys an invoice. :-)

gdb.cc
tsteven4: yes, it mimics what has always been.  I'm going to treat this as out of scope, which was algorithm complexity improvements.

gdb.h
tsteven4: that one is ok, but the using of QHash::size_type was wrong!

tsteven4: CI rocks!

PR #1145, opened 2023-07-24
Remove final remnants of 'exported' in filters and writers.

robertlipe:
To be submitted after https://github.com/GPSBabel/gpsbabel/pull/1144/

PR #1146, opened 2023-07-25
Remove final remnants of 'exported'

tsteven4:
I think this is what was meant to be in #1145.


duplicate.cc
```diff
@@ -81,11 +81,6 @@ void DuplicateFilter::process()
 
   auto wptlist = *global_waypoint_list;
 
-  auto compare_lambda = [](const Waypoint* wa, const Waypoint* wb)->bool {
-    return wa->gc_data->exported > wb->gc_data->exported;
-  };
-  std::stable_sort(wptlist.begin(), wptlist.end(), compare_lambda);
-
   QMultiHash<QString, Waypoint*> wpthash;
   for (Waypoint* waypointp : wptlist) {
```
tsteven4: No need to copy the global_waypoint_list now that we aren't sorting, but care is required to avoid detach risks, i.e.
the for statement becomes:
for (Waypoint* waypointp : qAsConst(*global_waypoint_list)) {

tsteven4:
In duplicate.cc, the comment "It looks odd ..." relates to the deleted sort and should also be deleted.

tsteven4:
@robertlipe please take ownership of this PR, it's all your work that accidentally got into #1144 which I removed before merging.

duplicate.cc
tsteven4: fixed

tsteven4:
@robertlipe can we kill #1145 and merge this?  I resolved all my issues.

robertlipe:
Indeed. I made a mess fo that. Sorry for hte buy work. I'll to set fire to 1144 right now.

PR #1148, opened 2023-07-27
fix bugs with simplify filter

tsteven4:
1. With the length option the last point deleted took the total error over the specified limit.  In the test case with a modified reference file this brought the error to 0.62262 compared to the limit of 0.621371.  This is fixed, resulting in the change of a reference file.

2. When computing the total error (if another point is deleted) it was possible to refer to an xte record that needed to be updated due to the deletion of one of its neighbors.  This is fixed, the impact of another potential deletion is computed after the xte records are  updated.

3. Another issue has not yet been addressed.  The simplify filter ignores track segments and treats each track as if it was continuous.  If a track point that starts a track segment is deleted then that segment is joined with the previous segment.
Should the simplify filter treat each track as a continuous, ignoring track segments?  If so, should it produce a track that only has one segment?
On the other hand, if we choose to filter each segment independently, how would we set the error or count limits for each segment?

### DIFF
--- a/reference/simplify_error_length.gpx
+++ b/reference/simplify_error_length.gpx
@@ -516,6 +516,11 @@
         <time>2012-04-12T13:09:52Z</time>
         <speed>5.373000</speed>
       </trkpt>
+      <trkpt lat="48.190376647" lon="11.822674759">
+        <ele>541.600</ele>
+        <time>2012-04-12T13:09:59Z</time>
+        <speed>6.961000</speed>
+      </trkpt>
       <trkpt lat="48.190509081" lon="11.824321132">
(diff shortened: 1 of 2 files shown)


robertlipe:
You are, as usual, correct that trkseg is a bolted-on mess of an
afterthought. Given the use case, I think simplify should try to preserve
it, but not necessarily factor it into calculations.

In some fantasy land, tracks would be arrays of trksegs would be arrays of
points. You can see how retrofitting that in would be tough. So the 'hit
the ends with a highlighter pen' approach was taken a compromise to allow
some amount of segmentation but without reworking everything that touched
tracks.

PR #1149, opened 2023-07-28
Reduce complexity of simplify filter

tsteven4:
Reduce the complexity of the simplify filter from O(n^2) to O(nlog(n)) ~= O(n).  The complexity of this filter has been an issue for users in the past.

The test creates random routes of various lengths at uses the simplify filter to reduce the number of points by half:

[test.txt](https://github.com/GPSBabel/gpsbabel/files/12194618/test.txt)

The results are shown in the following plot.  It is difficult to distinguish between O(nlog(n)) and O(n) empirically, but O(nlog(n)) is expected for QMap.  Using the error option instead of count is a bit slower, but the complexity is the same.

simplify

Issues:

1. Unlike the current code route priority is ignored by the simplify filter.  route priority could be implemented if desired.
2. There is a test failure that is resolved by #1148 
3. track segment markers are inherited as done previously.
4. tracks are simplified ignoring any track segments as done previously.

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -492,7 +492,6 @@ public:
   void waypt_del(Waypoint* wpt); // a.k.a. erase()
   // FIXME: Generally it is inefficient to use an element pointer or reference to define the element to be deleted, use iterator instead,
   //        and/or implement pop_back() a.k.a. removeLast(), and/or pop_front() a.k.a. removeFirst().
-  iterator waypt_del(iterator it) {return erase(it);}
   void del_rte_waypt(Waypoint* wpt);
   void waypt_compute_bounds(bounds* bounds) const;
   Waypoint* find_waypt_by_name(const QString& name) const;
@@ -525,6 +524,7 @@ public:
   using QList<Waypoint*>::front; // a.k.a. first()
   using QList<Waypoint*>::rbegin;
   using QList<Waypoint*>::rend;
+  using QList<Waypoint*>::size_type;
 };
 
 const global_trait* get_traits();
@@ -674,6 +674,7 @@ public:
   void copy(RouteList** dst) const;
   void restore(RouteList* src);
   void swap(RouteList& other);
+  void swap_wpts(route_head* rte, WaypointList& other);
   template <typename Compare>
   void sort(Compare cmp) {std::sort(begin(), end(), cmp);}
   template <typename T1, typename T2, typename T3>
@@ -728,6 +729,8 @@ void route_add_wpt(route_head* rte, Waypoint* wpt, QStringView namepart = u"RPT"
 void track_add_wpt(route_head* rte, Waypoint* wpt, QStringView namepart = u"RPT", int number_digits = 3);
 void route_del_wpt(route_head* rte, Waypoint* wpt);
 void track_del_wpt(route_head* rte, Waypoint* wpt);
+void route_swap_wpts(route_head* rte, WaypointList& other);
+void track_swap_wpts(route_head* rte, WaypointList& other);
 //void route_disp(const route_head* rte, waypt_cb); /* template */
 void route_disp(const route_head* rte, std::nullptr_t /* waypt_cb */); /* override to catch nullptr */
 //void route_disp_all(route_hdr, route_trl, waypt_cb); /* template */
--- a/defs.h
+++ b/defs.h
@@ -448,17 +448,6 @@ public:
   gpsbabel::DateTime creation_time;
 
   wp_flags wpt_flags;
-  /*
-   * route priority is for use by the simplify filter.  If we have
-   * some reason to believe that the route point is more important,
-   * we can give it a higher (numerically; 0 is the lowest) priority.
-   * This causes it to be removed last.
-   * This is currently used by the saroute input filter to give named
-   * waypoints (representing turns) a higher priority.
-   * This is also used by the google input filter because they were
-   * nice enough to use exactly the same priority scheme.
-   */
-  int route_priority;
 
   /* Optional dilution of precision:  positional, horizontal, vertical.
    * 1 <= dop <= 50
--- a/route.cc
+++ b/route.cc
@@ -144,6 +144,18 @@ track_del_wpt(route_head* rte, Waypoint* wpt)
 }
 
 void
+route_swap_wpts(route_head* rte, WaypointList& other)
+{
+  global_route_list->swap_wpts(rte, other);
+}
+
+void
+track_swap_wpts(route_head* rte, WaypointList& other)
+{
+  global_track_list->swap_wpts(rte, other);
+}
+
+void
 route_disp(const route_head* /* rh */, std::nullptr_t /* wc */)
 {
 // wc == nullptr
@@ -503,4 +515,11 @@ void RouteList::swap(RouteList& other)
   const RouteList tmp_list = *this;
   *this = other;
   other = tmp_list;
+}
+
+void RouteList::swap_wpts(route_head* rte, WaypointList& other)
+{
+  this->waypt_ct -= rte->rte_waypt_ct();
+  this->waypt_ct += other.count();
+  rte->waypoint_list.swap(other);
 }
--- a/smplrout.cc
+++ b/smplrout.cc
@@ -96,18 +96,18 @@ double SimplifyRouteFilter::compute_track_error(const neighborhood& nb) const
   switch (metric) {
   case metric_t::crosstrack:
     track_error = radtomiles(linedist(
-                         wpt1->latitude, wpt1->longitude,
-                         wpt2->latitude, wpt2->longitude,
-                         wpt3->latitude, wpt3->longitude));
+                               wpt1->latitude, wpt1->longitude,
+                               wpt2->latitude, wpt2->longitude,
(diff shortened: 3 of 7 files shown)


defs.h
```diff
@@ -728,6 +729,8 @@ void route_add_wpt(route_head* rte, Waypoint* wpt, QStringView namepart = u"RPT"
 void track_add_wpt(route_head* rte, Waypoint* wpt, QStringView namepart = u"RPT", int number_digits = 3);
 void route_del_wpt(route_head* rte, Waypoint* wpt);
 void track_del_wpt(route_head* rte, Waypoint* wpt);
+void route_swap_wpts(route_head* rte, WaypointList& other);
+void track_swap_wpts(route_head* rte, WaypointList& other);
```
GPSBabelDeveloper, line 722: Nit: use 'trk' istead of 'rte' here, please.

route.cc
```diff
@@ -143,6 +143,18 @@ track_del_wpt(route_head* rte, Waypoint* wpt)
...
 }
 
+void
+route_swap_wpts(route_head* rte, WaypointList& other)
+{
+  global_route_list->swap_wpts(rte, other);
+}
+
+void
+track_swap_wpts(route_head* rte, WaypointList& other)
```
GPSBabelDeveloper, line 153: Likewise

smplrout.cc
```diff
@@ -71,136 +72,75 @@
 #if FILTERS_ENABLED
 #define MYNAME "simplify"
 
-void SimplifyRouteFilter::free_xte(struct xte* xte_rec)
+inline bool operator<(const SimplifyRouteFilter::trackerror& lhs, const SimplifyRouteFilter::trackerror& rhs)
 {
-  delete xte_rec->intermed;
+  return ((lhs.dist > rhs.dist) || ((lhs.dist == rhs.dist) && (lhs.wptpos < rhs.wptpos)));
```
GPSBabelDeveloper, line 77: Is this ">=" ?

smplrout.cc
```diff
@@ -211,123 +151,139 @@ void SimplifyRouteFilter::routesimple_head(const route_head* rte)
...
-  if (xte_count >= 1) {
-    totalerror = xte_recs[xte_count-1].distance;
-  }
+  assert(!errormap.isEmpty());
+  double totalerror = errormap.lastKey().dist;
 
   /* while we still have too many records... */
-  while ((xte_count) &&
-         (((limit_basis == limit_basis_t::count) && (count < xte_count)) ||
+  while ((!errormap.isEmpty()) &&
```
GPSBabelDeveloper, line 203: This is awesome. Love it!

smplrout.cc
```diff
@@ -211,123 +151,139 @@ void SimplifyRouteFilter::routesimple_head(const route_head* rte)
...
+         (((limit_basis == limit_basis_t::count) && (count < errormap.size())) ||
           ((limit_basis == limit_basis_t::error) && (totalerror < error)))) {
-    i = xte_count - 1;
+
     /* remove the record with the lowest XTE */
+    neighborhood goner = errormap.last();
```
GPSBabelDeveloper, line 208: This might be _too_ clever and open data-specific bugs (not worth it) but if we're far away from our goal (e.g. totalerror target is a mile and we're deleting inches with current error) might it be worth more aggressively whittling down errormap by reaping more goners (dark! :-) before recomputing?

Maybe that computation is so cheap that it doesn't matter. Hard for me to visualize the relative costs.

smplrout.cc
```diff
@@ -211,123 +151,139 @@ void SimplifyRouteFilter::routesimple_head(const route_head* rte)
...
+
+  } /* end of too many records loop */
+
+  /* delete marked points */
+  // For lineary complexity build a new list from the points we keep.
+  WaypointList oldlist;
+  (*route_swap_wpts_fnp)(const_cast<route_head*>(rte), oldlist);
+
+  // mimic trkseg handling from WaypointList::del_rte_waypt
+  bool inhereit_new_trkseg = false;
```
GPSBabelDeveloper: Perhaps "inherit"_new_trkseg ?

smplrout.cc
```diff
@@ -71,136 +72,75 @@
...
-                                        wpt2->latitude, wpt2->longitude,
-                                        wpt3->latitude, wpt3->longitude));
+      track_error = radtometers(linedist(
+                                  wpt1->latitude, wpt1->longitude,
+                                  wpt2->latitude, wpt2->longitude,
+                                  wpt3->latitude, wpt3->longitude));
     }
     // error relative to horizontal precision
-    xte_rec->distance /= (6 * wpt3->hdop);
+    track_error /= (6 * wpt3->hdop);
```
GPSBabelDeveloper, line 135: That URL is no longer live, so it doesn't really explain anything. 

As arbitrary as GPS DOP is, I wonder if "6" is just an equally arbitrary magnifier, but since it's constant to all of them, does it really scale anything at all?

GPSBabelDeveloper (approved):
I really like this. This is solid engineering and the result is WAY easier to read.

No reason to respond to comments individually. As long as you give them two seconds of thought and are sure, you don't have to defend not changing anything.

Thanx a lot!

smplrout.cc
tsteven4: Its worse than that.  SimplifyRouteFilter::init converts the error option value to miles.  For the relative option, SimplifyRouteFilter::compute_xte computes the incremental change in meters.  I noticed this but didn't change it.

smplrout.cc
tsteven4: We are defining operator<, not operator<=, so the tests don't include equality.
QMap returns items based on this operator from smallest to largest.  Since we want the smallest track error to be last I flipped the inequality in the dist comparison.  The polarity of the wptpos comparison is select to retain the same wpt as we did previously using stable_sort, flipping it results in a test failure.

smplrout.cc
tsteven4: thanks

defs.h
tsteven4: I was just following what we had before, e.g. 
void track_add_head(route_head* rte);
void track_del_head(route_head* rte);

smplrout.cc
tsteven4: I just did 8e6 points -> simplify -> 4e6 points in 48 seconds (for the filter alone), so I ~~don't~~ think we should just stick with the algorithm.  I think this would have taken days with 1.8.0.

smplrout.cc
GPSBabelDeveloper: > Its worse than that.

It always is... :-)

Thanx for the whole she-bang.

robertlipe:
Wow.

Shaving off a few zilliseconds but introducing a risk of data-dependent
bugs isn't worth it.

tsteven4:
Here is a new test, timing the reader, filter and writer separately, over 50e3 to 8e6 points.  It does appear there is still some non-linear behavior in the simplify filter.  On the other hand, it is faster than the gpx reader in all cases tested.

smptest

tsteven4:
@robertlipe should we drop support for route_priority in the simplify filter?  I haven't bothered with it so far.  I don't think it would have a big impact on performance, but the only use of it I can find today is the arcdist filter.

robertlipe:
Yes, please.

I've had ideas on speeding up the GPX reader before, but I've never put
fingers to keyboard to see if it's worth it. The idea I've kicked around is
something like:

if tag = gpx
if tag = wpt do_wpts()
if tag = trk do_trks()

do_wpts() {
   while (tag = wpt)
     if tag == name
          wpt->name = tag
          tag == next tag, testing for </wpt
     else if tag == description
         ...
     else if tag == date

...and then put them in GPX order so you can pretty much fall through on
most cases.

But our current "get a perfect hash and do a switch" shouldn't be a
slacker. XML is just plain wordy. The above is harder to get right for
malformed GPX.  Half the speed of the unicsv writer doesn't seem
unreasonable. I've always found the "else if string compare" chains
distasteful but most of the tags can bail after the second or third
character, so maybe I should get over that.

Is it worth the juice? 8M wpts in 60 to read, 40 to simplify, 35 to write
doesn't SEEM unreasonable. 8M points in GPX is surely several hundred MB of
source. We've allocated 8M * sizeof(Waypoint) and trampled the caches
pretty badly. Not much to avoid that without making a base class that's
just a 4D point (llat) and subclassing it, only kicking up to the larger
one if we see additional tags. Might be kinder on memory, but it's not very
elegant.

There's always something that's the chokepoint. (I once had a coworker that
described this effect as "There's always a #1 killer of Americans.")  This
seems pretty reasonable to me. Do you think it's worth poking at harder?

That simplify performance ROCKS!

robertlipe:
Oh, and unicsv is presumably writing fewer points than gpx read....

tsteven4:
> Yes, please.

Yes drop support for route priority, or yes add support for route priority to this PR?

robertlipe:
Nuke it from orbit.

I think that was another pet project of Ron's.

Of course MY pet projects are different... :-)

robertlipe (approved):
I forgot to actually click the 'approve' button. Looks great, of course!

PR #1150, opened 2023-07-28
time read, filter and writer operations

tsteven4:
when using -D > 0.

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -24,6 +24,7 @@
 #include <cstring>                    // for strcmp
 
 #include <QCoreApplication>           // for QCoreApplication
+#include <QElapsedTimer>              // for QElapsedTimer
 #include <QFile>                      // for QFile
 #include <QIODevice>                  // for QIODevice::ReadOnly
 #include <QLocale>                    // for QLocale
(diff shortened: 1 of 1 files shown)


main.cc
```diff
@@ -451,6 +468,10 @@ run(const char* prog_name)
       filter = FilterVecs::Instance().find_filter_vec(argument);
 
       if (filter) {
+        if (global_opts.debug_level > 0)  {
+          timer.start();
+qDebug() << timer.elapsed();
```
robertlipe: This might be a stray.

Also, we should have streaming operators in these. It's marginal, but 
Warning() << MYNAME << "filter" << fltname << "too" << QString::Number ...
should be an option if you ever want to go that way.

Good addition. I like it!

tsteven4: The stray has been rounded up, thanks.

We do have the streaming operators, but I find resolving the spacing and quoting issues tiresome. There are a bunch of things like 
Warning().nospace().noquote() << MYNAME << ": abc " << variable <<" is big";

GPSBabelDeveloper:
You're not wrong.

C++ 20 gives us std::format which I find pretty usable. std::print didn't
make it until '23. :-/

PR #1151, opened 2023-08-03
Introduce efficient member functions for deleting waypoints

tsteven4:
New functions with complexity O(n) for deleting pre-marked waypoints are:
  WaypointList::del_wpts
  RouteList::del_wpts
  del_wpts
  route_del_wpts
  track_del_wpts

Use those functions instead of the inefficient versions:
WaypointList::waypt_del
RouteList::del_wpt
waypt_del
route_del_wpt
track_del_wpt.
When these functions are used while looping over a waypoint list the overall complexity is O(n^2).  This is because these functions themselves amortize to O(n).

Optimization of the track filter will be done separately.

### DIFF
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -19,6 +19,8 @@
 
  */
 
+#include "arcdist.h"
+
 #include <cmath>                  // for round
 #include <cstdio>                 // for printf, sscanf
 #include <cstdlib>                // for strtod
@@ -28,7 +30,6 @@
 #include <QtGlobal>               // for foreach, qPrintable, qint64
 
 #include "defs.h"
-#include "arcdist.h"
 #include "grtcirc.h"              // for RAD, gcdist, linedistprj, radtomi
 #include "src/core/datetime.h"    // for DateTime
 #include "src/core/logging.h"     // for Fatal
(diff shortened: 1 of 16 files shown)


arcdist.cc
```diff
@@ -161,14 +162,14 @@ void ArcDistanceFilter::process()
...
 
+  int delete_flag; // &delete_flag != nullptr
   unsigned removed = 0;
   foreach (Waypoint* wp, *global_waypoint_list) {
-    auto* ed = (extra_data*) wp->extra_data;
-    wp->extra_data = nullptr;
-    if (ed) {
+    if (wp->extra_data) {
+      auto* ed = (extra_data*) wp->extra_data;
+      wp->extra_data = nullptr;
```
robertlipe, line 169: This seems like an aliasing test for an optimizer. Both (briefly) point to the same extradata before wpt->extra_data is changed to point to nowhere. .. and ed is deleted later in this block.

But I don't see a less excessively clever alterative offhand.

discard.cc
```diff
@@ -56,81 +54,63 @@ void DiscardFilter::fix_process_wpt(const Waypoint* wpt)
...
-{
-  head = const_cast<route_head*>(trk);
+  const_cast<Waypoint*>(wpt)->extra_data = del? &delete_flag : nullptr;
 }
 
 void DiscardFilter::process()
 {
-  WayptFunctor<DiscardFilter> fix_process_wpt_f(this, &DiscardFilter::fix_process_wpt);
-  RteHdFunctor<DiscardFilter> fix_process_head_f(this, &DiscardFilter::fix_process_head);
+  auto waypoint_cb_lambda = [this](const Waypoint* wpt) -> void {
```
robertlipe, line 97: Nice!

robertlipe (approved):
Very nice.

However, instead of all this messing about in extra data and related evaluators, should we just add a member to our Point structure, should we add a "self destruct" or "to be deleted" member/flag?

Still, this whole exercise has been a fascinating foray into an achilles heel of our fundamental data structures. Thank you for tackling it. Even where the code was already messy, you left it overall better.

tsteven4:
> However, instead of all this messing about in extra data and related evaluators, should we just add a member to our Point structure, should we add a "self destruct" or "to be deleted" member/flag?

I added a dedicated flag in wpt_flags which cleans things up a bit.  There was a uncommitted version that used templates so I could pass lambdas as evaluators, that was even messier!

I think if we started over we would use std:: containers.  They are better documented, especially concerning complexity promises.  But at this point I think we are fairly well optimized for Qt containers, especially considering all the container copies we do (.e.g. foreach) and the Qt copy on write architecture.  Also, we don't have a complexity test suite similar to the tests I have done recently on simplify, ...

Issue #1152, opened 2023-08-05
trackfilter pack handling of segments


tsteven4:
The track filter pack option preserves any segment markers in the first track, and deletes any segment markers in subsequent tracks.  Thus all tracks and segments after the first track are packed into the last segment in the first track.

It would seem better to either:
1) preserve all segments regardless of which track they originated in
2) pack all points into a single segment

gpsbabel -i gpx -f trkseg.gpx -x track,pack -o gpx -F trkseg_pack.gpx
[trkseg.gpx.txt](https://github.com/GPSBabel/gpsbabel/files/12267678/trkseg.gpx.txt)

```
$ sdiff trkseg.gpx trkseg_pack.gpx 
<?xml version="1.0" encoding="UTF-8"?>				<?xml version="1.0" encoding="UTF-8"?>
<gpx version="1.0" creator="GPSBabel - https://www.gpsbabel.o	<gpx version="1.0" creator="GPSBabel - https://www.gpsbabel.o
  <time>2023-08-05T16:30:10.721Z</time>			      |	  <time>2023-08-05T16:31:57.829Z</time>
  <bounds minlat="30.046400000" minlon="-91.610567000" maxlat	  <bounds minlat="30.046400000" minlon="-91.610567000" maxlat
  <trk>								  <trk>
    <name>LOG-20020525</name>					    <name>LOG-20020525</name>
    <trkseg>							    <trkseg>
      <trkpt lat="30.062183000" lon="-91.610350000">		      <trkpt lat="30.062183000" lon="-91.610350000">
        <ele>1.000</ele>					        <ele>1.000</ele>
        <time>2002-05-25T17:06:21Z</time>			        <time>2002-05-25T17:06:21Z</time>
      </trkpt>							      </trkpt>
      <trkpt lat="30.062783000" lon="-91.610567000">		      <trkpt lat="30.062783000" lon="-91.610567000">
        <ele>0.000</ele>					        <ele>0.000</ele>
        <time>2002-05-25T17:09:55Z</time>			        <time>2002-05-25T17:09:55Z</time>
      </trkpt>							      </trkpt>
    </trkseg>							    </trkseg>
    <trkseg>							    <trkseg>
      <trkpt lat="30.052483000" lon="-91.595967000">		      <trkpt lat="30.052483000" lon="-91.595967000">
        <ele>0.000</ele>					        <ele>0.000</ele>
[...]

PR #1155, opened 2023-08-11
use efficient methods to delete trkpts in track filter.

tsteven4:
This PR uses track_del_marked_wpts instead of track_del_wpt.  track_del_marked_wpts scales with the number of points in the list, while track_del_wpt scales with the number of points in the list times the number of points deleted.

There are some intentional changes of behavior:

1. previously in option split all existing track segment markers after the first split were removed, but those before the split were conserved.  Now all pre-existing segment markers are removed, leaving only one segment per track.
2. previously in option pack all existing track segment markers after the first track were removed, but those in the first track were conserved.  Now all pre-existing segment markers are removed, leaving only one segment in the packed track.  This is issue #1152.

### DIFF
--- a/trackfilter.cc
+++ b/trackfilter.cc
@@ -22,16 +22,14 @@
 
 static constexpr bool TRACKF_DBG = false;
 
-#include <algorithm>                       // for for_each
+#include <algorithm>                       // for_each, sort, stable_sort
 #include <cassert>                         // for assert
 #include <cmath>                           // for nan
 #include <cstdio>                          // for printf
 #include <cstdlib>                         // for abs
 #include <cstring>                         // for strlen, strchr, strcmp
-#include <ctime>                           // for gmtime, strftime
+#include <ctime>                           // for gmtime, strftime, time_t, tm
 #include <iterator>                        // for next
-
-#include <algorithm>                       // for sort, stable_sort
 
 #include <QByteArray>                      // for QByteArray
 #include <QChar>                           // for QChar
@@ -272,7 +270,7 @@ void TrackFilter::trackfilter_pack_init_rte_name(route_head* track, const gpsbab
     if (track->rte_waypt_empty()) {
       dt = default_time;
     } else {
-      auto* wpt = track->waypoint_list.front();
+      const auto* wpt = track->waypoint_list.front();
       dt = wpt->GetCreationTime();
     }
     time_t t = dt.toTime_t();
@@ -388,7 +386,7 @@ void TrackFilter::trackfilter_merge()
 
     std::stable_sort(buff.begin(), buff.end(), trackfilter_merge_sort_cb);
 
-    Waypoint* prev = nullptr;
+    const Waypoint* prev = nullptr;
 
     for (auto* wpt : buff) {
       if ((prev == nullptr) || (prev->GetCreationTime() != wpt->GetCreationTime())) {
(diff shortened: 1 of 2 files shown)


trackfilter.cc
```diff
@@ -22,16 +22,15 @@
...
 static constexpr bool TRACKF_DBG = false;
 
+#include <algorithm>                       // for_each, sort, stable_sort
 #include <cassert>                         // for assert
 #include <cmath>                           // for nan
 #include <cstdio>                          // for printf
 #include <cstdlib>                         // for abs
 #include <cstring>                         // for strlen, strchr, strcmp
-#include <ctime>                           // for gmtime, strftime
+#include <ctime>                           // for gmtime, strftime, time_t, tm
```
robertlipe, line 31: Is this still needed? We've been trying to drop ctime.

trackfilter.cc
```diff
@@ -329,11 +324,20 @@ void TrackFilter::trackfilter_pack()
 
     route_head* master = track_list.first();
 
+    if (!master->waypoint_list.empty()) {
+      std::for_each(std::next(master->waypoint_list.cbegin()), master->waypoint_list.cend(),
```
robertlipe, line 328: I'm excited that some day std::foreach can parallelize things automatically for us. Some day....

trackfilter.cc
```diff
@@ -911,46 +900,45 @@ bool TrackFilter::trackfilter_points_are_same(const Waypoint* wpta, const Waypoi
...
         }
       }
-      if (cur_dist > .001 && cur_dist > 1.2* avg_dist) {
-        avg_dist = cur_dist = 0;
+
+      if (avg_dist == 0) {
+        avg_dist = cur_dist;
+      }
+
+      if (cur_dist > 6378.14 && cur_dist > 1.2 * avg_dist) {
```
robertlipe, line 931: Took me a moment to recognize such a specific magic number. Isn't that a global (hehe) constant for us?

robertlipe (approved):
Algorithmically very clever. As waysl, thank you!

trackfilter.cc
tsteven4: strftime exposed to user in title option :(

trackfilter.cc
tsteven4: These magic numbers were translations of previous values with units of radians or degrees.  The old comment
`Latitude spacing is about 27 feet per .00001 degree`
was incorrect, it is 1.11m or 3.6ft.

It seems to be arbitrary, but I though 6378 meters was easier to understand that 0.001 radians.  I have a hard time imagining this specific value is generally useful.

Would it be more useful to let the user pass this limit?  I' not aware of any interest, complaints or use related to this option.

kDistanceLimit seems arbitrary as well, but again 1.11 meters was easier to understand than 0.00001 degrees of latitude or longitude at the equator.  There was another magic number, ktoo_close = 0.000005 radians = 31.89 meters that we used to decide if we needed to check points for being the same, this magic has been replaced by kDistanceLimit which is used both to see if we should check, and to decide if the points are the same.  1.11m seems a bit tight to decide we are sitting still.

Perhaps it makes sense, and is generally applicable, to set kDistanceLimit to something like 4.9m https://www.gps.gov/systems/gps/performance/accuracy/

We had a long email 1/10/2019 about this.  The test cases referred to are not available.  I created one so we have some coverage, but it certainly isn't the real data that inspired this option.

GPSBabelDeveloper:
Ah, in the transformed forms, I recognize these magic numbers. They were,
indeed, pulled out of a hat.

No, let's not expose those knobs as Yet More Options.

PR #1167, opened 2023-08-29
convert googletakeout format to dynamic.

tsteven4:
Dynamic formats are constructed for each use, and destroyed afterwards.  This allows:
1) format class resources to be initialized by the format constructor, e.g. with default member initializers or member initializer lists.
2) eliminate the burden of returning resources held by a format class data member.  The destructor of the class data member will be called when the class is destroyed.
3) In some cases this can eliminate the need for a member variable that stores something set by rd(wr)_init and is used by read(write).  This alone can simplify the return of resources as any variables of automatic storage duration will be destroyed at the end of the enclosing block.
4) unintended communication between successive invocations of the format to be prevented.

In this case, the data member inputStream can become variable with automatic storage duration of the member function read().  Note that previously the inputStream was not cleaned up until program exit.

### DIFF
--- a/googletakeout.cc
+++ b/googletakeout.cc
@@ -217,16 +217,13 @@ GoogleTakeoutFormat::title_case(QString& title)
 }
 
 void
-GoogleTakeoutFormat::rd_init(const QString& fname) {
+GoogleTakeoutFormat::read()
+{
   if (global_opts.debug_level >= 4) {
     Debug(4) << "rd_init(" << fname << ")";
   }
-  inputStream = GoogleTakeoutInputStream(fname);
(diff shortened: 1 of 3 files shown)


postmaxin (approved):

tsteven4:
@postmaxin our evolution from our c beginnings to today has taken us from formats implemented as static global functions and data (which are now wrapped in instances of the LegacyFormat class), to static instances of the Format class held by vecs for the duration of execution, to dynamic instances of the Format class that are created and destroyed on each use.  We still have examples of each, but the desire is to move towards dynamic formats.

PR #1168, opened 2023-08-31
fix mkshort unique

tsteven4:
1. we were off by one when deciding if we could just concatenate
the generated suffix to the name, failing to use the last available
character.
2. when attempting to copy the generated suffix to a presumably
shortened name, and the generated suffix was longer than the name,
the target of the copy was outside the name buffer.  In the common
case that the rank of size_t was greater than the rank of int, the
target was well beyond the end of name, not before as one would expect
with signed arithmetic.

With the new algorithm when the target length is insufficient to
fit both the name and the suffix we will only truncate the name as
required to fit the truncated name and the complete suffix in the
target length(as opposed to the original length of the name).
We fatal if this is not possible.

Add a test case to exercise the make unique code.
@robertlipe please verify that the generated references mkshort6.csv, mkshort5.csv, mkshort4.csv have generated names as desired.


GPSBabelDeveloper:
LGTM

Thanx!

Instead of this suffering (too late) why not just make the the
lenght unsigned? It's never sensible for it to be negative ... or even
really smaller than, oh, "six" or so in practice.

I do agree with your sense of increased fatalism in the code. There are
some cases, mostly pathological, that just plain won't fit. I also later
developed a bad feeling that the code was very English-oriented, but the
feelign wasn't bad enough to include tables of words of the digits in the
top 45 languages.

We should also assert that '.' is never in the blacklist.


I do also, unsurprisignly, agree with your (heroic) efforts to preserve
fidelity for those cases. Those seem spot-on.

For a different spin on all this, peek at shortname.cc , which was never
finished but was my hobby while in flight during the work commute for a
trip or two, before we were even C++, I think. About half the body was this
terrible "test driven design" approach of a table of 2003 (ish) geocache
names and how they'd be reduced to receivers of that era.  Then again, as
our own mkshort.cc has become more QString-centric, it's also become less
terrible to look at on its own.

Trivia: in all my geocaching escapades, there were only two cases where I
know there were geocaches I couldn't find because they were lost by
GPSBabel on the way to the receiver. One involved very long geocache names
with a sequence number that wasn't at the end - we collapsed those all into
one and most of the series just wasn't on my hardware. The second I only
caught because I had two _different_ Garmins in my hunting at that time. An
often missed (and the kind of thing you'd approximately never find via
reverse engineering) nuance of the Garmin USB protocol is that on certain
models, if the sent packet is exactly 256 characters (255?) you must insert
a following packet that's basically empty to "push" that one on through,
otherwise it will be lost. So I'd used a shell script that loaded one GPX
file (maybe a merge of a couple...) to my StreetPilot and my GPSMap 60Cx
(60CSx?) and one receiver had a specific point and the other didn't. Took a
serious technology dive to find the truth on that one.

PR #1170, opened 2023-09-03
Update gpsbabel_man.xml

jidanni:
Without a dash this will resemble the "tar" command!

Oops... it seems I made a side patch... but I don't know how to delete it.

### DIFF
--- a/xmldoc/gpsbabel_man.xml
+++ b/xmldoc/gpsbabel_man.xml
@@ -24,7 +24,7 @@
     <cmdsynopsis>
       <command>gpsbabel</command>
       <arg choice="opt">
-        <option>iofFxsh</option>
+        <option>-iofFxsh</option>
       </arg>
       <arg choice="opt" rep="repeat">file</arg>
     </cmdsynopsis>


robertlipe (commented):
https://tdg.docbook.org/tdg/4.5/option.html#:~:text=Processing%20expectations&text=DocBook%20does%20not%20specify%20whether,Appendix%20F%2C%20Interchanging%20DocBook%20Documents.

Current approach is valid and, IMO, preferable as '-' is not an option.

Do you have a more authoritative source than DocBook's own "eh, whatever" answer?

jidanni:
All I know is the only man pages that don't use a dash are:

```
SYNOPSIS
       gpsbabel [iofFxsh] [file...]

SYNOPSIS
   Traditional usage
       tar {A|c|d|r|t|u|x}[GnSkUWOmpsMBiajJzZhPlRvwo] [ARG...]
```

Also there is still a dash here:


```
Usage:
    gpsbabel [options] -i INTYPE -f INFILE [filter] -o OUTTYPE -F OUTFILE
    gpsbabel [options] -i INTYPE -o OUTTYPE INFILE [filter] OUTFILE
```
So I would still keep one on the man page.

robertlipe:
That's not very authoritative, but I'll take it.

Before you go on another nitpicking bender, I'll tell you my next step in the man page is to delete it. The other doc is SO much better that I'm just not spending time on the man page.

PR #1171, opened 2023-09-03
convert mkshort from char* to QByteArray

tsteven4:
also, store the conflict counter directly in the QHash simplifying memory management.

This changes behavior.  The preference for trailing digits didn't really work.  The updated algorithm prefers trailing digits over anything that precedes the digits.  If there isn't room for all the trailing digits then the least significant digits are used.  This is demonstrated in the enhanced mkshort test (although testing both branches of is_utf8 requires fiddling with the code).

Future PR:
1) Can we always process utf8?  The only non-utf8 caller is garmin.  For utf8 we interpret the short length as the number of graphemes, for non-utf8 we interpret it as the number of chars.  In utf8 a grapheme can be encoded by multiple codepoints.  In the codecs likely used by garmin (windows-1252, US-ASCII) I believe a grapheme was always 1 char.  The fundamental question is are we trying to limit the number of bytes required to store the string, or the number of display positions it takes to display the string.  We have been transition to the later definition, but for garmin they are likely identical.
2) mkshort becomes a class.


mkshort.cc
```diff
@@ -387,51 +331,37 @@ mkshort(short_handle h, const QByteArray& istring, bool is_utf8)
...
+   * Whack leading "[Tt]he "
    */
-  if ((strlen(ostring) > hdl->target_len + 4) &&
-      (strncmp(ostring, "The ", 4) == 0 ||
-       strncmp(ostring, "the ", 4) == 0)) {
-    char* nstring = xstrdup(ostring + 4);
-    xfree(ostring);
-    ostring = nstring;
+  if ((ostring.size() > (hdl->target_len + 4)) &&
+      (ostring.startsWith("The ") || ostring.startsWith("the "))) {
```
robertlipe, line 344: Maybe ostrings.startWith("the", Qt::CaseInsensitive) ?

That first test may not even be needed in QString-land. Won't it just terminate the search early if the haystack contains fewer characters than the needle?

To bring a gun to a knife fight:
ostring.remove("^the", Qt::CaseInsensitive);
handles all this for us, I think.

If that works out, the above "Starts with "GC"" could do the same.

mkshort.cc
```diff
@@ -20,34 +20,28 @@
  */
 
 #include <cassert>     // for assert
-#include <cctype>      // for isspace, toupper, isdigit
-#include <climits>     // for INT_MAX
-#include <cstring>     // for strlen, memmove, strchr, strcpy, strncmp, strcat, strncpy
+#include <cctype>      // for isspace, isdigit
```
robertlipe, line 23: are those still being called on C strigns?

tsteven4:
If we can always process graphemes as speculated above in #2, then we can almost trivially change mkshort to use QStrings instead of QByteArrays, eliminated some back and forth.

mkshort.cc
```diff
@@ -196,51 +157,46 @@ mkshort_del_handle(short_handle* h)
...
-        (0 == case_ignore_strcmp(r->orig, &s[origslen - rl])) &&
-        (s[origslen - rl - 1] == ' ')) {
-      strcpy(&s[origslen - rl], r->replacement);
-      return ;
+    if ((s.size() > r.orig.size()) &&
+        (0 == r.orig.compare(s.mid(s.size() - r.orig.size()), Qt::CaseInsensitive)) &&
+        (s.at(s.size() - r.orig.size() - 1) == ' ')) {
+      s.truncate(s.size() - r.orig.size());
+      s.append(r.replacement);
+      return;
```
robertlipe, line 200: Is this snippet just hard to read in the review context, or does this fn really not live up to the plurality implied by the name? It looks like it replaces one and returns instead of continuing to work through the entire list. (I see the original does the same)
My read says that "Oh, You Eight One Two" would hit because they're multiple replacements but  "Dial Nine One One" might not get both "ones" because of the early return here.
Is a fix as easy as removing the return here?

mkshort.cc
```diff
@@ -387,51 +331,37 @@ mkshort(short_handle h, const QByteArray& istring, bool is_utf8)
...
-    l = strlen(tstring);
-    cp = ostring;
-    for (i=0; i<l; i++) {
-      if (!isspace(tstring[i])) {
-        *cp++ = tstring[i];
+    QByteArray tstring;
+    ostring.swap(tstring);
+    for (const auto ch : qAsConst(tstring)) {
+      if (!isspace(ch)) {
+        ostring.append(ch);
```
robertlipe, line 359: ostring.remove("[ \t]") would do this, I think. If we're concerned about formfeeds or any other form of whitespace, just add them inside the [].

mkshort.cc
```diff
@@ -443,54 +373,34 @@ mkshort(short_handle h, const QByteArray& istring, bool is_utf8)
...
+    QByteArray tstring;
+    ostring.swap(tstring);
+    for (const auto ch : qAsConst(tstring)) {
+      if (hdl->badchars.contains(ch)) {
+        continue;
+      }
+      if (!hdl->goodchars.isEmpty() && (!hdl->goodchars.contains(ch))) {
+        continue;
+      }
+      ostring.append(ch);
```
robertlipe, line 387: QString re = "[" + badchars + "]" 
ostring.remove(re);

HOWEVER, this all comes down if any of our bad characters are common in an RE list, like a [....

robertlipe (commented):
Re: your comments.
2) I started that with shortname.cc, but just never got back to it.
1) I'll admit I don't know how to handle multi-byte well in this. Since it's (mostly) optional and people haven't noticed that we can chop the 0xc2 from the 0x80 (or whatever) maybe it's not a problem in practice? {shrug} Nobody's complained...
0) I can't visualize the question.  The code was strongly opinionated that "Dalmatian #47" became "Dalm47", keeping the last number. Can you give examples of names that would trigger the difference and list before and after?

mkshort.cc
tsteven4: they are being called on chars, but that will go away if we convert mkshort from QByteArray to QStrings.

mkshort.cc
tsteven4: it only does one replacement maximum, like the previous code.

tsteven4:
@robertlipe My big question is can we convert the mkshort interface to QString only.

> In Unicode, a code point is an atomic part of text. However, a grapheme cluster corresponds most closely to a symbol displayed on screen or paper. It is defined as “a horizontally segmentable unit of text”.

Counting graphemes should give us results that fit into a number positions on a display = the short length.  But, the shortened text may occupy more than the short length number of bytes.  I would like to count graphemes only, and not count bytes.  The only caller who is counting bytes today is garmin.  Garmin uses US_ASCII and windows-1252 (and a recently added user defined codec).  For US_ASCII and windows-1252 the number of bytes = number of graphemes, so I think that case can also be covered by counting graphemes.

mkshort.cc
tsteven4: QByteArray doesn't support QRegExp/QRegularExpression.

mkshort.cc
tsteven4: QByteArray doesn't support QRegExp/QRegularExpression.

tsteven4:
the garmin change would be:
```
-    QByteArray ident = mkshort(mkshort_handle,
-                               global_opts.synthesize_shortnames ?
-                               str_from_unicode(src) :
-                               str_from_unicode(wpt->shortname),
-                               false);
+    QByteArray ident = str_from_unicode(
+                         mkshort(mkshort_handle,
+                                 global_opts.synthesize_shortnames ?
+                                 src : wpt->shortname));
```
I assert with US-ASCII or windows-1252 ident would have the same size.

mkshort.cc
tsteven4: We have a comment:
>  Right now, replacements are made only at the end of the string.

If we want to do replacement of all the matching words, anywhere they appear, that is easy enough.

mkshort.cc
tsteven4: QByteArray::remove doesn't support case sensitivity.

tsteven4:
> 2\. Can you give examples of names that would trigger the difference and list before and after?

There are examples in reference/mkshort4.log.  "abcd0123" with various short lengths becomes:
```
"0.00000","0.00000","abc0123"
"0.00000","0.00000","ab0123"
"0.00000","0.00000","a0123"
"0.00000","0.00000","0123"
"0.00000","0.00000","123"
"0.00000","0.00000","23"
"0.00000","0.00000","3"
```
For the base, and non utf8, the results are as follows.  note the last 2 violate the requested length.
```
"0.00000","0.00000","abc0123"
"0.00000","0.00000","ab0123"
"0.00000","0.00000","a0123"
"0.00000","0.00000","0123"
"0.00000","0.00000","0123"
"0.00000","0.00000","0123"
"0.00000","0.00000","0123"
```

For the base, with utf8, the results are as follows.  Note the comment /* ToDo: Keep trailing numeric data as described above! */
```
"0.00000","0.00000","abcd012"
"0.00000","0.00000","abcd01"
"0.00000","0.00000","abcd0"
"0.00000","0.00000","abcd"
"0.00000","0.00000","abc"
"0.00000","0.00000","ab"
"0.00000","0.00000","a"
```

tsteven4:
@robertlipe Thanks for your comments.  I am going to merge this before they are resolved so I can move forward.  We can tackle them later.

PR #1173, opened 2023-09-04
mkshort becomes a class

tsteven4:


robertlipe:
Nifty. Did Codelion or whatever, make this as easy as they claim?

Thanks

tsteven4:
> Nifty. Did Codelion or whatever, make this as easy as they claim?

The last commit, renaming, was thanks to JetBrains ReSharper.  Before that it was by hand.

So now we are down to the QByteArray vs. QString question.  Many of your review comments relate to methods available in QString but not QByteArray.  In parallel we have the related issue of !is_utf8 handling which counts characters vs. is_utf8 handling which count graphemes (and has for some time).

robertlipe:
CLion, at least, claims support for "extract into" and "refactor into" that
might have helped. Have looked at either ReSharper or Rider. (...or figured
out why they need three...)

My review comments were just based on me knowing QString better over time
and - obviously - forgetting that QByteString much even exists. they
shouldn't push the design decision.

If you're happy with the current QBA, LGTM. The interfaces play out,
largely thanks to that QBS->QString constructor. I can't tell that most
code will care either way,

What does the environment do to save us from ourselves either way?  We're
still indexing like characters = graphemes and using memcpy and strncpy and
such to move bytes, but moving that to a QString "just" saves us for those
characters fit into a single 16-bit value (i.e. most of the languages in,
but almost only, the Western World), right? Isn't the most fundamental
problem, that we need to change our whimsical character-killin' ways - one
that we have to solve either way?

PR #1175, opened 2023-09-08
Apply new resharper refactoring "Redundant dereferencing and taking address"

tsteven4:


robertlipe:
So much cleaner. Thanx

PR #1176, opened 2023-09-10
fix noreturn booboos.

tsteven4:
both these showed warnings on jammy


GPSBabelDeveloper:
Interesting. Probably a hint that the flow analysis in the optimizer is
falling down.

PR #1054, opened 2023-03-27
Extract engine noise from IGC, add to KML multipath

PacketFiend:
Scans the IGC "I" record to determine which extensions are present in the log, and extracts that data into `struct igc_fsdata : public FormatSpecificData`.

Currently only engine noise levels (ENL) are extracted, but the skeleton is there for any number of extensions.

If engine noise level data is present when generating a KML output file, we also now output that as part of a multipath ExtendedData section. This way, it can be overlaid onto the Google Earth elevation profile, alongside either altitude or speed.

I'm not sure exactly which string libraries to use (there's an ugly sscanf that should probably be changed) or whether or not I'm following the general culture of the project, so feel free to change this.

Running `const auto* fs_igc = reinterpret_cast<igc_fsdata*>(wpt->fs.FsChainFind(kFsIGC));` on every datum of a log file is probably not very efficient. I have some IGC files with over 25,000 entries. Is there a better way to do that?

Thank you Robert for your detailed explanation, this would have not been possible otherwise. I haven't set up tests yet, so I'm hoping all the standard tests will pass.

(Also adds .vscode/, bld/, and build/ to .gitignore. That's a personal preference but I think it's a good idea.)

PR #1177, opened 2023-09-12
fix SchemaArrayField/SchemaArrayData mismatch

tsteven4:
related to IGC additions from #1060 and #1054.

### DIFF
--- a/defs.h
+++ b/defs.h
@@ -248,30 +248,6 @@ public:
   unsigned int marked_for_deletion:1;		/* True if schedulded for deletion. */
 };
 
-// These are dicey as they're collected on read. Subsequent filters may change
-// things, though it's unlikely to matter in practical terms.  Don't use these
-// if a false positive would be deleterious.
-#
-class global_trait
-{
-public:
-  global_trait() :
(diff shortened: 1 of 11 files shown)


tsteven4:
@PacketFiend this does not differ from tsteven4:kmlaccumulate you looked at yesterday, but the history reflects the merge of #1060.  It is a draft as I want to clean things up a bit.

tsteven4:
@robertlipe I intend to remove global_trait traits.  The only usage was kml, and the kml usage for multitrack is specific only to track points or, for geoache, only to waypoints.

GPSBabelDeveloper:
I've lost the beat (traveling/vacation) but the reason I added the 'traits'
stuff was so that we could amortize the cost of figuring out what kind of
stuff we have that might require a preamble during add instead of making a
pass over the entire waypoint/track list. That was we could decide whether
to write the geocache or track template in the KML header
without actually scanning the whole list to find the presence/absence of
these things.

If you've found a way around this that's performant enough to meet your
high standards, I won't weep.

RJL

PR #1199, opened 2023-10-27
move Qt floor from 5.12 to 5.15

tsteven4:
### This will result in gpsbabel not being buildable on Ubuntu focal with the Ubuntu supplied Qt5, which is 5.12.8.

The next to newest LTS release, focal, reaches end of standard support April 2025, and end of life April 2030.
We are find on the newest LTS release jammy (Qt 5.15.3 and 6.2.4 are both available).
The next LTS release, noble, is scheduled for April 2025.

1.9.0 should be buildable on jammy.  Debian has built 1.9.0 with Qt6, but it hasn't filtered down to Ubuntu yet.

If we need to fix 1.9.0 we can create a branch from gpsbabel_1_9_0 and release 1.9.1.  By the time we release anything other than a patch version (e.g. 1.9.1) I think focal will have reached end of standard support, and there will be a new Ubuntu LTS release.

### Qt official support levels.  I don't see the Qt support levels as an issue at 5.15.  We lose macOS 10.12, Ubuntu 16.04 compared to 5.12

macOS with 5.12: macOS 10.12, 10.13, 10.14
macOS with 5.15: **macOS 10.13**, macOS 10.14, macOS 10.15, macOS 11, macOS 12
macOS with 6.2: macOS 10.14, 10.15, 11, 12
macOS with 6.5: macOS 11, 12, 13

windows with 5.12: 7, 8.1, 10
windows with 5.15: **7**, 8.1, 10, 11
windows with 6.2: 10,11
windows with 6.5: 10,11

These are the support levels for Qt which differs from what gets packaged by Ubuntu.
Ubuntu with 5.12: 16.04, 18.04
Ubuntu with 5.15: **18.04**, 20.04(focal)
Ubuntu with 6.2: 20.04(focal)
Ubuntu with 6.5: 22.04(jammy)


robertlipe:
Ok.

I generally. Think that people run old OSes because they don't WANT new app versions. In general, People running an LTS will be happiest with the build from their OSV, which never gets updated, of course. People with desktop systems that are actively maintained aren't running these time capsules.

One of the reasons I welcome newer versions. Is to welcome newer language features. {fmt}.is an example of something I'd love to be able. To use without shipping our own.

5.15 is worth some crawl. From us, but I don't want to stay there forever just for those windows 7 people.

tsteven4:
As far as Ubuntu is concerned, we could support the latest two LTS releases with Ubuntu supplied Qt as soon as 24.04 comes out (scheduled for April 2024).  If we are happy with windows 10 and macos 10.14 floors I think we should schedule the move to a Qt 6.2 floor for April 2024 (or when Ubuntu 24.04 noble ships).  We do need to watch out for TLS and SSL compatibility issues when using the Qt supplied builds like shown below.  This is another reason to use the Qt builds supplied by the packagers on Linux.
```
Incompatible version of OpenSSL (built with OpenSSL 1.x, runtime version is >= 3.x)
```
```qt.network.ssl: The backend "cert-only" does not support QSslKey
qt.network.ssl: Active TLS backend does not support key creation
qt.network.ssl: The backend "cert-only" does not support QSslKey
qt.network.ssl: Active TLS backend does not support key creation
qt.network.ssl: The backend "cert-only" does not support QSslKey
qt.network.ssl: Active TLS backend does not support key creation
qt.network.ssl: The backend "cert-only" does not support QSslKey
qt.network.ssl: Active TLS backend does not support key creation
qt.network.ssl: The backend "cert-only" does not support QSslKey
qt.network.ssl: Active TLS backend does not support key creation
qt.network.ssl: The backend "cert-only" does not support QSslKey
qt.network.ssl: Active TLS backend does not support key creation
qt.network.ssl: The backend "cert-only" does not support QSslSocket
qt.network.ssl: The backend named "cert-only" does not support TLS
qt.network.ssl: QSslSocket::connectToHostEncrypted: TLS initialization failed
```

PR #1202, opened 2023-10-30
convert garmin_fs functions from char* -> QString

tsteven4:
also

- move gpx specific garmin_fs_xml_convert to GpxFormat.
- covert GpxFormat::tag_type to a class enum.


garmin_fs.cc
```diff
@@ -76,130 +73,40 @@ garmin_fs_t::~garmin_fs_t()
   }
 }
 
-/* GPX - out */
-
-void
-garmin_fs_xml_convert(const int base_tag, int tag, const QString& qstr, Waypoint* waypt)
+bool
+garmin_fs_convert_category(const QString& category_name, uint16_t* category)
```
robertlipe, line 76: Should we phase this stye of calling convention out in favor of returning an optional category ans then testing has_value() in the caller?

Not a blocker for this - I don't want to punish rearrangement...

garmin_fs.cc
```diff
@@ -76,130 +73,40 @@ garmin_fs_t::~garmin_fs_t()
...
-      // There's nothing a user can really do about this (well, they could
-      // create a gpsbabel.ini that mapped them to garmin category numbers
-      // but that feature is so obscure and used in so few outputs that
-      // there's no reason to alarm the user.  Just silently disregard
-      // category names that don't map cleanly.
-      // warning(MYNAME ": Unable to convert category \"%s\"!\n", cdatastr);
+  // Is the name  "Category" followed by a number? Use that number.
+  if (category_name.startsWith(u"Category ", Qt::CaseInsensitive)) {
+    bool ok;
+    int i = category_name.mid(9).toInt(&ok);
```
robertlipe, line 81: ...toInt() would probably do that if it were written today.
Also, this could be toUint() and save one test below.

robertlipe (approved):
Nice modernization.

garmin_fs.cc
tsteven4: yes, https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-out-multi
or in this case as you point out an optional will do.

tsteven4: thanks, but I'm moving on to bigger fish.

PR #1203, opened 2023-10-31
refactor garmin_fs class

tsteven4:
- use a Qt container for ilinks.
- let the container manage memory, taking advantage of copy on write.
- replace garmin_fs_alloc with new.
- make garmin_fs_t* constant where possible, eliminating the possibility of the ilinks list detaching.
- return an optional from garmin_fs_t::convert_category instead of returning status and passing a pointer to the output value.
- retire garmin_fs_convert_category.  This was mostly duplicated between  garmin_fs_t::convert_category and the caller.
- create garmin_fs_t::print_categories from the original in garmin_txt, and use it in both garmin_txt and the gpx writer.
- add opt_utc to garmin_txt reader (it was already in the garmin_txt writer)
- correct bugs in garmin_txt function strftime_to_timespec.  %H was mistranslated, %p was missing, if the format code wasn't found a character from outside the buffer could be used.  add warning when a conversion isn't found.
- add a test proving categories can go from gpx to garmin_txt and back.  additionally the test exercises the ini file category support.
- improve debug message in convert_from_human_[time|date]_format.

### DIFF
--- a/garmin.cc
+++ b/garmin.cc
@@ -1275,7 +1275,7 @@ garmin_fs_garmin_after_read(const GPS_PWay way, Waypoint* wpt, const int protoid
 static void
 garmin_fs_garmin_before_write(const Waypoint* wpt, GPS_PWay way, const int protoid)
 {
-  garmin_fs_t* gmsd = garmin_fs_t::find(wpt);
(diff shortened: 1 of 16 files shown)


robertlipe:
Consistently good stuff.. %his is so much less fragile and easier to read.. Thanx.

Issue #1113, opened 2023-05-15
strftime/strptime specifiers -> QDate/QTime/QDateTime specifiers?


tsteven4:
We have lamented publishing the strftime/strptime specifiers.  We could replace these by [QDate](https://doc.qt.io/qt-6/qdate.html#fromString)/[QTime](https://doc.qt.io/qt-6/qtime.html#fromString)/[QDateTime](https://doc.qt.io/qt-6/qdatetime.html#fromString) specifiers.  I note the toString versions are independently documented with some variances.

Would we just be jumping from one lament to another?

Potentially it would allow us to get rid of strptime/strftime/<ctime>.  I think conversion of the remaining xcsv styles would be easy.  I think google wouldn't even give a seconds thought to interface stability while sun would agonize over it until bankruptcy.

robertlipe:
Maybe. I have my eye on a time when QDateTime becomes std::chrono, but that
might happen right before Sun's re-emergence and subsequent re-bankruptcy.

In theory, there are style files in the world that aren't in our tree.
Don't know how common that really is.

Somewhere fairly recently, I wrote a shim that took a strtime (strptime?)
string and made it into a QT timespec. Maybe it was the other way around.
There were lots of things it didn't handle, but it was another Paretto case
where doing 20% of the specifiers got us (more than) the 80% we needed.
It's a pretty mechanical transformation for the simple cases.

NOte that we actually have terrible test coverage of the *.style file
formats.

In Google-land, those *.style formats would be a (versioned) protobuf.
You're not wrong about your analogy. Google tooled for that kind of thing
very early.

robertlipe:
Aha. garmin_txt.cc:strftime_to_timespec

It let me delete something in garmin_txt what was a pain point, IIRC. Odd,
since the guts of that format are not my favorite code and I know it was on
the chopping block.

PR #1205, opened 2023-11-01
simplify formatting of date/time in garmin_txt

tsteven4:
This is a user visible change.

This is a test of a solution to #1113.

Currently garmin_txt readers and writers use a homegrown "human" date and time format.  This is converted to a strptime/strftime format in util.cc by convert_human_date_format, convert_human_time_format.

The garmin_txt reader then converts the strptime/strftime format to a QDateTime format, which is remarkably similar to what we started with!  The garmin_txt writer uses strftime.

This PR changes the format for date and time to that of QDateTime, eliminating all the conversions and the use of strftime in garmin_txt.

There are some potential issues with expanding this for general use and thus eliminating strptime/strftime.
1) XcsvFormat::sscanftime counts on strptime not setting fields that aren't used in the format.  To do this with Qt we have to scan the format string.  Scanning the Qt format string requires processing of quoted text, but that isn't too difficult.
2) QDateTime::fromString(const QString &string, const QString &format) has limited support for fractional seconds.  If a decimal point is in the format it is required. ".z" expects 1-3 digits.  ".zzz" expects exactly three digits.  Note strptime/strpftime didn't support fractional seconds at all.
3) The entire format string must match, partial matches are not supported.

As an aside, the txt/csv format currently used by garmin is incompatible with our implementation, which works with a historical garmin txt/csv format.  I suspect our garmin_txt format is mostly useful as a playground for development and to get test coverage of certain features such as garmin_fs category support and init files.

### DIFF
--- a/garmin_txt.cc
+++ b/garmin_txt.cc
@@ -382,7 +382,7 @@ print_date_and_time(const QDateTime& dt)
     return;
   }
   if (gtxt_flags.utc) {
-    *fout << dt.toUTC().addSecs(utc_offs).toString(date_time_format);
+    *fout << dt.toOffsetFromUtc(utc_offs).toString(date_time_format);
   } else {
     *fout << dt.toLocalTime().toString(date_time_format);
   }
@@ -736,7 +736,7 @@ static void
 garmin_txt_adjust_time(QDateTime& dt)
 {
   if (gtxt_flags.utc) {
-    dt = dt.toUTC().addSecs(dt.offsetFromUtc() - utc_offs);
+    dt.setOffsetFromUtc(utc_offs);
   }
 }
 
--- a/garmin_txt.cc
+++ b/garmin_txt.cc
@@ -29,10 +29,9 @@
 #include <cctype>                  // for toupper
 #include <cmath>                   // for fabs, floor
 #include <cstdint>                 // for uint16_t
-#include <cstdio>                  // for sscanf, fprintf, snprintf, stderr
-#include <cstdlib>                 // for abs
-#include <cstring>                 // for strstr, strlen
-#include <ctime>                   // for time_t, gmtime, localtime, strftime
+#include <cstdio>                  // for sscanf, fprintf, stderr
+#include <cstdlib>                 // for abs, div
+#include <cstring>                 // for strstr
 #include <optional>                // for optional
 #include <utility>                 // for pair, make_pair
 
@@ -52,7 +51,6 @@
 #include "formspec.h"              // for FormatSpecificDataList
 #include "garmin_fs.h"             // for garmin_fs_t
 #include "garmin_tables.h"         // for gt_display_modes_e, gt_find_desc_from_icon_number, gt_find_icon_number_from_desc, gt_get_mps_grid_longname, gt_lookup_datum_index, gt_lookup_grid_type, GDB, gt_get_icao_cc, gt_get_icao_country, gt_get_mps_datum_name, gt_waypt_class_names, GT_DISPLAY_MODE...
-#include "inifile.h"               // for inifile_readstr
 #include "jeeps/gpsmath.h"         // for GPS_Math_Known_Datum_To_UTM_EN, GPS_Math_WGS84_To_Known_Datum_M, GPS_Math_WGS84_To_Swiss_EN, GPS_Math_WGS84_To_UKOSMap_M
 #include "src/core/datetime.h"     // for DateTime
 #include "src/core/logging.h"      // for Fatal
@@ -84,7 +82,7 @@ static const char* datum_str;
 static int current_line;
 static QString date_time_format;
 static int precision = 3;
-static time_t utc_offs = 0;
+static int utc_offs = 0;
 static gtxt_flags_t gtxt_flags;
 
 enum header_type {
@@ -122,7 +120,7 @@ static std::array<QList<std::pair<QString, int>>, unknown_header> header_mapping
 static QStringList header_column_names;
 
 static constexpr double kGarminUnknownAlt = 1.0e25;
-static constexpr char kDefaultDateFormat[] = "dd/mm/yyyy";
+static constexpr char kDefaultDateFormat[] = "dd/MM/yyyy";
 static constexpr char kDefaultTimeFormat[] = "HH:mm:ss";
 
 static bool is_valid_alt(double alt)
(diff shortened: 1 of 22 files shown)


tsteven4:
@robertlipe before I explore expanding this idea to replace strptime/strftime I'd like to get your opinion on this approach.

garmin_txt.cc
```diff
@@ -190,16 +189,9 @@ get_option_val(const char* option, const char* def)
 static void
 init_date_and_time_format()
```
robertlipe, line 190: This is a constructor wearing C groucho glasses?

garmin_txt.cc
```diff
@@ -364,30 +356,37 @@ print_position(const Waypoint* wpt)
 }
 
 static void
-print_date_and_time(const time_t time, const bool time_only)
+print_duration(int time)
```
robertlipe, line 359: Consider a name for time that hints what it is. 'seconds', I think? print_duration_as_hms() ? I'm not sure.

garmin_txt.cc
```diff
@@ -444,7 +443,7 @@ print_distance(const double distance, const bool no_scale, const bool with_tab,
 }
 
 static void
-print_speed(const double distance, const time_t time)
+print_speed(const double distance, int time)
```
robertlipe, line 446: While you're in here, can ouo please do a sarch and replace and make these 'meters' and 'seconds' maybe?

robertlipe (commented):
As you note, I'm not sure why 'garmin_txt' exists. I thought we killed this before. I think it's like a copy/paste of some ancient mapsource screen or something that's not actually read or written by anything, but I could be wrong. (We definitely had something like that in our past, though.)

You're right that on their own, convert_human_foo and strftime_to_foo look pretty ridiculous. (It also looks like the authors of each should talk to the other...which is pretty sad, because I'm pretty sure they're both me. :-) ) My master plan was to internally replace strftime/strptime and friends with QDateTime formatters, but keep the stuff visible in the style files as C89 specifiers for compatibility.  My overkill kicked in and I thought about adding versioning to the style files and converting (or not) based on a found version number/token so we could preserve any (likely non-existent) style files that aren't checked in. Thus I turned a 20 minute search and replace into a deadlocked mess. Sorry.

A version number "2.0" would be a good time to rip off that band-aid, fix the style files, and just say that QDateTime specifiers are what we use, though we might be painting ourselves into QDateTime corner instead of a C89 corner when there's a third or even fourth unpainted corner in the room, std::chrono and our (currently short-lived) friend, std::format, though I think we're pretty locked in to Qt at this point.

Reference: https://en.cppreference.com/w/cpp/chrono/duration/formatter

So for the questions that I think you did ask, and some bonus that you didn't...

I agree that two is probably too many. (Three formatting specifier standards is definitely too many.)


1) sscanftime() has very few callers and is visible in only two tokens. addhms() is weird. We have latitude (especially if we're willing to change the style files or subtly change how XT.*TIME works) on changing calling conventions and/or results. Maybe we just change the rules and say that you can only express times as an HMS molecule instead of  H, M, and S atoms. Someone with a spreadheet with separate fields like that might be sad, but I think we can justify that change for us to stay out of the token processign business ourselves.

2) Fractional seconds in C89 was one reason we landed up in Qt land.  Is the issue that it misparses > 3 digits? so hh:mm:ss.1234 doesn't get that .4Ms? I don't think that makes me sad. Milliseconds is a very reasonable bottom floor when parsing ASCII input, IMO.

3) Are you saying the reading and writing pairs must match? That seems natural. Are you saying that mixing Qt-isms and C89-isms in one format string doesn't work?  I'd treat that as a given as there are some tokens with different definitions.


Aside: we have extremely wide latitude in what we do with garmin_txt, up to and including vaporizing it. I was just experimenting with the conversion of format strings there as it had a small blast radius if I broke it.

Aside2: I was never really convinced our .ini files were that useful. Do we have evidence people use them? Ditto sessions. In hindsight, I probably let Olaf run a bit too wild at the time... (Notice how many people missed CET?)

We have only 16 style files. I'd wager that way more than half of them could disappear without notice. arc, custom, garmin_poi might be the only ones of those that are really used by end users. The rest are just kind of along because styles were meant to be super light-weight alternatives to writing C.  IMO, our unicsv took a lot of that load and there just aren't free text formats in programs like there used to be. openoffice and tabsep exist to interface with other programs, like awk or perl or whatever. I DO think that GSAK uses its own custom style for internal data trading with us, but they're frozen on an ancient GPSBabel build and unlikely to track us anyway.  So I suspect we really can quit dancing around the compatibility on those files.

tsteven4:
> 2\. Fractional seconds in C89 was one reason we landed up in Qt land.  Is the issue that it misparses > 3 digits? so hh:mm:ss.1234 doesn't get that .4Ms? I don't think that makes me sad. Milliseconds is a very reasonable bottom floor when parsing ASCII input, IMO.

No, the issue is QDateTime::fromString returns an invalid date time.
```
"0" "s" QDateTime(1900-01-01 00:00:00.000 MST Qt::LocalTime)
"0." "s" QDateTime(Invalid)
"0.0" "s" QDateTime(Invalid)
"0.00" "s" QDateTime(Invalid)
"0.000" "s" QDateTime(Invalid)
"0.0000" "s" QDateTime(Invalid)
"10" "s" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10." "s" QDateTime(Invalid)
"10.0" "s" QDateTime(Invalid)
"10.00" "s" QDateTime(Invalid)
"10.000" "s" QDateTime(Invalid)
"10.0000" "s" QDateTime(Invalid)
"0" "s.z" QDateTime(Invalid)
"0." "s.z" QDateTime(Invalid)
"0.0" "s.z" QDateTime(1900-01-01 00:00:00.000 MST Qt::LocalTime)
"0.00" "s.z" QDateTime(1900-01-01 00:00:00.000 MST Qt::LocalTime)
"0.000" "s.z" QDateTime(1900-01-01 00:00:00.000 MST Qt::LocalTime)
"0.0000" "s.z" QDateTime(Invalid)
"10" "s.z" QDateTime(Invalid)
"10." "s.z" QDateTime(Invalid)
"10.0" "s.z" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.00" "s.z" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.000" "s.z" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.0000" "s.z" QDateTime(Invalid)
"0" "s.zzz" QDateTime(Invalid)
"0." "s.zzz" QDateTime(Invalid)
"0.0" "s.zzz" QDateTime(Invalid)
"0.00" "s.zzz" QDateTime(Invalid)
"0.000" "s.zzz" QDateTime(1900-01-01 00:00:00.000 MST Qt::LocalTime)
"0.0000" "s.zzz" QDateTime(Invalid)
"10" "s.zzz" QDateTime(Invalid)
"10." "s.zzz" QDateTime(Invalid)
"10.0" "s.zzz" QDateTime(Invalid)
"10.00" "s.zzz" QDateTime(Invalid)
"10.000" "s.zzz" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.0000" "s.zzz" QDateTime(Invalid)
"0" "ss" QDateTime(Invalid)
"0." "ss" QDateTime(Invalid)
"0.0" "ss" QDateTime(Invalid)
"0.00" "ss" QDateTime(Invalid)
"0.000" "ss" QDateTime(Invalid)
"0.0000" "ss" QDateTime(Invalid)
"10" "ss" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10." "ss" QDateTime(Invalid)
"10.0" "ss" QDateTime(Invalid)
"10.00" "ss" QDateTime(Invalid)
"10.000" "ss" QDateTime(Invalid)
"10.0000" "ss" QDateTime(Invalid)
"0" "ss.z" QDateTime(Invalid)
"0." "ss.z" QDateTime(Invalid)
"0.0" "ss.z" QDateTime(Invalid)
"0.00" "ss.z" QDateTime(Invalid)
"0.000" "ss.z" QDateTime(Invalid)
"0.0000" "ss.z" QDateTime(Invalid)
"10" "ss.z" QDateTime(Invalid)
"10." "ss.z" QDateTime(Invalid)
"10.0" "ss.z" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.00" "ss.z" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.000" "ss.z" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.0000" "ss.z" QDateTime(Invalid)
"0" "ss.zzz" QDateTime(Invalid)
"0." "ss.zzz" QDateTime(Invalid)
"0.0" "ss.zzz" QDateTime(Invalid)
"0.00" "ss.zzz" QDateTime(Invalid)
"0.000" "ss.zzz" QDateTime(Invalid)
"0.0000" "ss.zzz" QDateTime(Invalid)
"10" "ss.zzz" QDateTime(Invalid)
"10." "ss.zzz" QDateTime(Invalid)
"10.0" "ss.zzz" QDateTime(Invalid)
"10.00" "ss.zzz" QDateTime(Invalid)
"10.000" "ss.zzz" QDateTime(1900-01-01 00:00:10.000 MST Qt::LocalTime)
"10.0000" "ss.zzz" QDateTime(Invalid)
```

GPSBabelDeveloper:
Gross. So we suppressmtrsiling 0., anything after 3 spaces and reparse.
With s.z on failure?

Is it. More amenable to things that look more like times and have an hours
and minutes component?

PR #1206, opened 2023-11-01
move floor to C++20

tsteven4:
drop Visual Studio 2017.
use default member intializers for bit-fields.
use std::numbers::pi, std::numbers::inv_pi


tsteven4:
We have difficulty with certain combinations of Qt and Xcode.  Qt 6.5.3 doesn't work with Xcode 14.2, but works with 14.3.1.  There are other problematic combinations as well.

moc fails with
> Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/concept:318:1: error: Parse error at "::"

https://bugreports.qt.io/browse/QTBUG-117765
https://bugreports.qt.io/browse/QTBUG-118086

tsteven4:
@robertlipe the downside here is Qt 6.5.3 didn't work with Xcode 14.2.  Qt had a fix in 6.5.4, but that isn't public.  Alternatively Xcode 14.3.1 works.

robertlipe:
Lovely. Thank you. I'd thought about dusting off #1207, but yours is (of course) in much better shape while mine was mostly "try all the things" mode. I may see if there's anything worth actually picking from it as I was mostly just experimenting with what was new. (I'd started a list a few days ago for another project) Offhand:
* std::format Yes! 
* ranges are one of those things that CAN be really overdone for better or worse. Things like manipulating pixels in a buffer becomes SO readable with these, but even things like trying to columnize output into thirds gets easier with the totally non-obvious use (now - ask us in ten years) use of a zip_view adapater.
* endian, byteorder, Welcome to 1988.
* spans are handy for non-owning views (they were called something else in a previous code base I used a lot) and they play realy nicely with ranges. It's a (void*, size) that you can pass to code you don't trust to treat the pointer as a char*. Let's optimizers see intent, too. Some of our CSTR() style grief could turn into those.
* Formal declaration that 2s compliment won...yet denial that integers actually work like they do on everything made in the last 40 years.
* map.contains is nice, but we don't use map that isn't QMap.
* initializers in loop:for is nice eye candy, but not life-changing.
* I Have yet to see modules working anywhere sensibly, even now. (It probably does inside my former walled garden...)
* I think that chrono is finally complete enough that we could replace QDate, but that doesn't sound like much fun at all.
* I can't believe that in 2020 starts_with() and ends_with() are headiners...
* to_array fixes the (totally dumb) requirement of needing to specify the number of initializers in the decl. std::array<int, 3> = {0, 1, 2} becomes std::array<int> = to_array({0, 1, 2}) and in some cases (that might be one of them) the to_array() is even implicit. I think there's an ambiquity if the array is complex numbers or something.
* coroutines seem like something I would care about if I were writing protocols, but I've been looking for a foot that fits that shoe for a few years now just to try it on.
* I don't use popcount and rotate here, but I have other projects where it'll be nice to dump the asms for the inevitable 1 opcode functions. See above about single opcode tasks.
* mindpoint and lirp/lerp seem like they should be useful in some of our filters, but weve never really worried about internal overflow, so maybe they're just for bigger problems than we have.
* As a systems programmer to_address() just makes me mad at how out-of-touch the committee is with that world. You really think I'm going to memcpy bytes into a pointer because you don't think I should be casting bit patterns into register bank addresses? You think I'm going to use to_address and from_address? Turkeys.
If I were building GPSBabel from scratch today, lat and lon would have unique types as would radians and degrees. Multiplying something bo 90_d to get a quarter turn and not pass the wrong "double" to things like grt_circle_distance() would be so nice. But not nice enough to open the code up and try to retrofit it.

The colors of https://en.cppreference.com/w/cpp/20 are interesting. The narrowing and widening of this table from release to release is, too. Several of the players from the C++11 days are just gone. This project relies on those first five columns. I'm not certain we rely on anyplace other than MacOS that has Clang-only/primarily, so maybe we can mentally collapse those, but there's a very solid message that three years after formalization those columns are pretty solidly green. (The ones missing are legit hard ... and ones we don't care about.) 

In a former job, we relied on products built on (because we were the ones building...) Edison and Intel tools. I'm legit surprised to see Nvidia and Cray representing actually making a strong go of their on on keeping up the with language "just" for their numeric computing world. C++ hasn't quite reached the complexity of web browsers where we have WebKit/Blink (where the former is a quite signncant fork) and Gecko as pretty much the only 2/3 [distinct](https://en.wikipedia.org/wiki/Browser_engine#Timeline) code bases capable of consuming that specification.

It's interesting the all of the entries on the left (general purpose computing) are multi-platform and all but one are open-source. We sure didn't see THAT coming in the 80's and 90's when a compiler seat license was $4500...and the one that's not open source but now a rather solid green column was so so incredibly terrible then.

Maybe the development and the developers have hit an equilibrium and pace where the survivors are actually keeping up with the produced standards... Yay? Still, I'm glad I'm not in the compiler business. Beyond EDG and MS, it's not clear there still even IS a compiler business since the other contestants are all propped up by platform sales and not really reliant upon sales of a purchaseable product called a "compiler". 

Anyway...


I'm not that worried about building on old Macs. We haven't had anyone report that they were building on a Mac (or Windows) in many years. 

My 2012 iMac (the first Ivy Bridge ones with the thin edges that make them Extra Hard to work on) hasn't been able to build GPSBabel well in a long time. My M1 handles it like a breeze. If Tim Apple doesn't mess it up and I can get a M4 Mini in Q4 or so, I'll be looking for one of those - and a new monitor to replace this perfectly lovely one that's attached to it since they REMOVED the ability to use target display. (Hmm. Maybe i could boot from an ancient OS and run it as a Thunderbolt monitor...) Four 3.4Ghz cores with 33G is still not a _bad_ computing environment, if only it ran a supported OS. I'm annoyed that they don't support older hardware than they do, but that's the tradeoff.

humminbird.cc
```diff
@@ -185,9 +186,9 @@ HumminbirdBase::geocentric_to_geodetic_hwr(const double gc_lat)
 {
   constexpr double cos_ae = 0.9966349016452;
   constexpr double cos2_ae = cos_ae * cos_ae;
-  const double gcr = gc_lat *M_PI / 180.0;
+  const double gcr = gc_lat *  std::numbers::pi / 180.0;
```
robertlipe: There might be a stray space in there.

robertlipe (approved):
Awesome. Thank you.

tsteven4:
> * If I were building GPSBabel from scratch today, lat and lon would have unique types as would radians and degrees. Multiplying something bo 90_d to get a quarter turn and not pass the wrong "double" to things like grt_circle_distance() would be so nice. But not nice enough to open the code up and try to retrofit it.

There is a least one outstanding related bug.  The bend filter passes locations in radians to linepart. and expects the returned position to be in radians.

tsteven4:
another bug, simplify passes locations in degrees to gcdist.

tsteven4:
My experiments with chrono and gps_clock/utc_clock revealed that the first supported release was [releases/gcc-13.1.0](https://github.com/gcc-mirror/gcc/releases/tag/releases/gcc-13.1.0)

the feature test macro __cpp_lib_chrono was not helpful to distinguish between 11.4 (jammy) and 13.2 (noble), both show 202002L.

https://github.com/gcc-mirror/gcc/commit/1736bf5a61c7364c5da6fa52e5e2bfdbc9507c97#diff-56d7e34fe989c005ccd0599f1772bf7180bf5bcbc4e5c9c359e3e463492e6720

robertlipe:
Yeah, the feature test stuff tends to work really well with language
features and less well for library features.

If it helps us move a generation, I don't think I mind carrying a
patched/replacement/subsetted header-only implementation like this around
for a little while, perhaps shoving it into a special namespace or
something.

The papers say Jammy is around until 2034, but I certainly have no
intention of it being a gate for the next ten years. We've seen that people
stay on LTS releases to use whatever ancient versions come with the OS and
resist moving from that and most of our Linux (human) users are on more
traditional desktoppy, contemporary-tracking, non-LTS versions.

I get that "the most recent LTS" isn't something to casually toss away, but
I don't mind some grubbiness to let us NOT hold things back...but we're
already using C++20 in at least some places so maybe we just have some of
those "C++ version X except for library features Y because it causes
problems on platform Z" conditions that reality has forced us to carry
around since, uhm, forever.

tsteven4:
I think chrono is problematic both with libstdc++ versions and Qt versions.  

1. Despite the Qt documentation QDateTime::fromStdTimePoint(std::chrono::time_point<std::chrono::system_clock, std::chrono::milliseconds> time) did not exist before Qt 6.7.0 other than in the documentation!  The latest ubuntu LTS, noble which was released 4/2024, ships with Qt 6.4.2.
2. The existence of QDateTime::fromStdTimePoint(const std::chrono::time_point<Clock, Duration> &time)
depends on __cpp_lib_chrono >= 201907L and  __cpp_concepts >= 201907L.  The latest ubuntu LTS, noble, has libstdc++ 13.2 with  __cpp_lib_chrono: 201611L

And all I was playing with was SkytraqBase::gpstime_to_qdatetime and it's fiddling with leap seconds.  It's not worth it.

robertlipe:
> And all I was playing with was SkytraqBase::gpstime_to_qdatetime and it's
fiddling with leap seconds. It's not worth it.

Agreed. That's several levels deep of minority cases. Walk away. Run!

PR #1208, opened 2023-11-02
refactor garmin_txt date time handling

tsteven4:
This is an alternative to #1205.

Instead of using QDateTime format strings directly, we translate our own "human" format strings to QDateTime format strings.  This means there isn't a user interface change for garmin_txt which isn't important.  

If we extend this to retire strptime/strpftime, we may want a user interface change there from strptime/strftime format strings to our "human" strings.  I would argue that would be a better direction than attempting to translate the more complex strptime/strftime strings, and I think our next release is 2.0.0 which should give us some liberty to change the user interface.  Our human formats offer more limited functionality than strptime/strftime, making it easier to change the underlying implementation as this PR does.  It would also be consistent across formats & filters.

Neither our human format strings or strptime/strftime format strings support fractional seconds.  To cover the existing use cases of strptime/strptime this isn't necessary.  Qt's implementation for fractional seconds with user supplied format strings is problematic with fractional seconds as was discussed in #1205.

Escaping a single quote in a QDateTime::fromString format string is under-documented, and different from QDateTime::toString!  The details are in convert_human_date_format and convert_human_time_format.

### DIFF
--- a/reference/garmincategories.gpx
+++ b/reference/garmincategories.gpx
@@ -5,7 +5,7 @@
     <bounds minlat="39.973869717" minlon="-105.498962400" maxlat="40.003967283" maxlon="-105.465850367"/>
   </metadata>
   <wpt lat="39.973869717" lon="-105.465850367">
-    <time>2013-03-09T20:45:12Z</time>
+    <time>2013-03-09T08:45:12Z</time>
     <name>Hwy 119</name>
     <cmt>The Diagonal</cmt>
     <desc>The Diagonal</desc>
--- a/defs.h
+++ b/defs.h
@@ -966,8 +966,8 @@ QDateTime dotnet_time_to_qdatetime(long long dotnet);
 long long qdatetime_to_dotnet_time(const QDateTime& dt);
 QString strip_html(const QString& utfstring);
 QString strip_nastyhtml(const QString& in);
-QString convert_human_date_format(const char* human_datef, bool read);	/* "MM,YYYY,DD" -> "%m,%Y,%d" */
-QString convert_human_time_format(const char* human_timef, bool read);	/* "HH+mm+ss"   -> "%H+%M+%S" */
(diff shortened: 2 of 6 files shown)


tsteven4:
@robertlipe do you like this direction better than #1205?

I skipped your suggested renames for now to minimize the diffs.

tsteven4:
One problem with generalizing our human date and time formats is that they both use '[mM]' for different things (month or minute), which makes supporting a human datetime format as xcsv sometimes uses, e.g. GMT_TIME, problematic.  garmin_txt assumes date always precedes time, so if that is general enough for xcsv it is a way out.

tsteven4:
> so if that is general enough for xcsv it is a way out.

not really.  GMT_TIME has one format string and it isn't clear we could break it in pieces.

GPSBabelDeveloper:
That's a more shallow cut that delivers most (all) of the results. It's
hard to argue with results. I'm quite OK with this. It keeps us out of lots
of unanticipated trouble.

Add a static_assert() in main or something that ensures the escaping
remains the same.

Maybe strftime_to_timespec() was just a bad idea, trying to solve a problem
better fixed at the source.

Go for it.

tsteven4:
I intend to make the human formats case sensitive, which allows distinguishing **m**inutes and **M**onths as in ISO8601, Qt:DateTime, etc.

reference/garmincategories.gpx
```diff
@@ -5,7 +5,7 @@
     <bounds minlat="39.973869717" minlon="-105.498962400" maxlat="40.003967283" maxlon="-105.465850367"/>
   </metadata>
   <wpt lat="39.973869717" lon="-105.465850367">
-    <time>2013-03-09T20:45:12Z</time>
+    <time>2013-03-09T08:45:12Z</time>
```
robertlipe, line 8: Was this an intentional change from 24H to 12H?

robertlipe (approved):
It seems we've had some related work in this area in just the last few weeks...

But I'm OK with the direction here. Your call whether this goes in or not.

PR #1215, opened 2023-11-13
jeeps function pointers, GPS_PPacket const correctness, GPS_Serial_[OSP]Packet -> GPS_Serial_Packet

tsteven4:
use using for jeeps function pointers.

const correctness for GPS_PPacket use.  _The important bit here is in gpssend.cc, Build_Serial_Packet used to receive a 1KB packet by copy, now it is by const reference._

original jeeps used GPS_SXXX, GPS_OSXXX, GPS_PXXX for the structure
tag, a typedef of the structure, and a pointer to a typedef of the
structure.  In the case of GPS_PPacket this convention was no longer
followed.  rename GPS_PPacket -> GPS_Packet.

GPS_Serial_[SOP]Packet -> GPS_Serial_Packet, GPS_Serial_Packet*.


robertlipe:
Good call. This part is the code has had a hard life.

PR #1217, opened 2023-11-14
use c++11 fixed width integer types in jeeps.

tsteven4:
thanks resharper, it just took a little preparation and help with non-widows code.

### DIFF
--- a/CMakeSettings.json
+++ b/CMakeSettings.json
@@ -1,0 +1,26 @@
+﻿{
+	"configurations": [
+		{
+			"name": "x64-Debug",
+			"generator": "Ninja",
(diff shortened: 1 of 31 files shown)


tsteven4:
any reason not to do these as well?
```
typedef unsigned char UC;
typedef uint16_t      US;
```

robertlipe:
If it's easy, go for it's please.

tsteven4:
It is quite easy to do an "inline" refactoring of a typedef like these.  However, the refactoring isn't offered for the equivalent using statement.

In any event I am going to postpone inlining of UC and US to finish some other work.

PR #1222, opened 2023-11-16
convert garmin to Format class

tsteven4:
This has been tested with a GPS12 for read, write and read of real time position on windows.


robertlipe (approved):
Man, you make this look pretty mechanical these days!

PR #1226, opened 2023-11-18
convert mtk formats to Format class.

tsteven4:


tsteven4:
This formats use of a static local variable in MtkLoggerBase::mtk_parse is questionable.  This is only relevant if the reader is used more than once.

PR #1234, opened 2023-11-19
simplify usage of homebrew on CI macos-11.

tsteven4:
homebrew installs are taking an extreme amount of time on macos-11, with the following message:
  Warning: You are using macOS 11.
  We (and Apple) do not provide support for this old version.
  It is expected behaviour that some formulae will fail to build in this old version.
  It is expected behaviour that Homebrew will be buggy and slow.
To alleviate this pain skip the installation of jing and all it's dependencies on macos-11.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -75,7 +75,10 @@ jobs:
         # brew update # skip update for now to avoid link issues AND many slow dependency upGRADES.
         brew install ninja
         brew install docbook docbook-xsl fop gnu-sed
-        brew install jing-trang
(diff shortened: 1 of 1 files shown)


robertlipe:
Yep. Not everything needs to be exercised in every combination. Good call.

On the older MacOS versions, Homebrew gets increasingly annoying. You try
to install libfoo which depends on python to print a error message so the.
It runs off and builds python, but that wants a specific version of GCC
which is no longer bottled, so it runs off and builds THAT and so on.

Examples made up and picked for absurdity, but I know I ran into something
"simple" that turned into 2 hours of build time because of dominos like
that.

PR #1238, opened 2023-11-20
Remove XCSV_ENABLED, MAXIMAL_ENABLED, MINIMAL_ENABLED

robertlipe:
- Remove XCSV_ENABLED, MAXIMAL_ENABLED, MINIMAL_ENABLED


robertlipe:
OMG, how I hate these commits from years ago come back to haunt us in Git sometimes.

vecs.cc
```diff
@@ -145,10 +140,9 @@ struct Vecs::Impl {
   GarminFitFormat format_fit_fmt;
   GeoJsonFormat geojson_fmt;
   GlobalsatSportFormat globalsat_sport_fmt;
-#endif // MAXIMAL_ENABLED
+  QstarzBL1000Format qstarz_bl_1000_fmt;
```
tsteven4: We only need static instances of formats that aren't dynamic.  The QStarzBL1000 is dynamic, so no static instance is needed.  A format is dynamic if the last entry (factory) is not nullptr, in which case it should be a reference to fmtfactory and the first entry (vec) should be nullptr.

I have lazily been converting format to dynamic.  It involves an evaluation of all the data members, and any lurking static locals, to decide if the format is intentionally storing data between uses, e.g. the gpx read version.  Additionally the Format constructor must be used ("using Format::Format;").  This makes Format::fname available, which can alleviate the need for rd[wr]_init, e.g. geo.

robertlipe: Integrated. Thanx.

CMakeLists.txt
```diff
@@ -338,8 +338,8 @@ list(SORT HEADERS)
 target_sources(gpsbabel PRIVATE ${SOURCES} ${HEADERS})
 
 # We don't care about stripping things out of the build.  Full monty, baby.
-target_compile_definitions(gpsbabel PRIVATE MAXIMAL_ENABLED)
 target_compile_definitions(gpsbabel PRIVATE FILTERS_ENABLED)
+target_compile_definitions(gpsbabel PRIVATE SHAPELIB_ENABLED)
```
tsteven4, line 342: This will break GPSBABEL_WITH_SHAPELIB:STRING=no.  This define is handled in shapelib.cmake, which is included.  Please remove the SHAPELIB_ENABLED change.

tsteven4 (changes requested):

tsteven4:
Also, can we remove` target_compile_definitions(gpsbabel PRIVATE CSVFMTS_ENABLED)` from CMakeLists.txt now?

tsteven4:
Suggested fixes in #1339

PR #1239, opened 2023-11-20
Remove unused shortname options

### LINKED ISSUE #982: garmin, ozi, xcsv mkshort options whitespace_ok, must_unique broken since 2006

robertlipe:
- PDF maker: format screen and input same. Shrink padding slightly.
- Disallow 'fields' option when writing unicsv.
- Remove mkshort options. Fixes #982


robertlipe:
Merge note: I think that this branch has aged badly and it's confused by code that moved from .cc -> .h . Needs attention.

Issue #1241, opened 2023-11-22
garmin gpi corruption of bitmap color data for bitmaps with 24 bits per pixel.


tsteven4:
We attempt to convert bitmaps with 24 bits per pixel to 32 bits per pixel 

> 24 bpp seems to be not supported, convert to 32 bpp

https://github.com/GPSBabel/gpsbabel/blob/e251cac4f51c857435bb8e64d003a2d4db4e8f13/garmin_gpi.cc#L1179-L1180

gbfgetint16 returns an int16_t.  If the most significant bit is set, then the cast to int32_t will sign extend, resulting in an int32_t value with the two most significant bytes being 0xff.  At that point the value of the third byte read by gbfgetfc is irrelevant.

Example:

a 24 bit color of 0x5a84b6
results in a 32 bit color of 0xffff84b6

tsteven4:
fixed in #1243

Issue #1242, opened 2023-11-22
erroneous garmin gpi warning when writing with a 24 or 32 bits per pixel bitmap


tsteven4:
When using the bitmap option with a bitmap that don't use a color table (e.g. 24 or 32 bits per pixel) an inappropriate warning may be output such as
`garmin_gpi: Code error in load_bitmap_from_file, expected output size 1060, actual output -28.`

The warning is only valid for bitmaps that use color tables (e.g. 8 bits per pixel).

tsteven4:
fixed in #1243

PR #1243, opened 2023-11-22
fix errors in gpi writer related to bitmaps.

tsteven4:
Erroneous warnings "garmin_gpi: Code error in load_bitmap_from_file, expected output size 1060, actual output -28."

Corrupted colors when input bitmap has 24 bits per pixel.


tsteven4:
This resolves #1241 and #1242.

PR #1258, opened 2024-01-30
update zlib to 1.3.1

tsteven4:


robertlipe:
LGTM

Readying for C23 by finally letting go of styles that were deprecated in C89.

PR #1268, opened 2024-04-16
support builds with shapelib (as documented).

tsteven4:
we document the ability to build without shapelib:
```
GPSBABEL_WITH_SHAPELIB:STRING=no|pkgconfig|included*|custom
  no: build without shapelib.  functionality will be limited.
```
This is supported in shapelib.cmake, but CMakeLists.txt defeats this functionality by always defining SHAPELIB_ENABLED.


robertlipe:
There was a time I cared about requiring libs of many hundreds of bytes.

That time has past. Maybe we just say we require what we require and quit fretting over the combinations.

A one-line fix isn't exactly the breaking point. I'm just saying that in the future (the current) let's not worry much about this kind of thing. I'd rather have one build that works really well than six builds that support different subsets of $THINGS.

robertlipe (approved):

PR #1283, opened 2024-05-23
refactor PVT interface

tsteven4:
- eliminate unused parameters
- pass gpsdevh* instead of gpsdevh** when function doesn't modify the original.


robertlipe:
Lgtm

PR #1285, opened 2024-06-12
delete obsolete jeeps functions that were never used

tsteven4:
since introduction in 2002.


robertlipe:
Lgtm

I'm sure you could find more in math and friends... They don't particularly
bother me.

Maybe we should build and link with function sections...

PR #1287, opened 2024-06-28
Fix unicsv reader delimiter detection

### LINKED ISSUE #1286: When input cvs fields contain `; `, gpsbabel throws the "unicsv" Invalid latitude error
When input cvs fields contain `; `, gpsbabel throws the `"unicsv" Invalid latitude 1e+25 in waypoint` error on the first data line (tried on various lines).

It works as expected with a field structure (header and data) of separated fields like so:
`"lat","lon","name","foo","bar","baz","blah","boom", "...", "..."`

With some fields (`foo` and several after it) beforehand concatenated and separated with ` – ` or with `# `, all is well also. This works:
`"lat","lon","name","foo – bar – baz – blah – boom – ...", "..."`
`"lat","lon","name","foo# bar# baz# blah# boom# ...","..."`

But if the separator of these already merged fields in the input are separated with a semicolon, the latitude (incidentally the first data field to be processed, maybe that means something?) is `Invalid`.
`"lat","lon","name","foo; bar; baz; blah; boom; ...","..."` causes the error.

`gpsbabel` is called with `"${<path/to/gpsbabel>}" -i unicsv -f "${TEMPFILE}" -o gpx -F "${OUTFILE}"`.

This is different from #591, as no `fields option` is involved.

Observed with `GPSBabel Version 1.9.0`
on `openSUSE Tumbleweed 20240624`, 
with `Kernel 6.9.5-1-default (64-bit)`.

tsteven4:
Any delimiter sequence that is enclosed has never been considered
a delimiter when splitting a line into fields.  However, previously
an enclosed potential delimiter could be detected when scanning the
first line, and then erroneously used as the delimiter.

This resolves #1286 .

### DIFF
--- a/csv_util.cc
+++ b/csv_util.cc
@@ -148,7 +148,8 @@ csv_dequote(const QString& string, const QString& enclosure)
 /*****************************************************************************/
 QStringList
 csv_linesplit(const QString& string, const QString& delimited_by,
-              const QString& enclosed_in, const int line_no, CsvQuoteMethod method)
+              const QString& enclosed_in, const int line_no, CsvQuoteMethod method,
+              bool* delimiter_detected)
 {
   QStringList retval;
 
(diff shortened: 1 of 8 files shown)


tsteven4:
The test demonstrates a strange mapping in the gpx writer where Waypoint::description becomes the content of the gpx cmt element and Waypoint::notes becomes the gpx desc element.

PR #1290, opened 2024-07-09
fix faulty assertion in garmin.

tsteven4:
This was reported on gpsbabel-code by Dirk.

> I'm using GPSBabel with a Garmin eTrex H. Recently updated to Ubuntu MATE 24.04 LTS and since then GPSBabel crashes with the following error message:

> gpsbabel: ./garmin.cc:380: void rw_init(const QString&): Zusicherung »!QString(valid_waypt_chars).contains('-')« nicht erfüllt.

> '-' can be a valid waypoint character for some older Garmin devices as correctly defined in garmin.cc.
> Hence line 306 of the current garmin.cc

> assert(!QString(valid_waypt_chars).contains('-'));

> makes GPSBabel unusable for those devices.

The purpose of the assertion was to make sure we didn't have any PCRE metacharacters in our character class definition that we invert to get the invalid characters.  In practice the dash was the last character in the class so wouldn't be considered a metacharacter anyway, but our assertion check wasn't sophisticated enough to allow this.

It appears that some distributions don't set CMAKE_BUILD_TYPE to something like Release that causes NDEBUG to be defined, thus assertions are enabled in their released code.

In any event, now that we have a requirement for Qt >= 6.2, we can use QString::removeIf(Predicate pred) and avoid QRegularExpression entirely.

### DIFF
--- a/garmin.cc
+++ b/garmin.cc
@@ -21,17 +21,18 @@
 
 #include "garmin.h"
 
-#include <cassert>               // for assert
 #include <climits>               // for INT_MAX
 #include <cmath>                 // for atan2, floor, sqrt
 #include <cstdio>                // for fprintf, fflush, snprintf, snprintf
+#include <cstdint>               // for int32_t
 #include <cstdlib>               // for strtol
 #include <cstring>               // for memcpy, strlen, strncpy, strchr
 #include <ctime>                 // for time_t
 #include <utility>               // for as_const
 
 #include <QByteArray>            // for QByteArray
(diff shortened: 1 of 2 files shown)


tsteven4:
I have verified that gpsbabel_1.9.0+ds-3_amd64.deb from debian sid has assertions enabled.

tsteven4:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076083

PR #1307, opened 2024-07-29
use a signed type for waypt_counts and route counts.

tsteven4:
Although larger in Qt6, the underlying type is signed.

This reduces the number of tidy bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions warnings by 108, without picking up any other tidy warnings.

Also see the discussion at https://google.github.io/styleguide/cppguide.html#Integer_Types


tsteven4:
-Wformat= warnings reduced by 3.
-Wsign-conversion reduced by 52.

PR #1311, opened 2024-08-07
Use Helmert transform for conversions between OSGB36 and WSG84

tsteven4:
This came up on:
[Gpsbabel-misc] Accuracy of BNG > GPX (WGS84) transformation
Mon, 5 Aug 2024 19:37:42 +0200

Previously we used the Molodensky transform.  This PR adds an implementation of the Helmert transform, and uses it to transform between OSGB36 and WGS84.  Transforms between other coordinate systems still use Molodensky.  Using the Helmert transform for other use cases would require knowledge of the translation, rotation and scaling parameters between those coordinate systems.

https://www.ordnancesurvey.co.uk/documents/resources/guide-coordinate-systems-great-britain.pdf states that we should expect errors up to 3.5m with the Helmert transform.

This does NOT implement OSTN15.  If you really need better accuracy then you should convert your coordinates outside of GPSBabel.

### DIFF
--- a/jeeps/gpsmath.cc
+++ b/jeeps/gpsmath.cc
@@ -1543,17 +1543,16 @@ void GPS_Math_Helmert(double Sx, double Sy, double Sz,
                       double sppm,
                       double rXs, double rYs, double rZs)
 {
-  double s = sppm/1000000.0;
-  double rX = (rXs/(60.0*60.0))*GPS_PI/(180.0); /* radians */
-  double rY = (rYs/(60.0*60.0))*GPS_PI/(180.0); /* radians */
-  double rZ = (rZs/(60.0*60.0))*GPS_PI/(180.0); /* radians */
+  double s = sppm * 1e-6;
+  constexpr double radians_per_second = (1.0/(60.0*60.0)) * GPS_PI/180.0;
+  double rX = rXs * radians_per_second; /* radians */
+  double rY = rYs * radians_per_second; /* radians */
+  double rZ = rZs * radians_per_second; /* radians */
 
   // rotation/scaling matrix = [1+s -rz ry;rz 1+s -rx;-ry rx 1+s];
   *Dx = tX + (1 + s)*Sx + -rZ*Sy + rY*Sz;
   *Dy = tY + rZ*Sx + (1 + s)*Sy + -rX*Sz;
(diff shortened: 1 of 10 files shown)


jeeps/gpsmath.cc
```diff
@@ -1516,6 +1517,98 @@ void GPS_Math_Molodensky(double Sphi, double Slam, double SH, double Sa,
...
+                              double rXs, double rYs, double rZs)
+{
+  double s = sppm/1000000.0;
+  double rX = (rXs/(60.0*60.0))*GPS_PI/(180.0); /* radians */
+  double rY = (rYs/(60.0*60.0))*GPS_PI/(180.0); /* radians */
+  double rZ = (rZs/(60.0*60.0))*GPS_PI/(180.0); /* radians */
+
+  // forward rotation/scaling matrix is [1+s -rz ry;rz 1+s -rx;-ry rx 1+s]
+  // compute inverse of forward Helmert rotation/scaling matrix
+  // https://www.wolframalpha.com/input?i2d=true&i=Power%5B%7B%7B1%2Bs%2C-Subscript%5Br%2Cz%5D%2CSubscript%5Br%2Cy%5D%7D%2C%7BSubscript%5Br%2Cz%5D%2C1%2Bs%2C-Subscript%5Br%2Cx%5D%7D%2C%7B-Subscript%5Br%2Cy%5D%2CSubscript%5Br%2Cx%5D%2C1%2Bs%7D%7D%2C-1%5D
```
robertlipe, line 1594: That (this) is both impressive and terrifying.

tsteven4: Within rounding errors GPS_Math_Inverse_Helmert reverses GPS_Math_Helmert.  I coded GPS_Math_Inverse_Helmert when I was developing the test from BNG to WGS84 and back to BNG.  The issue with getting that round trip test to work ended up being elsewhere, so at this point we are just changing the signs of the 7 parameters and using GPS_Math_Helmert to go backwards.  This approximation was mentioned in the ordinance survey reference.  The selection of inversion method in GPS_Math_UKOSMap_To_WGS84_H is done with conditional compilation.  I left GPS_Math_Inverse_Helmert in because it exactly undoes GPS_Math_Inverse_Helmert which might be useful if we have trouble with the approximate method.  I don't have a citation for the exact method, but it is straight forward linear algebra (thanks to WolframAlpha).

PR #1314, opened 2024-08-08
use constexpr for constants, match parms in declaration and definition.

tsteven4:


jeeps/gpsmath.h
```diff
@@ -7,7 +7,7 @@
 #include <QString>   // for QString
 
 
-constexpr double GPS_PI = 3.141592653589;
+constexpr double GPS_PI = 3.14159265358979323846;
```
robertlipe, line 10: Are either M_PI (wasn't stable in C, I think it is in C++) or std::numbers::pi available to us? That might be too new, but it's not like systems haven't provided this in some form forever.

https://en.cppreference.com/w/cpp/numeric/constants

robertlipe: In case MSVC is lagging, they provide them, it just takes extra squeeze according to
https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170
I don't know if it's their doc or their reality in saying "it's not defined in C++" when tht's not been true for years.
Plan E (are we up to F?) might be
constexpr double const_pi() { return std::atan(1)*4; }


Of course, another plan is to quit thinking about this and move on.

If you think it's wise to adop std::numbers, I'll whack the one in defs.h and include and use this in skytraq, grtcirc, humminbird, and lowrance. That's one less file we include in the world.

robertlipe: P.S. I don't care that deeply about this define. I just had a reason to learn about std::numbers for another project just a few weeks ago and thought "finally!", so it left an impression.

tsteven4:
The mathematical constants finally appear in <numbers> in C++20, we require only C++17.   M_PI is POSIX.  It is also in \<QtMath\> as well as qDegreesToRadians and qRadiansToDegrees.

I leaning toward your plan "quit thinking about this and move on", this PR is already merged.

The extra digits did change the mantissa by ~2000 on my 64 bit system.

PR #1316, opened 2024-08-11
correct simplify filter with maximum allowable error option and the metric being overall route length or relative error

tsteven4:
Both these combinations suffered from passing locations in degrees to gcdist instead of the required radians.

For the error option using the overall route length:
- the existing code sets a limit of 1000m, but the actual route length changes by -21.459m leaving 333 points.
- the modified code set a limit of 20.34m and the actual route length changes by -20.259m leaving 333 points.
- the existing and modified code each keep a different set of 333 points.
- the change in route length is lower with the modified code than the existing code even though both keep 333 points.

  | error(m) | total length(km) | delta length(m) | points
-- | -- | -- | -- | --
garmin-edge-800-output.gpx |   | 25.169843 |   | 1286
ref | 1000 | 25.148384 | -21.459 | 333
dut | 20.2 | 25.149752 | -20.091 | 334
dut | 20.34 | 25.149584 | -20.259 | 333
dut | 20.5 | 25.149415 | -20.428 | 332

For the error option using the relative error based on HDOP:
- the new simplify-relative test (without the corrections to the call to gcdist) uses a limit of 34.3 and keeps 33 out of 100 points.
- the new simplify-relative test with the corrections to the call to gcdist uses a limit of 0.61 and keeps the same 33 points.  The limit to keep the same number of points has changed to approximately (pi/180)*34.3 = 0.60.

### DIFF
--- a/smplrout.cc
+++ b/smplrout.cc
@@ -126,8 +126,8 @@ double SimplifyRouteFilter::compute_track_error(const neighborhood& nb) const
                wpt2->latitude, wpt2->longitude,
                frac, &reslat, &reslon);
       track_error = radtometers(gcdist(
-                                  wpt3->latitude, wpt3->longitude,
-                                  reslat, reslon));
+                                  RAD(wpt3->latitude), RAD(wpt3->longitude),
+                                  RAD(reslat), RAD(reslon)));
     } else { // else distance to connecting line
       track_error = radtometers(linedist(
                                   wpt1->latitude, wpt1->longitude,
--- a/smplrout.cc
+++ b/smplrout.cc
@@ -57,7 +57,7 @@
 */
 
 #include <cassert>
-#include <cstdlib>              // for strtol
+#include <cstdlib>              // for strtod, strtol
 #include <iterator>             // for prev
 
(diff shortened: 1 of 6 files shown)


smplrout.cc
```diff
@@ -297,11 +297,15 @@ void SimplifyRouteFilter::init()
     count = strtol(countopt, nullptr, 10);
     break;
   case limit_basis_t::error: {
-    int res = parse_distance(erroropt, &error, 1.0, MYNAME);
-    if (res == 0) {
-      error = 0;
-    } else if (res == 2) { /* parameter with unit */
-      error = METERS_TO_MILES(error);
+    if (metric == metric_t::relative) {
+      error = strtod(erroropt, nullptr);
```
robertlipe, line 301: If we wanted to be all C++17 fancy and make the argument immune to the locale, we could use std::from_chars().

OTOH, maybe that's a feature and not a bug. Maybe we should use things like from_chars when parsing, say, kml where the radix point is ALWAYS a '.' and continue to use strtod and just allow the ambiguity of Europeans using 3,14 as an argument instead of 3.14.

No change requested. Just musing aloud.

PR #1320, opened 2024-08-12
fix bend filter usage of linepart.

### LINKED ISSUE #1317: Bend filter passes and receives position in improper units to/from linedist
The bend filter passes positions in radians to linepart, which expects degrees, and interprets the resulting position as radians instead of degrees.  As a result the interpolated point is  not on the great circle between the two source points.  However it takes long route segments for this to be apparent.
With a test route containing these points (lr.kml)
latitude(deg), longitude(deg)
140.615728,35.752604
166.492918,42.992831
202.123988,46.203079
232.377086,42.925678

```
gpsbabel.ref -i kml -f lr.kml -x transform,rte=trk,del -x bend,distance=1000000,minangle=5 -o kml,line_color=990000ff -F lrref.kml
gpsbabel.dut -i kml -f lr.kml -x transform,rte=trk,del -x bend,distance=1000000,minangle=5 -o kml,line_color=9900ff00 -F lrdut.kml
```

The following results.  Blue is the original trace (lr.kml), red is from the current code (lrref.kml), green is from test code(lrdut.kml) that passes and receives the proper units from linepart.  Each set of code splits original point RPT002 into 2 points approximately 1000km from the original.  It can be seen that the original code creates two new points that are not on the great circle from original points RPT001 to RPT002, and RPT002 to RPT003.  The points created by the modified code are on the appropriate great circles. 
bend-2

The points from the original code are (lrref.kml)
140.615728,35.752604
155.512639,39.920985
179.115698,44.130713
189.440172,45.060921
214.666966,44.844735
232.377086,42.925678

The points from the modified code are (lr.dut)
140.615728,35.752604
154.851042,40.634774
178.578290,45.401172
-170.887231,46.365336
-144.937135,45.830297
232.377086,42.925678

tsteven4:
This resolves #1317.


tsteven4:
Hopefully a subsequent PR will make the interface to the great circle routines more fool proof.  This PR limits itself to resolving the issue and provided test coverage so we can isolate the effects of an interface change.

PR #1321, opened 2024-08-12
use implicit conversions for lat,lon degs<->rads in grtcirc interface

tsteven4:


tsteven4:
@robertlipe I'm still noodling on this, but feel free to have a look.  In addition to automatically converting positions in radians and degrees this also prevents passing lat,lon pairs in the wrong order.  There may be lurking TODOs at this point, squawk if you find some.  
One potential pitfall related to returning a PositionDeg/PositionRad is that if the caller uses auto then the return type is out of the callers control, and if they dig inside they will get degrees or radians depending on the function definition.  OTOH if they assign the returned value to PositionDeg/PositionRad, then they will get that.

tsteven4:
How did we do?

usages of RAD: 0
usages of kRadiansPerDegree: 3 all in defs.h
references to DEG: 1 in garmin.cc
references to kDegreesPerRadian: 4 all in defs.h

bend.cc
```diff
@@ -55,19 +55,16 @@ void BendFilter::init()
 
 Waypoint* BendFilter::create_wpt_dest(const Waypoint* wpt_orig, const Waypoint* wpt_orig_adj) const
 {
-  double distance = radtometers(gcdist(RAD(wpt_orig->latitude), RAD(wpt_orig->longitude),
-                                       RAD(wpt_orig_adj->latitude), RAD(wpt_orig_adj->longitude)));
+  double distance = radtometers(gcdist(wpt_orig->position(),
```
robertlipe, line 58: Since the majority of the callers do this, should we maybe just have a gcdist_meters() and a gcdist_radians()? I don't think we can do an overload based on return type ('auto' wouldn't know how to cope) if we did anything fancy with returning different types - a double wearing groucho glasses called meters or something...

defs.h
```diff
@@ -246,6 +255,35 @@ struct bounds {
   double min_alt;	/* -unknown_alt => invalid */
 };
 
+struct PositionRad; // forward declare
```
robertlipe, line 259: Nice! This is where I was heading, but I got all wrapped up in C++ boilerplate goo. This should mostly melt away at compile time and it REALLY helps the readability of the whole interface.

defs.h
```diff
@@ -246,6 +255,35 @@ struct bounds {
...
+struct PositionDeg
+{
+  PositionDeg() = default;
+  explicit(false) inline PositionDeg(const PositionRad& posr); /* converting ctor */
```
robertlipe, line 263: Don't we WANT the callers to be explicit?
Or are these less aggressively typecast-exploded since they're not templates?

grtcirc.cc
```diff
@@ -68,10 +58,15 @@ double radtometers(double rads)
   return (rads * EARTH_RAD);
 }
 
-double gcdist(double lat1, double lon1, double lat2, double lon2)
+double gcdist(PositionRad pos1, PositionRad pos2)
 {
   errno = 0;
 
+  const double lat1 = pos1.latR;
```
robertlipe, line 65: Here, and in heading, would marking these as const help the optimizer see that we're really just peeking into the arg list and there's no need to allocate stack for these?
Ditto for sdlat/sdlon and the same down in heading().
Then again, the optimizer can probably actually figure out...

grtcirc.cc
```diff
@@ -92,8 +87,13 @@ double gcdist(double lat1, double lon1, double lat2, double lon2)
...
+static double heading(PositionRad pos1, PositionRad pos2)
 {
+  double lat1 = pos1.latR;
+  double lon1 = pos1.lonR;
+  double lat2 = pos2.latR;
+  double lon2 = pos2.lonR;
+
   double v1 = sin(lon2 - lon1) * cos(lat2);
   double v2 = cos(lat1) * sin(lat2) - sin(lat1) * cos(lat2) * cos(lon2 - lon1);
   /* rounding error protection */
```
robertlipe, line 99: Is the hipster way of spelling this if (fabs(v1) < std::numeric_limits<double>::epsilon()) ?

Maybe we intentionally want a higher threshold here of "too small to care" instead of "too small to represent".

I don't mind either way. If it introduces noise in our tests, for example with a bunch of .00000000123 junk of imaginary precision, let's just stay with the existing "too small to coount".

grtcirc.cc
```diff
@@ -68,10 +58,15 @@ double radtometers(double rads)
...
-double gcdist(double lat1, double lon1, double lat2, double lon2)
+double gcdist(PositionRad pos1, PositionRad pos2)
 {
   errno = 0;
 
+  const double lat1 = pos1.latR;
+  const double lon1 = pos1.lonR;
+  const double lat2 = pos2.latR;
+  const double lon2 = pos2.lonR;
+
```
robertlipe, line 69: I get "this should never happen", but after we explictly call clamp() on res, is there any world there res can ever be a NaN?

That whole little block just seems weird. (I also know you didn't touch it, so it's not necessarily part of this PR.)

grtcirc.cc
```diff
@@ -301,20 +298,20 @@ double linedist(double lat1, double lon1,
  * Ref: http://mathworld.wolfram.com/RotationFormula.html
  */
 
-void linepart(double lat1, double lon1,
-              double lat2, double lon2,
-              double frac,
-              double* reslat, double* reslon)
+PositionDeg linepart(PositionRad pos1, PositionRad pos2, double frac)
```
robertlipe, line 301: Hooray for sensible calling conventions over outargs! That c-ism is overdue to die.

robertlipe (approved):
I like this quite a lot. The calling conventions in this area have  bugged me for years but not quite enough to go fix it. Thanx for running with it. I'll drop my noodle.

I think you've made it much harder (impossible?) to just call these with a couple of doubles where we don't know if they're degrees or radians and that's a great thing.

No required changes here, I think. Just things to mull over.

defs.h
tsteven4: No, they must not be explicit.  The "explicit(false)" is trying to tell clang-tidy to dummy up about
 >single-argument constructors must be marked explicit to avoid unintentional implicit conversions [google-explicit-constructor]

If they are explicit then the implicit conversions fail, e.g.:
```
arcdist.cc:64:41: error: could not convert ‘arcpt2->Waypoint::position()’ from ‘PositionDeg’ to ‘PositionRad’
   64 |           dist = gcdist(arcpt2->position(), waypointp->position());
      |                         ~~~~~~~~~~~~~~~~^~
      |                                         |
      |                                         PositionDeg
```

grtcirc.cc
tsteven4: Or for an old hipster:
https://github.com/GPSBabel/gpsbabel/blob/ba542e9dff6839e9723b7a436c69ff2b05d55181/src/core/nvector.cc#L234-L235

grtcirc.cc
tsteven4: I agree it's weird.  Evidently they are trying to avoid domain errors in asin, the clamps on zr, xr are harder to justify.

The clamp and EDOM code you cited came in at https://github.com/GPSBabel/gpsbabel/commit/644c09b5e "Check valid range for n before calling acos(n). Set errno always to zero."  Apr 29, 2006

grtcirc.cc
robertlipe: Indeed. I think that numeric_limits is the new spelling of those #defins from math.h. Same idea.

grtcirc.cc
tsteven4: One goal was to make the diff only be about the interface changes.  I did wonder about the optimizer, and if these copies would be better off as references but I think that kind of optimization has very little value in gpsbabel.  But old habits die hard, I still prefer to "* inv_pi" as opposed to "/ pi".

tsteven4:
> I think you've made it much harder (impossible?) to just call these with a couple of doubles where we don't know if they're degrees or radians and that's a great thing.

Its possible, but you have to work at it a bit:
https://github.com/GPSBabel/gpsbabel/blob/b099cb4eb1060fbe6fd65fbed0d3ad56f55e2770/waypt.cc#L203-L208

bend.cc
tsteven4: maybe so, but I left that out of scope.

grtcirc.cc
robertlipe: I'm with you. Maybe it's because I know that most FPUs can multiply by a constant reciprocal faster than they can divide.

tsteven4: Yes, that one was important in our real time embedded control ISR.  I even had an FPGA doing some of the high frequency division.

PR #1322, opened 2024-08-14
Modernize Humminbird internals

robertlipe:
- **Modernize Humminbirds internals**


humminbird.cc
```diff
@@ -247,24 +284,20 @@ HumminbirdBase::humminbird_read_wpt(gbfile* fin)
 
   auto* wpt = new Waypoint;
 
-  // Could probably find a way to eliminate the alloc/copy.
-  char* s = xstrndup(w.name, sizeof(w.name));
-  wpt->shortname = s;
-  xfree(s);
-
+  wpt->shortname = QStringFromRaw(sizeof(w.name), w.name);
```
tsteven4: Isn't this all we need here?  If so can we eliminate QStringFromRaw or replace its contents along these lines?

wpt->shortname = QByteArray(w.name, qstrnlen(w.name, sizeof(w.name)));

You could use QByteArray::fromRawData instead but I suggest saving that copy is noise and prone to misuse (source must continue to exist, result not '\0' terminated).

GPSBabelDeveloper:
I gave up on this last night. It was driving me crazy as you can tell from
my hit-and-miss attempts to debug it in CI.

I first missed from char* interfaces, but  later moved to those.
strnlen might be the ticket. My goal was to get rid of xstrndup as this is
about the only caller left. My final version last night went back to
using xstrndup
and I was still having luck.

I need to go back and be sure that I'm using the struct sizes I should be
using and such. I probaby got carried away with a yank and put and have
they wrong field size or something.

I'll pick it back up tonight after dr. appts. Of course, if that function
goes back to being a one-liner (and I'd forgotten about QByteArray having
"better" ctors here) then this function goes away. I don't know why I
swapped the arguments. Maybe I shouldn't have been programming at all last
night. I was able to get some failures with the debug malloc on MacOS, but
it consistently worked on MacOS and not any other system that we care
about, so it became quite the frustrating hunt.

And, yes, while it's my nature to count clock cycles from so many years of
doing systems software, I'll readily conced that humminbird's speed is
veery low on my list of reasons to not be sleeping.

I'll pick it back up. Thanx for the hint and the pep-talk that this
shouldn't be NEARLY as time-consuming as it was.

PR #1323, opened 2024-08-15
update shapelib to 1.6.1

tsteven4:


tsteven4:
This resolves the following build warnings on windows:
```
D:\a\gpsbabel\gpsbabel\shapelib\safileio.c(170): warning C4113: 'SAFile (__cdecl *)(const char *,const char *)' differs in parameter lists from 'SAFile (__cdecl *)(const char *,const char *,void *)'
D:\a\gpsbabel\gpsbabel\shapelib\safileio.c(171): warning C4113: 'int (__cdecl *)(const char *)' differs in parameter lists from 'int (__cdecl *)(const char *,void *)'
```

robertlipe:
Heh. They're adding big endian. (Odd, since nothing that has any business
running GPSBabel has been big endian in many years at this point...) In
another window, I'm removing some of ours...and rolling my eyes at most of
what's left.

tsteven4:
Like us (#1315) they are changing from run time endian detection to compile time endian detection.

PR #1326, opened 2024-08-16
explain humminbird geodetic to geocentric conversions.

tsteven4:
Document the humminbird geodetic to geocentric conversions. An ellipse was derived to match the data described in gpsbabel-code 2008-08-14. The related math is explained in terms of this derived parameter.

### DIFF
--- a/humminbird.cc
+++ b/humminbird.cc
@@ -167,28 +167,40 @@ struct HumminbirdBase::group_body_t {
 };
 
 
-/* Takes a latitude in degrees,
- * returns a latitude in degrees. */
+/* Takes a geodetic latitude in degrees,
+ * returns a geocentric latitude in degrees. */
 double
 HumminbirdBase::geodetic_to_geocentric_hwr(const double gd_lat)
 {
-  constexpr double cos_ae = 0.9966349016452;
-  constexpr double cos2_ae = cos_ae * cos_ae;
+  /* For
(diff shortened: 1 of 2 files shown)


tsteven4:
I have an idea for what the actual humminbird conversions between lat,lon and hwr file data are, as opposed to the approximate conversions from 2008 that we use today.  I suspect that the mercator projection formulas for the ellipsoid from "Snyder, J. P. Map projections — A working manual. Professional Paper 1395, U.S. Geological Survey, 1987" may be a match.  To investigate this I need a rosetta stone of appropriate known lat lon points and a humminbird created hwr file that contains them.  If there is a humminbird user out there who can create such a sample I would like to discuss a set of appropriate points.  @bjkowalski?

tsteven4:
The mercator projections on an ellipsoid are looking good.  We need the previously mentioned set of reference points to wrap this up.  If we move to the mercator projections for the ellipsoid then geodetic_to_geocentric_hwr and geocentric_to_geodetic_hwr which were explained in this PR go AWAY.

A more modern reference is:

 "The Mercator Projections", Peter Osborne, 2013
https://zenodo.org/records/35392

tsteven4:
I'm still interested in a set of crafted set of reference points from a humminbird user, but I am killing this PR and will incorporate any resulting changes in a new PR.

PR #1327, opened 2024-08-19
Humminbirdhash

tsteven4:
- refactor humminbird handling of wpt numbers.  This fixed some subtle bugs.  

  1. The map was shared and not cleared allowing communication between invocations of the humminbird reader & writer (reader->reader, reader->writer, writer->writer, etc.)
  2. When reading, waypoints that were dropped due to their status were still registered in the map.  If they happened to referred to by a route then deallocated memory would be accessed.
  3. humminbird_write_waypt_wrapper checked the map using the [] operator, which causes a silent insertion.  However, in this case the silently inserted entry was overwritten a few lines later with the correct value.

  It also allows gb_int2ptr and gb_ptr2int to be retired.

- convert some macros to constexpr.
- consistently use strncpy to move strings to humminbird char arrays.`
- fix clang-diagnostic-missing-field-initializers

### DIFF
--- a/humminbird.cc
+++ b/humminbird.cc
@@ -21,13 +21,13 @@
 
 #include "humminbird.h"
 
-#include <QMap>                 // for QMap
+#include <QHash>                // for QHash
 #include <Qt>                   // for CaseInsensitive
 #include <QtGlobal>             // for qRound
 
 #include <cmath>                // for atan, tan, log, sinh
-#include <cstdio>               // for snprintf, SEEK_SET
-#include <cstring>              // for strncpy, memcpy, memset
+#include <cstdio>               // for SEEK_SET
+#include <cstring>              // for strncpy
 #include <numbers>              // for inv_pi, pi
 
 #include "defs.h"               // for Waypoint, be_read32, be_read16, be_write32, fatal, be_write16, route_head, track_add_wpt
@@ -61,10 +61,6 @@ Still, they're useful in the code as a plain signature.
 #define WPT_MAGIC		0x02020024L
 #define WPT_MAGIC2		0x02030024L // New for 2013.  No visible diff?!
 #define RTE_MAGIC		0x03030088L
-
-#define EAST_SCALE		20038297.0 /* this is i1924_equ_axis*pi */
-#define i1924_equ_axis		6378388.0
-#define i1924_polar_axis	6356911.946
 
 #define BAD_CHARS		"\r\n\t"
 
@@ -172,8 +168,6 @@ struct HumminbirdBase::group_body_t {
 double
 HumminbirdBase::geodetic_to_geocentric_hwr(const double gd_lat)
 {
-  constexpr double cos_ae = 0.9966349016452;
-  constexpr double cos2_ae = cos_ae * cos_ae;
   const double gdr = gd_lat * std::numbers::pi / 180.0;
 
   return atan(cos2_ae * tan(gdr)) * 180.0 * std::numbers::inv_pi;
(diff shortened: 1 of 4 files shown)


humminbird.cc
```diff
@@ -229,7 +223,7 @@ HumminbirdBase::humminbird_rd_deinit() const
 void
 HumminbirdBase::humminbird_read_wpt(gbfile* fin)
 {
-  humminbird_waypt_t w{0};
+  humminbird_waypt_t w{};
```
robertlipe, line 226: Does default contruction on a struct (of POD types without initializers) really DTRT here?

I thought it was random noise in these structures that caused me so much grief last week on things that weren't MacOS.

robertlipe (approved):
Changes look great. We've spent a lot of brain cells on a format that's really not that strategic to us lately. :-|

humminbird.cc
tsteven4: This was a resharper suggested fix I agree with, although the previous code worked fine.  It does DTRT.

In the case you cite I believe the new code uses [value initialization](https://en.cppreference.com/w/cpp/language/value_initialization).  Although in this case as the note says [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization) is performed.  Thus I think the difference is {0} has an explicit initializer for num, and the rest of the fields are implicitly initialized, while with {} all fields are implicitly initialized.

The warning "clang-diagnostic-missing-field-initializers" was telling us there were fields that weren't explicitly initialized and at least one field that was explicitly initialized.  The warning does not apply to an empty initializer.

There certainly were test issues, and similar issues that didn't have test coverage, related to uninitialized data.  All the xcalloc (not xmalloc) calls are a hint that someone in the past thought those needed to be initialized.  You applied the fix in some additional cases, but that isn't a bad thing [Always initialize an object](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es20-always-initialize-an-object).  It is too easy to miss a field in something like humminbird_read_wpt.  The memset/memcpy to strncpy change is relying on that for name[19].

GPSBabelDeveloper:
Interesting. I was using CLion and the version that shipped last week,
after I did this work, has default clang warnings WAY too hyper for me.
I've been aggressively hushing many of the silly ones.

No, I don't think I'll be accepting trailing return types on every method
we define. Just No.

So I got lucky and "fixed" the failure in CI by setting only the first
member, num, to zero instead of everything to zero. I remember seeing
discussion on the nuances between aggregate initialization and direct
initialization but now that I look carefully at examples like:

std::vector<int> vec0(5, 9); // 9, 9, 9, 9, 9
std::vector<int> vec1{5, 9}; // 5, 9
I just want to cry. Gross!
If you don't make those structs into "real" classes with initializers,
I will. This is a bomb waiting for us again.

...and I don't even LIKE humminbird. It was already hanging on by a thread.

tsteven4:
> So I got lucky and "fixed" the failure in CI by setting only the first member, num, to zero instead of everything to zero.

No, you had initialized all members.  You provided an explicit initializer for the first member, but the rest used implicit initializers.
>Implicitly initialized elements
For a non-union aggregate, each element that is not an explicitly initialized element is initialized as follows ...

As for the keywords struct vs. class they are nearly identical (https://en.cppreference.com/w/cpp/language/class)
>The keywords class and struct are identical except for the default [member access](https://en.cppreference.com/w/cpp/language/access) and the default [base class access](https://en.cppreference.com/w/cpp/language/derived_class)

The cppcoreguidelines [suggest](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-struct) using struct or class based on:
>Use class if the class has an invariant; use struct if the data members can vary independently

but generally we don't follow this suggestion.

>If you don't make those structs into "real" classes with initializers,
I will. This is a bomb waiting for us again.

I don't see a need to declare these using class instead of struct.  However, providing default brace initializers for all non-class members is something to consider.  It is more verbose in the struct/class definition, but every usage is initialized.  I generally don't provide a default initializer for fields that are classes, rely on the class itself to provide an appropriate default, but provide default  member initializers for non-class fields, e.g. "int x{};" or if I rely on a particular initial value "int y{22.33};"  For an example search for "Data Members" in kml.h.

tsteven4:
There has been recent interest in humminbird, so I don't think we should axe it.

PR #1328, opened 2024-08-19
clang-tidy-18 fixes performance-unnecessary-copy-initialization

tsteven4:


robertlipe:
LGTM
Interesting that's the only getter it singles out.

tsteven4:
we don't have many getters that return references.

PR #1334, opened 2024-09-05
modernize some dynamic memory usage.

tsteven4:


util.cc
```diff
@@ -572,8 +573,8 @@ rot13(const QString& s)
 QString
 convert_human_date_format(const char* human_datef)
 {
-  char* result = (char*) xcalloc((2*strlen(human_datef)) + 1, 1);
-  char* cout = result;
+  auto result = std::make_unique<char[]>((2*strlen(human_datef)) + 1);
```
robertlipe: Instead of all this pointer bashing and hoping we fit , should cout just go away and result nena qstring that's appended to.

we could do a reserve() if were worried about excessive reallocs - and I'm not because I don't think we'll realloc more than 2 or 3x In this loop since we can always do an .append() and the container classes tend to double when they DO realloc and datetime specifiers are just not infinitely long.)

It's less mechanical and a little beyond the original scope, but I think it'll be a healthy direction for this. 

This is a good example of a function that needs unit testing, but we just don't have a great framework for such things.

tsteven4: good catch.  The coverage is better than I would have guessed, https://app.codacy.com/gh/GPSBabel/gpsbabel/files/124389575662/coverage

robertlipe (approved):
I like it. Thanx.

I suspect the coverage on those functions is good because they weren't arbitrary "turn everything into everything" but instead "turn what we used into what we need" handlers.

PR #1342, opened 2024-09-23
String handling in gbfile

tsteven4:
Keep the file name and module name in QStrings, converting to 8 bit on use when necessary.  This eliminates usage of xstrdup by gbfile.

supply default initializers for gbfile, and use new/delete instead of xcalloc/xfree.

retire char* xstrdup(const QString& s);.


robertlipe:
Nice. xstrdup() is almost gone!

PR #1343, opened 2024-09-25
catch gdb debug up with gbfile changes.

tsteven4:


robertlipe:
Nice.

I've meant for a long time to let a format register a modulename
(eliminating the MYNAME token pasting trick) into the core and then let the
streaming Warning variants issue that as part of their work.

You've really never bonded with Warning() and friends. Is that just because
touching everything is a PITA (it is) or do you have a reason to not like
them? I'll admit that strstream can be kind of ugly for precise numeric
formats, but most of our use is debug, so it doesn't much matter if we have
matching leading zero fills and such.

Maybe we're close enough to being able to use <fmt> to reduce some of that
chatty << stuff.

Just musing aloud about potential upcoming projects if I have a bit...

PR #1344, opened 2024-09-25
use container for unicsv fields def.

tsteven4:
and cleanup an extraneous break from #1164.

### DIFF
--- a/unicsv.cc
+++ b/unicsv.cc
@@ -21,6 +21,7 @@
 
 #include "unicsv.h"
 
+#include <algorithm>               // for find_if
 #include <cmath>                   // for fabs, lround
 #include <cstdio>                  // for NULL, sscanf
 #include <ctime>                   // for tm
@@ -59,7 +60,7 @@
  * ! Please use always underscores in field names !
  * we check a second time after replacing underscores with spaces
  */
-const UnicsvFormat::field_t UnicsvFormat::fields_def[] = {
+const QVector<UnicsvFormat::field_t> UnicsvFormat::fields_def = {
   /* unhandled columns */
   { "index",	fld_terminator, kStrAny },
   { "no",		fld_terminator, kStrEqual },
@@ -160,8 +161,7 @@ const UnicsvFormat::field_t UnicsvFormat::fields_def[] = {
   { "found",	fld_gc_last_found, kStrAny },
   { "placer_id",	fld_gc_placer_id, kStrAny },
   { "placer",	fld_gc_placer, kStrAny },
-  { "hint",	fld_gc_hint, kStrAny },
-  { nullptr,		fld_terminator, 0 }
+  { "hint",	fld_gc_hint, kStrAny }
 };
 /* helpers */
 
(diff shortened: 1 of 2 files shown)


robertlipe:
TIL you can std::find_if in a QVector. Unexpected.

We have a bunch of places that do these short table linear searches. Just
recently I read/watched an article how at the modern CPU level, CPUs are so
crazy good at unrolling, speculating, and SIMD-ifying things like this that
the old school loops were just really hard to beat by being clever.

They've bugged me slightly through the years, yet they've never shown up in
an interesting place (except huge geocaching GPX file reads) in any
profiling I've ever done..

Maybe the context was strlen() or how terrible zero-terminated strings were
thought to be, but how strlen ist almost free on recent memory
systems/cpus, which isn't intuitive at all. The punchline was that what we
thought was good was terrible and vice versa. I remember finding it
profoundly interesting, but I can't remember the specifics of what it was
talking about, which makes it profoundly unhelpful to apply. :-( It seems
like the writer was, like, the author of libc++ from th clang team or
someone equally similarly authoritative. Maybe it was Chandler Carruth...

tsteven4:
> TIL you can std::find_if in a QVector. Unexpected.

https://doc.qt.io/qt-6/containers.html#qt-containers-and-std-algorithms

GPSBabelDeveloper:
Cool. That phrase in the doc is new in Qt6.2 LTS; it wasn't in 5.12LTS.
Maybe equivalent words are there and I've just failed to notice them.

There must be some very clever ABI contracts made in the
implementation...or maybe the Qt guys just looked at the implementation and
knocked them off in the name of compatibility. It doesn't seem like that
came accidentally or easily.

Still, that's very cool and unexpected.

tsteven4:
It's not new, but the documentation is expanded.  The 5.12, 5.15 doc says 

> whereas the STL-style iterators are slightly more efficient and can be used together with Qt's and STL's [generic algorithms](https://doc.qt.io/qt-5/qtalgorithms.html).

> STL-style iterators have been available since the release of Qt 2.0. They are compatible with Qt's and STL's [generic algorithms](https://doc.qt.io/qt-5/qtalgorithms.html) and are optimized for speed.

We make extensive use of std::sort, and a few cases of std::for_each.

robertlipe:
I've known about that section forever. (OK, I really only go back to Qt3.)
I always took that to mean "for (mystuff::const_iterator i  = list.begin();
i != list.end(); ++i)"  vs QT_FOR_EACH or the more java-ish "iterator that
takes" while(i.hasNext()) something = i.next() approach.  I always just
hated typing the STL iterators.

Or maybe I've totally lost my mind and never had any idea what any of this
meant - even though I'm pretty sure that at least a few of those
std::for_each and std::sorts are my own. I guess the more general case of
mixing QContainers and std::containers willy-nilly just hadn't dawned upon
me.

I'll take
for (thing : things)
over
for (mystuff::const_iterator i  = list.begin(); i != list.end(); ++i)

every time when I'm typing it now, even if I'm walking a boring ole
std::array.  It really puts up bumper rails and makes it really hard to
access something outside of things[] via an off-by-one or fencepost or such.


I'm still not fully on board with https://safecpp.org/P3390R0.html yet,
though. Maybe I'll warm up to it more over time and once we see how it goes
down with the industry. I just don't inherently love things like

P5:     int y = 2; <https://safecpp.org/P3390R0.html#cb45-13>P6:
<loan R2> = ^y;  // ^y is loan 'R2
<https://safecpp.org/P3390R0.html#cb45-14>P7:     ref = <loan R2>; //
'R2 : 'R0 @ P7 <https://safecpp.org/P3390R0.html#cb45-15>P8:
f(*ref);
P9:     drop y
I'm even down from the ledge where I'm willing to burn double-time
compiles and 3-5% runtimes just to
 never ever hear about someone in the industry (especially me)
fumbling bounds violation and
 lifetimes again. I think both Sutter and Carbon are off wandering
around int the weeds lost, though.
I did dig Jen's points that were highlighted in
https://www.theregister.com/2024/09/20/cisa_software_cybercrime_villains/
. I DO think that with even a few minor changes in the
 defaults of C++ we could banish whole classes of bugs. The C culture
is to ignore changes in the
 language, so I don't know what you can do there culturally, but these
aren't my problems to solve.

PR #1345, opened 2024-09-26
use containers for xml_init lists.

tsteven4:


tsteven4:
I resisted std::ranges::for_each although it would have worked fine with these const lists.  With non-const shared lists I think there is a possibility of std::ranges::for_each causing a deep copy.

robertlipe:
It seems weird that we have  std::for_each and a std::ranges::for_each (and
find_if and several other things) but here we are.

We could actually keep them all in flash and use them in place, not making
a merged list at all and then just do a scan of the common tags and then a
scan of the format_specific tags.

OTOH, it's probably been 15 years since anyone has run out of memory doing
anything sane with GPSBabel or noticed loops like this in a profile, so I'm
not sure I care about any of this.

Sometimes it's OK to just walk away. :-)

tsteven4:
We are putting them in a QHash.

PR #1347, opened 2024-10-10
further qstringification of osm

tsteven4:
as well as refactoring arrays -> containers.

This eliminates c strings except for numerous c string literals.  The handling of string literals wrt QStrings is fluid, what was best in Qt5 is different from Qt 6.2 which is different from Qt 6.5.  For example in Qt 6.5 the QXmlStreamWriter  interface changed from QString to QAnyStringView.  In any event I am leaving that alone for now.


osm.cc
```diff
@@ -376,24 +371,24 @@ void
 OsmFormat::osm_features_init()
 {
   /* the first of osm_features is a place holder */
-  for (int i = 1; osm_features[i]; ++i) {
+  for (int i = 1; i < osm_features.size(); ++i) {
```
robertlipe, line 374: In either 23 or 26 we could write something like
For (int I = 0, const auto fest& : osm_features) {
   Blah(feat, i++)
}
... And have I go out of scope at the right time. There was sadness that the increment couldn't be in the loop control. I don't remember if that ball is still in play or if that's just the way it is, but it's still a solid minor step up. 

But this isn't close for us, I don't think. 

This is really nice PR!

robertlipe (approved):

PR #1349, opened 2024-10-13
Revert "fix an annoying Wstringop-truncation warning. (#1348)"

tsteven4:
This reverts commit a4dec0acfbeb6d93a9a73e3673da983bfdd7511b.

The compiler is smart enough to figure out there isn't a Wformat-truncation issue, and the changes didn't related to the Wstringop-truncation  warning.

### DIFF
--- a/jeeps/gpsapp.cc
+++ b/jeeps/gpsapp.cc
@@ -88,7 +88,7 @@ static void   GPS_D152_Send(UC* data, GPS_PWay way, int32_t* len);
 static void   GPS_D154_Send(UC* data, GPS_PWay way, int32_t* len);
 static void   GPS_D155_Send(UC* data, GPS_PWay way, int32_t* len);
 
-static void   GPS_D120_Get(US cat_num, char*s);
+static void   GPS_D120_Get(int cat_num, char*s);
 
 static void   GPS_D200_Get(GPS_PWay* way, const UC* s);
 static void   GPS_D201_Get(GPS_PWay* way, UC* s);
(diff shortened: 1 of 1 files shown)


robertlipe:
These data structures in the geriatric garmins are just a snake pit.

They're a mixture of
C strings with zero termination.
Fixed length buffers with spaces to pad them.
Fixed length buffers with nulls to pad them.
Buffers that crash the device if a space is present (that I think are zero
terminated)

...and not enough institutional knowledge to know which is which.  It looks
like D120 isn't terribly ancient (w/in the last 20 years), so maybe I have
some device that could exercise this. Maybe.

So even beyond the horror of strncpy's "we'll terminate it, except when we
don't,"  there's the issue of how this data is handled on the device
if/when it appears over the wire. It's quite likely that we're doing things
that intentionally don't result in zstrings. Those should probably be
memfoo instead of strfoo, but here we are.

The jeeps serial code is probably the hairiest stuff left in this project
that I'm largely just afraid to touch on good faith alone. "Obvious"
changes rarely are. Years ago, I wanted to rip it all up and replace it
with a table-driven approach that described all the packets instead of one
function per packet, but even then rounding up all the needed hardware
wasn't practical, so I faced either risking breaking things I didn't have
(and didn't really care much about) or just drawing a hard line and saying,
"we support only the following N models."

All these years later, I'm wondering how many Garmins are still _really_ in
use that aren't USB/GPX devices, making most of this code moot anyway.

tsteven4:
Yes, the garmin string code is located very close to [Mussentuchit Wash](https://www.google.com/maps/place/Mussentuchit+Wash/@38.6938653,-111.1843349,15z/data=!3m1!4b1!4m6!3m5!1s0x874bd49c93c592e1:0x21ef659e07045009!8m2!3d38.6938658!4d-111.1843349!16s%2Fg%2F1vr1x30g!5m1!1e4?entry=ttu&g_ep=EgoyMDI0MTAwOS4wIKXMDSoASAFQAw%3D%3D)

PR #1352, opened 2024-10-18
introduce OptionString,

tsteven4:
a replacement for OptionCString. Formats and filters can be manually converted, with the potential to eliminate c character string usage.  In contrast to OptionCString, OptionString can be implicitly cast to a bool or to a const QString&, but not a const char*.

OptionCString::get() usages can often use an implicit cast from OptionString instead, although the usage of get() is still valid.

By using QptionString::get() any of the QString functions that accept a const QString& can be used such as QString::toInt(),, QString::toDouble(), ...


robertlipe (approved):
Seven minutes from merge to turndown is some kind fo a new record. :-)

Darned if that's not a lot cleaner, though.

Since OCS didn't actually go away, I assume there's more. Go for it!

tsteven4:
> Since OCS didn't actually go away, I assume there's more.

There are random usages of c strings, for example xml_parse_time, KmlFormat::kml_lookup_gc_icon, ... .   One intent here is to enable the removal of c strings required by option handling.  That alone requires manual intervention in every format and filter that use non ARGTYPE_BOOL options.  At least we are set up to do this one format/filter at a time as energy allows.  I was encouraged that this already is reducing our dependency on <cctype>, <cstdlib>, <cstring>.  And I agree, cleaner code is a nice benefit as well.

PacketFiend:
Hey fellas,

Wondering if anything more needs to be done with the IGC bits because of this? It does make pretty heavy use of options...

tsteven4:
@PacketFiend I have related changes for igc on deck.

tsteven4:
At this point OptionString::toInt and ::toDouble will ***fatal*** if there is a conversion error, and a non-null status argument was not passed.  This is by far the most common use case.  This might leave a ff_type_serial device in an intermediate state.  This risk can be avoided by adding strict validation of ARGTYPE_INT and ARGTYPE_FLOAT in Vecs.  This was accomplished by adding characterization for options of ARGTYPE_INT and ARGTYPE_FLOAT.  For integers we need to know the base (automatic, 10, 16).  For integers and floats we need to know if trailing data is allowed, typically used to allow units to be entered along with the value.

I plan to add OptionInt and OptionFloat corresponding to ARGTYPE_INT and ARGTYPE_FLOAT, and add methods to those that retrieve the value converted by Vecs instead of converting again.

The overloads of OptionString::toInt and ::toDouble allow the unusual use cases to be identified by a tool such as JetBrains ReSharper.  This was used to help validate the proper introduction of the characterization bits in argtype.

One wrinkle is parse_distance and parse_speed which convert a QString, not an Option*.  Callers may or may not get that string from an Option*.  These cases have all been examined, the characterization bit to allow trailing data set in argtype, and those that were ARGTYPE_STRING have been changed to ARGTYPE_FLOAT.

tsteven4:
> I plan to add OptionInt and OptionFloat corresponding to ARGTYPE_INT and ARGTYPE_FLOAT, and add methods to those that retrieve the value converted by Vecs instead of converting again.

This has been done.  All the conversions of ARGTYPE_INT and ARGTYPE_FLOAT are done in vecs, although they may be done again in parse_distance or parse_speed.

This is ready for some serious review.

PR #1356, opened 2024-10-22
Qtize rgbcolors.

tsteven4:


rgbcolors.cc
```diff
@@ -219,31 +205,28 @@ static int HexByte(const char* hex)
...
+            (match.captured("green").toInt(nullptr, 16) << 8) +
+            (match.captured("blue").toInt(nullptr, 16) << 16));
   }
 
-  for (auto i : color_table) {
-    if (0 == case_ignore_strcmp(cname, i.cn)) {
-      return (i.b << 16) +
-             (i.g << 8) +
-              i.r;
-    }
```
robertlipe, line 244: Do you share my suspicion that contains().does 80% of the work that value() does? 

Nice cleanup!

tsteven4: yes, sometimes supplying a default value with QHash::value(const Key &key, const T &defaultValue) alleviates that concern, but not in this case.

PR #1367, opened 2024-11-01
Use QtLogging Message Pattern to replace MYNAME

tsteven4:
All the action is in main.cc and fatal.cc.  The other 78 files are deletions of MYNAME and descendants.  The bulk of the changes were first done by the following script, the remainder by hand.
```
#!/bin/bash -x
for file in *.cc
do
  sed -i -e 's/MYNAME *"[:-] */"/' $file
  sed -i -e 's/<< *MYNAME *<< *":\? */<< "/' $file
  sed -i -e 's/( *MYNAME *" *:\? */("/' $file
  sed -i -e 's/<< MYNAME "-/<< "/' $file
done
```

### DIFF
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -22,7 +22,7 @@
 #include "arcdist.h"
 
 #include <cmath>                  // for round
-#include <cstdio>                 // for printf, sscanf
+#include <cstdio>                 // for sscanf
 #include <tuple>                  // for tie, tuple
 
 #include <QByteArray>             // for QByteArray
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -198,7 +198,7 @@ void ArcDistanceFilter::process()
   }
   del_marked_wpts();
   if (global_opts.verbose_status > 0) {
-    printf("arc: %u waypoint(s) removed.\n", removed);
+    info("%u waypoint(s) removed.\n", removed);
   }
 }
 
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -190,7 +190,7 @@ void ArcDistanceFilter::process()
         }
         if (global_opts.debug_level >= 1) {
           gbWarning("Including waypoint %s at dist:%f lat:%f lon:%f\n",
-                  qPrintable(wp->shortname), ed->distance, wp->latitude, wp->longitude);
+                  gbLogCStr(wp->shortname), ed->distance, wp->latitude, wp->longitude);
         }
       }
       delete ed;
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -37,7 +37,6 @@
 
 
 #if FILTERS_ENABLED
-#define MYNAME "Arc filter"
 
 #define BADVAL 999999
 
@@ -112,7 +111,7 @@ void ArcDistanceFilter::process()
     QString line;
 
     gpsbabel::TextStream stream;
-    stream.open(arcfileopt, QIODevice::ReadOnly, MYNAME);
+    stream.open(arcfileopt, QIODevice::ReadOnly);
 
     auto* arcpt1 = new Waypoint;
     auto* arcpt2 = new Waypoint;
@@ -134,7 +133,7 @@ void ArcDistanceFilter::process()
       int argsfound = sscanf(CSTR(line), "%lf %lf", &arcpt2->latitude, &arcpt2->longitude);
 
       if ((argsfound != 2) && (line.trimmed().size() > 0)) {
-        warning(MYNAME ": Warning: Arc file contains unusable vertex on line %d.\n", fileline);
+        warning("Warning: Arc file contains unusable vertex on line %d.\n", fileline);
       } else {
         Waypoint* arcpttmp = arcpt1;
         arcdist_arc_disp_wpt_cb(arcpt2);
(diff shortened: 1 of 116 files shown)


tsteven4:
The bulk edit changed printf statements which don't go through the Qt Logging system.  These need adjustment to use the message pattern.

tsteven4:
There is another issue with newlines.  Traditionally we don't implicitly insert them, e.g. with warning(), but at the same time we expect qDebug() << ... to insert them.  gdb in particular expects only explicitly inserted newlines in the DBG printf statements which we want to capture in logging.

tsteven4:
Another issue is the use of printf without a trailing newline.  These are sometimes used to accumulate character output which will  eventually be terminated with a newline.  These newline-less printf statements cannot be logged directly, instead we must transform the character output into lines and log those.  DebugLog is a new class to do that, demonstrated in globalsat_sport.

tsteven4:
gbser and jeeps logging aren't included.

tsteven4:
Making allowances for missing/alternate MYNAME content, the log of testo with "-D 9" added to gpsbabel and the auxiliary functions disabled (\*compare\*, \*check) we have only a few mis-compares (out of 1,251,735 lines).  These are all related to the alternative representation of the g conversion specifier.  The dropped trailing zeros seem to be a bug in QString::vasprintf.  The sign of nan may be implementation defined.  This seem unimportant to me.
```
$ diff log4x log7x
4971c4971
< exif: offs 0x000000CA: ifd=2 id=0x829A t=0x0005 c=   1 s=   8 o=0x00000242 +0.00312500(1/320)
---
> exif: offs 0x000000CA: ifd=2 id=0x829A t=0x0005 c=   1 s=   8 o=0x00000242 +0.003125(1/320)
5070c5070
< exif: offs 0x000000E6: ifd=2 id=0x829A t=0x0005 c=   1 s=   8 o=0x00000258 +0.00800000(1/125)
---
> exif: offs 0x000000E6: ifd=2 id=0x829A t=0x0005 c=   1 s=   8 o=0x00000258 +0.008000(1/125)
5138c5138
< exif: offs 0x000000CA: ifd=2 id=0x829A t=0x0005 c=   1 s=   8 o=0x00000242 +0.00312500(1/320)
---
> exif: offs 0x000000CA: ifd=2 id=0x829A t=0x0005 c=   1 s=   8 o=0x00000242 +0.003125(1/320)
5247c5247
< exif: offs 0x000001B4: ifd=2 id=0x9102 t=0x0005 c=   1 s=   8 o=0x0000026C -nan(0/0)
---
> exif: offs 0x000001B4: ifd=2 id=0x9102 t=0x0005 c=   1 s=   8 o=0x0000026C +nan(0/0)
5253c5253
< exif: offs 0x000001FC: ifd=2 id=0x9206 t=0x0005 c=   1 s=   8 o=0x0000029C -nan(0/0)
---
> exif: offs 0x000001FC: ifd=2 id=0x9206 t=0x0005 c=   1 s=   8 o=0x0000029C +nan(0/0)
```

tsteven4:
The Qt Logging system is line based with implied newlines, while our traditional logging is printf based with explicit newlines.

Sometimes we use multiple printf commands to assemble a line of log output.

Another wrinkle is live status like waypt_status_disp with it's '\r' characters and mtk_logger with it's erase wheel '\b' character and mtk_read's Reading status update with '\b's.  waypt_status_disp can be exercised with a modified dg100.test, but exercising the above portions of mtk requires a device.

To deal with this I created a LegacyLogMessageHandler that we use with our traditional printf based logging.  We use the Qt [defaultMessageHandle](https://github.com/qt/qtbase/blob/d3b5353380797f3b67599ccebc5dc916057681e5/src/corelib/global/qlogging.cpp#L1742-L1760)r for qDebug, qInfo, qWarning, ... and our src/core/logging based messages.  Our legacy handler only formats a message with qFormatLogMessage (which inserts the "id: ", the replacement for MYNAME) on new log lines and outputs any messages regardless of newline content immediately.  The defaultMessageHandler provide by Qt always formats a message with qFormatLogMessage and always uses an implied newline.

tsteven4:
Our legacy logger is currently implemented with QString::vasprintf, as is Qt printf style logging.  If [PIGS_FLY](https://github.com/GPSBabel/gpsbabel/blob/400dcfb7e62ebb7198e99e0b8beb7481f24ca84c/fatal.cc#L34-L47) we can switch back to character string conversion.  QString::vasprintf expects utf-8 encoded format and character strings arguments, while xvasprintf would want local8Bit encoding.  The encoding is switchable be redefining [gbLogCStr ](https://github.com/GPSBabel/gpsbabel/blob/400dcfb7e62ebb7198e99e0b8beb7481f24ca84c/defs.h#L54)

tsteven4:
The "[id]: " message insertion is set up by [setMessagePattern](https://github.com/GPSBabel/gpsbabel/blob/400dcfb7e62ebb7198e99e0b8beb7481f24ca84c/main.cc#L197-L204) in main.  As main runs formats and filters it updates the message pattern.

robertlipe (approved):
Wow! What a slog.

When I first started kicking around the idea of moving to c++ and having formats a s classes, MYNAME was one of the justifications  for doign so. I'm happy you sent this to the grave it deserves.

Thanx!

tsteven4:
One thing not converted is printf usage in gbser*  or jeeps.

PR #1374, opened 2024-11-09
delete obsolete files.

tsteven4:
we have std::optional instead of optionaldouble.h
we have MakeShort class now instead of shortname.cc

we should consider removing src/core/ziparchive.[ch], which haven't made it into production since their introduction in 2015.


robertlipe:
Lgtm

Please feel free to nuke zip, too. I've lost interest in doing Garmin ggz
and kmz. If you don't, I can.

Thanx

PR #1376, opened 2024-11-10
Attestation

tsteven4:


robertlipe:
I know very little about this, but shouldn't we be signing only the
executables in releases and not. Nifjtlies or build artifacts?

tsteven4:
I'm trying to generate attestations for the Continuous Release installers, not the build artifacts.  Unfortunately this stuff only runs after the PR is merged into master.  When we do a release it would be nice if we had attestations for those builds, but that will mean absolutely no post build changes are allowed, and we will need to make sure the release images have attestations.  At this point I just aiming for the Continuous releases.

GPSBabelDeveloper:
I defer to your wisdom.

Perhaps we move to a model of a branch for release (candidate) builds and
attest those and not all the nighrlies and presubmits. There was some other
case that would simplify.

But, again, I defer.  I'm just saying we don't always have to stick with
The Way it's Always Been.

PR #1385, opened 2024-11-17
Refactor GUI map preview

tsteven4:
Among other cleanups this enforces storage of which map items are visible to GMapDlg::model_.  Duplication of this information, for example in GpxItem::visible, has been eliminated.  This allows simplification as we can often use common routines passing a parent or child of the model instead of separate routine for wayoints/tracks/routes. It also improves maintainability and correctness as the possibility of disagreement is eliminated (which could occur previously).   It allows GMapDlg::gpx_ to be const, which ensures we aren't operating on copies of information, as we were in GMapDialog::[show|hide[All[Waypoints|Track|Routes]() (although the code worked anyway, the operation on the temporary copy was just obfuscation.  Overall we are down 347 LOCs.

This started as remove of foreach.  That is readily achievable in the GUI, but fraught with peril in the CLI due to routines that operate on global lists that you may be iterating over.

I still need to turn debug off before merging.

### DIFF
--- a/gui/gmapdlg.cc
+++ b/gui/gmapdlg.cc
@@ -214,6 +214,27 @@ GMapDialog::GMapDialog(QWidget* parent, const Gpx& mapData, QPlainTextEdit* te):
 }
 
 //------------------------------------------------------------------------
+void GMapDialog::trace(const QString& label, const QStandardItem* it)
+{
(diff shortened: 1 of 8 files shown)


GPSBabelDeveloper:
Wow. Thanks!

PR #1387, opened 2024-11-20
cleanup GUI special member functions

tsteven4:
These tidy checks had a large impact:
cppcoreguidelines-prefer-member-initializer
modernize-use-default-member-init
modernize-use-equals-default

Some instances of other tidy issues were resolved:
clang-analyzer-optin.cplusplus.VirtualCall
cppcoreguidelines-special-member-functions
readability-inconsistent-declaration-parameter-name
readability-named-parameter
and others.

Overall we have 1334 fewer tidy checks on our code base, a
reduction of 15% of the total!

The user defined dtor for class SettingGroup was eliminated
by using std::vector<std::unique_ptr<VarSetting>> instead of a
QList of raw pointers.
The user defined dtor for class FilterWidget was eliminated
by using std::vector<std::unique_ptr<FilterOption>> instead of a
QList of raw pointers.

PR #1389, opened 2024-11-25
update dev archive patch for recent PR merge.

tsteven4:
#1387 invalidated previous patch.


robertlipe:
Leave that 'on', please.

tsteven4:
1. I think that would cause trouble.  the dev docker image is a pre-release image with a version # like `GPSBabel 1.9.0+b8068b91` . The intent of that string is to say this was build from 1.9.0 with additional changes from git commit b8068b91 (as is used in other projects).  The "shorter string wins" rule in upgrade.cc is going to suggest an upgrade to 1.9.0.  
2. Not only is an upgrade to 1.9.0 a downgrade, the docker image is only upgradeable by me pushing a new image to dockerhub, after which the user can "docker pull".

PR #1400, opened 2025-01-22
Linuxarm64

tsteven4:
Add CI job to build and test on Linux arm64.

This ran into rounding errors in our regression tests with the igc format and valgrind.  The tests all passed without valgrind, but many of the igc tests failed reference comparisons when run under valgrind.  This failures occur on arm64 with valgrind 3.22.0 and 3.23.0.  They do not occur with valgrind 3.24.0.  See https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/2095501

To work around this I created a ppa with valgrind 3.24.0: https://launchpad.net/~tsteven4/+archive/ubuntu/valgrind-backports

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -92,9 +92,8 @@ jobs:
         brew install ninja
         brew install docbook docbook-xsl fop gnu-sed
         # brew install is taking forever on macos-11, skip jing-trang and all it's dependencies.
-        if [ "${{ matrix.os }}" != "macos-11" ]; then
-          brew install jing-trang
-        fi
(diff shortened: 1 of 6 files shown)


robertlipe:
What a tough adventure. Thanx for solving a murder mystery where the
detective was the murderer.

I know we sometimes get into legit edge cases where printf, <<, to_chars,
and such round in different allowable ways, but I'm always surprised at
what we get away with out in those sub-mm bits where we get exact matches
regardless if our double math is done on a 64 or 80 bit  number and all the
different CPU types that we've been on on through the years. For example, I
never expected to just byte-swap a Dragonball double and have it
numerically preserveded on a contemporary system, but we've done
effectively that for quite some time now.

PR #1406, opened 2025-01-26
Remove penultimate call to xstrdup in project from mtk.

robertlipe:


tsteven4:
It's actually the antepenultimate call, don't forget jeeps/gpsusbcommon.cc

robertlipe:
> It's actually the antepenultimate call, don't forget jeeps/gpsusbcommon.cc

Busted.

That last call is in a pretty awkward place. I'm not sure that anyone but me has ever executed that code anyway, and it's deep in chain that leaks anyway, but it's a device path, so it doesn't make sense to care. So I may not fix that one "well." It's pretty clear we can soon remove xstrdup to remove it from temptation's path.

We're down to a single caller of xfopen, so maybe we just move it into gbfile. xrealloc and xcalloc are also only used there and could probably turn into a QByteArray pretty easily.

About our biggest use of xfoo any longer is, unsurprisingly, xmalloc and it tends to be in formats that we try not to touch.

Looking at xstrtoi(), I just don't know what we were thinking. Nothing looks for ERANGE. That can probably turn into std::fromChars() with only a dash of thinking about boundary cases.

It's mostly busy work that brings only risk and little value but gives us some closure to a decade of whittling these down.

tsteven4:
xstrtoi was my extension of strtol/strtoll for the integer type.  I was probably motivated by some tool complaining about loss of precision with strtol.

tsteven4:
xstrtoi was used much more extensively before I refactored format/filter option handling.  Those uses have been consolidated in parse_integer which uses stoi.

PR #1408, opened 2025-02-11
update Qt 6.8.1 -> 6.8.2

tsteven4:


robertlipe:
I love how low-drama this is these days...

tsteven4:
Actually there was a little drama.  CI failed to install Qt 6.8.2 on macos.  For some reason the macos release of Qt was delayed (or silently re-released?).  For other platforms it was available ~ 1/30/2025, but for macos it wasn't until 2/11/2025.  Add about a day for the checksums to be generated and thus make the release usable.

PR #1409, opened 2025-02-13
for Qstrings and QByteArrays prefer isEmpty() over size() and length()

tsteven4:
This applies the clang-tidy readability-container-size-empty check to QString and QByteArrays.

This check normally doesn't work because unlike QList QString and QByteArray don't have an empty() method.  By editing qstring.h and qbytearray.h an empty() method can be added.  When running clang-tidy use the export-fixes option as the fixes will use empty() instead of isEmpty().  Edit the fixes file changing the replacment text to use isEmpty() instead of empty(): 'g/ReplacementText/s/\\.empty()/.isEmpty()/g' and 'g/ReplacementText/s/->empty()/->isEmpty()/g', and then run clang-apply-replacements.

You may note some replacements where the original code didn't use size() or length(), but instead used equality comparisons to "" or QString().  These were generated by the readability-container-size-empty check, and should work fine.  I did compare the optionList generated by formatload and found it equivalent.

### DIFF
--- a/arcdist.cc
+++ b/arcdist.cc
@@ -132,7 +132,7 @@ void ArcDistanceFilter::process()
       arcpt2->latitude = arcpt2->longitude = BADVAL;
       int argsfound = sscanf(CSTR(line), "%lf %lf", &arcpt2->latitude, &arcpt2->longitude);
 
-      if ((argsfound != 2) && (line.trimmed().size() > 0)) {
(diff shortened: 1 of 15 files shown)


robertlipe:
Nice!

Some day, we should get that going on patches coming in so we don't
introduce regressions.

PR #1413, opened 2025-02-23
add cmake option: upgrade chk and stats reporting.

tsteven4:
Many linux distributions prefer not to have the upgrade check and statistics reporting.

Add a cmake options to disable them.  This removes gui/upgrade.* from the build and the related items from the menus.  By default the upgrade check and statistics reporting are ON as they have traditionally been.

cmake -DGPSBABEL_UPGRADE_CHECK=OFF ...


tsteven4:
One point of view is that users of apps distributed by packagers should check for the existence of an upgrade by the methods used by that system, e.g. apt-get, dnf, ...  These report if an updated package is available independent of an upstream release.

tsteven4:
@jspricke can you comment on the desirability of this?  I believe it could replace some of your patches.  I don't see a need for you to incorporate it until something else drives a change.

jspricke:
This would be great, thanks for proposing it.

vincele (commented):
This looks very useful, and the modifications sensible enough for my (untrained) eyes.

PR #1419, opened 2025-03-06
create preference privacy tab.

tsteven4:
1. create a privacy tab on the file menu preferences item. This includes the old preferences for
   - Check for newer version on start.
   - Anonymously report usage data. and new items for
   - Enable upgrade check menu item.
   - Enable map preview feature. Tool tips are added describing the related data and/or policies.
2. enable or disable the upgrade check menu item and the map preview feature based on the preferences.
3. Add tool tip for the map preview checkbox and the Help "Check for Upgrade" action that describes how to enable these items.
4. Only report the run count if reporting of usage data is allowed.

### DIFF
--- a/gui/notices.txt
+++ b/gui/notices.txt
@@ -1,0 +1,18 @@
+Privacy
+
+The command line interface program, gpsbabel.exe, will not transfer any information to other networked systems unless specifically requested by the user or the person installing or operating it.
+
+The graphical user interface program, GPSBabelFE.exe, may transfer information to other networked systems.
+  - If enabled, the automatic upgrade check transmits information about your installation and computer environment, such as the Installation ID, GPSBabel GUI and CLI versions, CPU architecture, operating system name and version, and language, to gpsbabel.org.
+  - If enabled, anonymous usage data is sent to gpsbabel.org. This helps us understand what features of the program are being used.
+  - If enabled and selected, the "Check for Upgrade" item in the "Help" menu sends the same information as the automatic upgrade check to gpsbabel.org.
+  - If enabled and selected, the "Preview in Google Maps" item in the "More Options" dialog uses the Google Maps API(s) per Google's Terms of Service (http://www.google.com/intl/en/policies/terms) and the Google Privacy Policy (http://www.google.com/policies/privacy).
+All of these items can be controlled through the Privacy tab reached through the Preferences item of the File menu.
+
+Third Party Content
+
+GPSBabel uses:
+  - libusb (https://libusb.info/)
+  - Qt (https://www.qt.io/download-open-source).
+  - Shapelib (http://shapelib.maptools.org/).
+  - zlib (https://zlib.net/)
--- a/gui/notices.txt
+++ b/gui/notices.txt
@@ -3,11 +3,11 @@ Privacy
 The command line interface program, gpsbabel.exe, will not transfer any information to other networked systems unless specifically requested by the user or the person installing or operating it.
 
 The graphical user interface program, GPSBabelFE.exe, may transfer information to other networked systems.
-  - If enabled, the automatic upgrade check transmits information about your installation and computer environment, such as the Installation ID, GPSBabel GUI and CLI versions, CPU architecture, operating system name and version, and language, to gpsbabel.org.
-  - If enabled, anonymous usage data is sent to gpsbabel.org. This helps us understand what features of the program are being used.
-  - If enabled and selected, the "Check for Upgrade" item in the "Help" menu sends the same information as the automatic upgrade check to gpsbabel.org.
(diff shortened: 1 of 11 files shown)


tsteven4:
I've added a screen to the windows installer about privacy and third party content (gui/notices.txt).

gui/notices.txt
```diff
@@ -0,0 +1,18 @@
+Privacy
+
+The command line interface program, gpsbabel.exe, will not transfer any information to other networked systems unless specifically requested by the user or the person installing or operating it.
+
+The graphical user interface program, GPSBabelFE.exe, may transfer information to other networked systems.
+  - If enabled, the automatic upgrade check transmits information about your installation and computer environment, such as the Installation ID, GPSBabel GUI and CLI versions, CPU architecture, operating system name and version, and language, to gpsbabel.org.
+  - If enabled, anonymous usage data is sent to gpsbabel.org. This helps us understand what features of the program are being used.
```
robertlipe: Do we need to spell out who "us" is and the benefit to the user/to us what each of these points are?

I'm not pitching that; it's a question after being in way too many bikeshedding meetings about sentences like these.

gui/notices.txt
```diff
@@ -0,0 +1,18 @@
...
+  - If enabled and selected, the "Check for Upgrade" item in the "Help" menu sends the same information as the automatic upgrade check to gpsbabel.org.
+  - If enabled and selected, the "Preview in Google Maps" item in the "More Options" dialog uses the Google Maps API(s) per Google's Terms of Service (http://www.google.com/intl/en/policies/terms) and the Google Privacy Policy (http://www.google.com/policies/privacy).
```
robertlipe: I think I understand the change in senses here, but I don't know how it could be only either enabled OR selected. This makes me wonder if these two checkboxes are in the wrong place and the should all just be in settings.

gui/notices.txt
robertlipe: Should we get ahead of concerns that we never ever transmit any location data from user files or devices? Well, except in the mapping case, but even then, it's not to us.

gui/preferences.cc
```diff
@@ -49,8 +49,14 @@ Preferences::Preferences(QWidget* parent, QList<Format>& formatList,
 
   ui_.startupCheck->setChecked(babelData_.startupVersionCheck_);
   ui_.reportStatisticsCheck->setChecked(babelData_.reportStatistics_);
+  ui_.upgradeMenuCheck->setChecked(babelData_.upgradeMenuEnabled_);
+  ui_.mapPreviewCheck->setChecked(babelData_.mapPreviewEnabled_);
   ui_.ignoreVersionMismatchCheck->setChecked(babelData_.ignoreVersionMismatch_);
 
+#ifdef DISABLE_MAPPREVIEW
```
robertlipe, line 56: Why do we optionally make these even less potentially usable by throwing it out after we've written it?

gui/upgrade.cc
```diff
@@ -124,9 +124,10 @@ UpgradeCheck::updateStatus UpgradeCheck::checkForUpgrade(
   args += QStringLiteral("&ugacc=%1").arg(babelData_.upgradeAccept_);
   args += QStringLiteral("&ugoff=%1").arg(babelData_.upgradeOffers_);
   args += QStringLiteral("&ugerr=%1").arg(babelData_.upgradeErrors_);
-  args += QStringLiteral("&rc=%1").arg(babelData_.runCount_);
 
   if (babelData_.reportStatistics_) {
+    args += QStringLiteral("&rc=%1").arg(babelData_.runCount_);
```
robertlipe, line 129: You know, I don't know if there are joins on any of this stuff that are going to flip out if any of these fields are missing. We dont' exactly have a way of reporting errors in this kind of thing back upstream in any useful way. It's not like there's any automated tests that have a mock database where we inject all the permutations of all these flags and ensure the middleware (the code that handles the post and stores them) will handle it.

Most of the time I use runcount it's in hand-crafted queries to solve specic problems, though, so I guess even if that post (or any of these posts) just totally crashes, we're not out anything more than we're giving up anyway.

robertlipe (approved):
I'm pretty clammy to this whole thing, but the code is, of course, perfect. Go for it.

gui/preferences.cc
tsteven4: When I introduced this it was to enable building without QtWebEngine, which lagged the release of other Qt6 modules.

gui/notices.txt
tsteven4: I'll change to:
enabled in the "Privacy" tab and selected in the "Help" menu.
&
enabled in the "Privacy" tab and checked in the "More Options" dialog.

tsteven4: sort of, see commit.

gui/notices.txt
tsteven4: us -> the developers at gpsbabel.org

tsteven4: I added "This never includes any data from user files or devices."  (dropping your location qualifier) to the upgrade check and stats reporting.

PR #1423, opened 2025-03-10
add options for geojson name and description keys.

tsteven4:
also, name created tracks.

This allows an improved reading of the "Boulder Area Trails" from https://opendata-bouldercounty.hub.arcgis.com/search?tags=recreation with "-i geojson,name=TRAILNAME".

### DIFF
--- a/geojson.cc
+++ b/geojson.cc
@@ -67,10 +67,10 @@ GeoJsonFormat::geojson_waypt_pr(const Waypoint* waypoint) const
   // Build up the properties.
   QJsonObject properties;
   if (!waypoint->shortname.isEmpty()) {
-    properties[NAME] = waypoint->shortname;
+    properties[name_opt] = waypoint->shortname;
   }
   if (!waypoint->description.isEmpty()) {
-    properties[DESCRIPTION] = waypoint->description;
+    properties[desc_opt] = waypoint->description;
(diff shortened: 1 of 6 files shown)


robertlipe:
LGTM. I've always been torn on options like this as they're not awesome in the GUI, but this makes it awesome more some people.

tsteven4:
@robertlipe 
> I've always been torn on options like this as they're not awesome in the GUI,
I am trying to understand your meaning.
The GUI does initialized the line edit with the default value initially.  But after that the any changes are remembered in the format list and the user can't no longer see the default or reset it to the default without knowing what the default is.
If that is your concern we could
1) create a tool tip "The default value is ...".  Some users will be unaware they can see tool tips by hovering.
2) consider resetting the line edit to the default when the check box when and if the user clears the checkbox.  If the user is deselecting an option do they really want to remember the value?

robertlipe:
You're right, of course. Tooltips are a discoverability bad dream. If we took the high road, we'd have to invent another channel in -^4 or something in order to smuggle them from the core to the GUI and find a way to translate them. Blech.

I'm fine doing nothing additional here and just merging it as-is. It's a very slight problem that we have, but I'm not sure it's worth solving well - if at all.

PR #1434, opened 2025-04-13
Add windows arm64 build to CI

tsteven4:
Also, make the windows x64 build runnable on windows arm64 in emulation mode.

The windows arm64 build does not include the map preview feature.  QtWebEngine is not available for windows arm64 yet.


tsteven4:
This creates one installer for windows x64 compatible systems, and one installer for windows arm64 systems.  It would be possible to produce one larger installer to handle all the windows x64 and arm64 without emulation.  A disadvantage of that would be a larger installer.  An advantage would be that users wouldn't mistakenly run the x64 executables on arm64 hardware using emulation.  We could also prevent this when we build the x64 installer.
 
In the future we should have access to windows arm64 runners, and we could produce the arm64 installer on them without cross compiling.

PR #1445, opened 2025-05-27
add GUI to snap builds

tsteven4:
The snap adds support for the GUI, which is accessed as gpsbabel.gpsbabelfe according to snapcraft conventions.  The CLI continues to be available as gpsbabel.  You can create an alias so you can invoke the GUI using the traditional name gpsbabelfe:
```sudo snap alias gpsbabel.gpsbabelfe gpsbabelfe```

This uses the gpu-2404 snap interface, although it is mostly hidden by the gnome extension.

This has been tested on a Windows 11 pro host running with Ubuntu 24.04 VMs under both WSL 2 and Hyper-V.  It should work on other Linux systems that support snapd as well.  Specifically, the GUI map preview feature worked properly.


tsteven4:
\# snap installation instructions:
\# Install the snap.  Because it is unsigned the --dangerous option is required.  Change the snap name to match the snap you downloaded.
sudo snap install --dangerous gpsbabel_1.10.0-72-gbef896d1_amd64.snap
\# Setup an alias so the GUI can be accessed as gpsbabelfe in addition to the default name gpsbabel.gpsbabelfe. (optional)
sudo snap alias gpsbabel.gpsbabelfe gpsbabelfe
\# List the snap connections (optional)
snap connections gpsbabel
\# Manually connect the browser-support interface for the GUI map preview feature.  (optional)
sudo snap connect gpsbabel:browser-sandbox
\# Manually connect the raw-usb interface to allow access to usb devices.  (optional)
sudo snap connect gpsbabel:raw-usb
\# Manually connect the removable-media interface to allow access to removable media.  (optional)
sudo snap connect gpsbabel:removable-media
\# Manually connect the serial-port interface to allow access to any serial ports.  This command may fail with 'error: snap "snapd" has no "serial-port" interface slots' if no serial ports exist. (optional)
sudo snap connect gpsbabel:serial-port

PR #1452, opened 2025-07-23
fix github action failures on windows of "Test Installer" job

tsteven4:
The "Test Installer" job has started failing.  Apparently it if failing because installation of the vc redistributable is indicating the need for a restart.  The restart can be avoided by adding the /norestart flag, which allows the test to run successfully.

However, we should consider that when users run the installer this flag will also prevent a restart.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -198,7 +198,7 @@ jobs:
       shell: pwsh
       run: |
         $install_path=Join-Path $(Get-Location) bld
-        .\bld\GPSBabel-*-*-Setup-arm64.exe /SILENT /NORESTART /DIR="${install_path}" /LOG=bld\log.txt
+        .\bld\GPSBabel-*-*-Setup-arm64.exe /SILENT /NORESTART /DIR="${install_path}" /LOG=bld\log.txt /MERGETASKS="!vcredist"
         Start-Sleep -Seconds 60
         Get-ChildItem bld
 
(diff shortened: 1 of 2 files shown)


robertlipe (commented):
Do we have the ability to pass arguments in here from CI to enable this only from CI?

It's pretty sad that installing application software still requires reboots. What year is this?

tsteven4:
49672d2 adds a task to run the vcredist install.  The task is enabled by default.  It can be disabled on the command line when running setup, as in https://github.com/GPSBabel/gpsbabel/blob/49672d266d1fc37179b5ae8de72b3621e822261f/.github/workflows/windows.yml#L201
, or by unchecking the associated check box on the installer screen.

This opens up the possibility that an interactive user disables the vcredist install by unchecking the associated box.  If an appropriate vcredist is not installed gpsbabel may fail to run.

tsteven4:
If the vcredist is missing you may see
image

tsteven4:
Additional error messages may refer to MSVCP140.dll and VCRUNTIME140.dll.

GPSBabelDeveloper:
Thanx for the warning that people can do dumb things.

I still think this helps keep CI doing without us messing with it, so it's
a nice win.

Well done. Thank you.

tsteven4:
While this fix was effective, it appears it was also fixed by a change in the vm image.
In our scheduled batch job this image failed
```
VM Image
  - OS: Windows (Arm64)
  - Source: Partner
  - Name: Microsoft Windows 11 Desktop by Arm Limited
  - Version: 20250716.19.1
  - Included Software: https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md
```
but this one worked
```
VM Image
  - OS: Windows (Arm64)
  - Source: Partner
  - Name: Microsoft Windows 11 Desktop by Arm Limited
  - Version: 20250728.24.1
  - Included Software: https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md
 ```

Issue #1453, opened 2025-08-08
snap Build (ubuntu-24.04-arm) build failures


tsteven4:
A recent change to mesa-2404 is causing build failures when we detect unexpected output on std error.
`: line 27: cd: /snap/gpsbabel/x1/gpu-2404/bin/../X11/locale: No such file or directory`

See https://github.com/canonical/mesa-2404/issues/63

tsteven4:
This is fixed in 25.0.7-snap191, which is available on the edge and beta channels, but not candidate and stable yet.

tsteven4:
A workaround was added in #1458.  After mesa-2404 >= 25.0.7 gets to stable we can back out that PR.

PR #1456, opened 2025-08-19
fix build without usb to fail

peatfaerie:
s/fatal/gbFatal/


tsteven4:
Thanks, good catch.  The regression failure is another issue, #1453.

PR #1464, opened 2025-09-19
use local schema with jing

tsteven4:
We recently had failures because of certificate expiration on
https://docs.oasis-open.org/docbook/xml/5.0/rng/docbook.rng.
We have a history failures with jing following redirects which led
to using https over http.
Try to avoid all this by using a local copy of the schema with jing.


robertlipe (approved):
Awesome.

PR #1466, opened 2025-10-01
feat: Add runtime validation for option types

robertlipe:
This PR introduces a runtime validation mechanism to catch mismatches between declared and actual option types in `arglist_t` tables.

**Key changes:**
- A `get_type()` virtual method was added to the `Option` class hierarchy to enable runtime type identification.
- Validation logic was added to `Vecs::validate_args` to compare the declared `argtype` with the actual `Option` type, calling `gbFatal` on a mismatch.
- This addresses an issue where an incorrect `argtype` in the `kalman` filter's options was not caught at startup.
- The `Option` class hierarchy was refactored to fix various build errors, including breaking a circular dependency between `defs.h` and `option.h`.


src/core/nvector.cc
```diff
@@ -135,6 +135,14 @@ double NVector::distance(double latitude_a_degrees, double longitude_a_degrees,
...
 }
 
+double NVector::euclideanDistance(const NVector& a, const NVector& b)
+{
+    double dx = a.getx() - b.getx();
+    double dy = a.gety() - b.gety();
+    double dz = a.getz() - b.getz();
+    return std::sqrt(dx*dx + dy*dy + dz*dz) * gpsbabel::MEAN_EARTH_RADIUS_METERS;
+}
+
```
tsteven4, line 145: double NVector::euclideanDistance(const NVector& n_EA_E, const NVector& n_EB_E)
{
    // https://www.ffi.no/en/research/n-vector, Example 5, dAB.
    return (n_EA_E - n_EB_E).norm() * MEAN_EARTH_RADIUS_METERS;
}

robertlipe:
Oof. Yes. Thers

robertlipe:
Oof. Yes. Theres way more stuff in this or than I meant to expose. I'll integrate that and fix the fallout. 

Thanx for a actually reading my scribbling.

defs.h
```diff
@@ -46,8 +46,9 @@
 #include "geocache.h"                // for Geocache
 #include "formspec.h"                // for FormatSpecificData
 #include "inifile.h"                 // for inifile_t
-#include "option.h"
 #include "session.h"                 // for session_t
+
```
tsteven4, line 50: As the mismatched options are already detectable at test please revert all changes to this file.

xmldoc/formats/options/xcsv-utc.xml
```diff
@@ -0,0 +1 @@
+<para>Write timestamps with offset x to UTC time.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-VAT.xml
```diff
@@ -0,0 +1 @@
+<para>Include Total Energy Variometer in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-TRT.xml
```diff
@@ -0,0 +1 @@
+<para>Include True Track in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-TAS.xml
```diff
@@ -0,0 +1 @@
+<para>Include True Airspeed in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-SIU.xml
```diff
@@ -0,0 +1 @@
+<para>Include number of satellites in use in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-OAT.xml
```diff
@@ -0,0 +1 @@
+<para>Include Outside Air Temperature in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-GSP.xml
```diff
@@ -0,0 +1 @@
+<para>Include Ground Speed in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-GFO.xml
```diff
@@ -0,0 +1 @@
+<para>Include G Force in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-FXA.xml
```diff
@@ -0,0 +1 @@
+<para>Include Fix Accuracy in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-ENL.xml
```diff
@@ -0,0 +1 @@
+<para>Include Engine Noise Level in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

xmldoc/formats/options/igc-ACZ.xml
```diff
@@ -0,0 +1 @@
+<para>Include Z Acceleration in IGC output.</para>
```
tsteven4, line 1: this is outside of scope of the kalman filter.

vecs.cc
```diff
@@ -1073,7 +1073,42 @@ bool Vecs::validate_args(const QString& name, const QVector<arglist_t>* args)
       if (arg.argval == nullptr) {
         Warning() << name << "option" << arg.argstring << "does not point to an Option instance.";
         ok = false;
+        continue;
```
tsteven4, line 1076: As the mismatched options are already detectable at test please revert all changes to this file.

tools/make_gpsbabel_doc.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 set -ex
 
+make -j8
```
tsteven4, line 4: This assumes we generate Makefiles, and is incompatible with Ninja.  cmake knows about the dependency of doc generation on gpsbabel and will ensure it is available before the doc is generated.

Please revert all changes to this file.

src/core/nvector.h
```diff
@@ -83,6 +83,7 @@ class NVector : public Vector3D
   static LatLon crossTrackProjection(double latitude_a_degrees, double longitude_a_degrees, double latitude_b_degrees, double longitude_b_degrees, double latitude_x_degrees, double longitude_x_degrees);
   static double crossTrackDistance(const NVector& n_EA_E, const NVector& n_EB_E, const NVector& n_EX_E);
   static double crossTrackDistance(double latitude_a_degrees, double longitude_a_degrees, double latitude_b_degrees, double longitude_b_degrees, double latitude_x_degrees, double longitude_x_degrees);
+  static double euclideanDistance(const NVector& a, const NVector& b);
```
tsteven4, line 86: parameter names n_EA_E, n_EB_E to match other functions and .cc file.

src/core/nvector.cc
```diff
@@ -135,6 +135,14 @@ double NVector::distance(double latitude_a_degrees, double longitude_a_degrees,
   return result;
 }
 
+double NVector::euclideanDistance(const NVector& a, const NVector& b)
```
tsteven4, line 138: see earlier comment regarding parameter naming and implementation
(return (n_EA_E - n_EB_E).norm() * MEAN_EARTH_RADIUS_METERS;)

option.h
```diff
@@ -20,6 +20,7 @@
 #define OPTION_H_INCLUDED_
 
 #include <QString>     // for QString, operator!=
+#include "defs.h"
```
tsteven4, line 23: As the mismatched options are already detectable at test please revert all changes to this file.

option.cc
```diff
@@ -140,3 +140,28 @@ bool OptionDouble::trailing_data_allowed() const
 {
   return allow_trailing_data_;
 }
+
```
tsteven4, line 143: As the mismatched options are already detectable at test please revert all changes to this file.

GEMINI.md
```diff
@@ -0,0 +1,140 @@
+This is GPSBabel, known publicly as https://gpsbabel.org and
```
tsteven4, line 1: Do we really want to add this file?  If so, it needs work.  For example we may or may not use Makefiles, and if we do they need to be generated, so the Quick build instructions are wrong.

gemini-tmp/kalman.h
```diff
@@ -0,0 +1,110 @@
+/*
```
tsteven4, line 1: the whole gemini-tmp directory shouldn't be included.

gemini-tmp/kalman.cc
tsteven4, line 1: the whole gemini-tmp directory shouldn't be included.

gui/coretool/core_strings.h
tsteven4, line 1: This file shouldn't be changed, please restore it.

tsteven4 (changes requested):
This review addresses extraneous changes in this PR, but doesn't get into the intent of the PR.  Please address these issues, and merge with master, and I will do an additional review of the core of this PR, the addition of a kalman filter.   

Once we get all the extraneous things flagged out I don't think you need to break this up into multiple PRs.

xmldoc/filters/kalman.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-gap_factor.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-interp_max_dt.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-interp_min_multiplier.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-max_speed.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-profile.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-q_scale_pos.xml
tsteven4, line 1: needs contentneeds content

xmldoc/filters/options/kalman-q_scale_vel.xml
tsteven4, line 1: needs content

xmldoc/filters/options/kalman-r_scale.xml
tsteven4, line 1: needs content

testo.d/kalman.test
```diff
@@ -0,0 +1,40 @@
+rm -f ${TMPDIR}/kalman_out.gpx
+./gpsbabel -i gpx -f ${REFERENCE}/track/kalman_in.gpx -x kalman -o gpx -F ${TMPDIR}/kalman_in_filtered.gpx -o gpx -F ${TMPDIR}/kalman_out_filtered.gpx
```
tsteven4, line 2: These should all be just gpsbabel, not ./gpsbabel.  In test gpsbabel is a function.

kalman.h
```diff
@@ -0,0 +1,111 @@
...
+#include "src/core/nvector.h"
+#include <QtCore/QVector>
+#include <deque>
+#include <vector>
+#include "option.h" // Added for OptionDouble
+
+struct KalmanExtraData {
+    bool is_zinger_deletion = false;
+};
+
```
tsteven4, line 32: I suggest moving the definition of KalmanExtraData to the top of the private section in class Kalman to limit the scope.  This isn't anything that needs to be known outside of the class implementation.

kalman.cc
```diff
@@ -0,0 +1,522 @@
...
+    F_(6, 6),
+    H_(3, 6),
+    R_(3, 3),
+    Q_(STATE_SIZE, STATE_SIZE),
+    r_scale_option_(),
+    q_scale_pos_option_(),
+    q_scale_vel_option_()
+{
+//	global_opts.debug_level = debugLevelInfo; // TODO: get from options
+    global_opts.debug_level = std::max(global_opts.debug_level, debugLevelInfo); // TODO: get from options
```
tsteven4, line 60: I think you are ready to just let main set global_options.debug_level, eliminating the body of this ctor.

kalman.cc
```diff
@@ -0,0 +1,522 @@
...
+    Matrix d_squared_matrix = y.transpose() * S_inv * y;
+    double d_squared = d_squared_matrix(0, 0);
+
+    if (d_squared < CHI_SQUARED_THRESHOLD) {
+      Matrix K = P_ * H_.transpose() * S_inv;
+
+      x_ = x_ + (K * y);
+      P_ = (Matrix::identity(6) - (K * H_)) * P_;
+    } else {
+    }
```
tsteven4, line 503: empty clause

robertlipe:
It's hopefully clear that I never intended a majority of this PR to BE IN this PR and certainly not to be pushed publicly. Sorry that you've invested so much in it. I should have marked it as a draft, but I knew that only one person ever reads this stuff and thought he wasn't going to see it. 

Then I got distracted and you invested in this. Sorry.

Whether I pull 80% of this into other PRs or pull the 20% of this into a single PRthat's actually worth human attention seemed like 6's (six of one, half a dozen...) at the time. I was choosing the latter and starting to break it into smaller pieces that can go in one at a time, though ordering might be tricky.

Things like gbDebug (or whatever we choose to call it) are arguably projects of their own. Maybe each module should have a Logging object that persists that could store its debugging level) inside it so that input could have one level and output could have its own level and we wouldn't keep creating static instances of these empty objects to pretend they're not globals. Then all the member calls become logging_.Warning(foo) and logging_.Debug(6, ...) and such. Maybe it.s logging_.(Warning, foo) vs. logging_.(Debug, 6, foo) or such. I've really not explored that. It's an example of a half-baked idea that just fell out during this process. 

I had two big downfalls:
1) Getting tired of being asked every time "can I run git diff? Can I run git reset" and answering 'y'. Chosing "don't ask me again" also gives it permission to do anything possible with git, like make more branches, push the tree, bring in all the scribbly files that yuou might exist in my ~/src/gpsbabel space, etc.
2) Trying overlapping projects in the same workspace. This human can (and has for years) easily keeps separate "those files that have needed doc for years" and such and while it's convenient to have overlapping agents working on things, I've now learned that it makes a disastrous mess of things, owing somewhat to # 1, that then becomes Real Work to separate back apart.

I'm finally happy with the results of kalman.cc - I just have to do the work to get it product-worthy. It was probably the most difficult project I've tackled with AI (and clearly one I couldn't have done without out it) but it was still a really painful slog.

I'll be more careful to mark non-trivial things as WIP to be more respectful of your review input.

tsteven4:
I tired this with some real world cycling data from a garmin edge device that was recording without issues.  There wasn't a need to filter this data at all.  However, without parameters, the filter introduced some artifacts.  The recorded track and points are blue, the filtered track and points are red.  I hadn't read your doc yet and didn't select a profile.
./bld/gpsbabel -i gpx -f ht.gpx -o kml -F ht.kml -x kalman -o kml -F htf.kml
Subsequently I tried
./bld/gpsbabel -i gpx -f ht.gpx -o kml -F ht.kml -x kalman,profile=cycling -o kml -F htfcyc.kml
but that showed the same artifacts.
The full data set is 24931 points.

I can get some problematic recordings from a phone that might benefit from filtering, but I haven't got them yet.

ht1
ht2
ht3

tsteven4:
Perhaps the speeds are relevant, at the start of the artifacts they were 33.2 mph, 32.9 mph, 33.4 mph (14.8 mps, 14.7mps, 14.9mps).  Faster speeds occur regularly when descending steeper terrain.  The average speed of the entire data set was 15.4mph (6.9mps), so maybe the cycling profile is automatically selected (is auto selection described in document?).  However, max_speed_option_.has_value() is true even when not set because of the default provided, so auto select isn't in play.  I tried max_speed=30 but that didn't fix it.

GPSBabelDeveloper:
They are indeed relevant. It looks like it's over-aggressively
smoothing/filtering. I've looked at this enough that this MIGHT be the same
issue three times,  It looks like a low Q as you can see in the super cool
https://github.com/GPSBabel/gpsbabel/pull/1477/files#diff-66e756d9695282bf21acb4caf360a34cfc38d451f07129ffcf1d74d9491794c1
Counterintuitively, the low q is slow to react and aims for where the
moving object used to be. It's the family sedan that's slow to react. A
high Q is the blue line that's fast to track, but has high 'inertia' and
overshoots on a change (like oversteer) so when there's a change it's
disoriented until a new direction is clear, then it quickly snaps back.
It'll depend on direction and such, but maybe a higher Q is appropriate for
cycling.

We're likely to have data-specific issues until we get these profiles
dialed in. My test case is a cycling track (from a flat-lander that's on a
brightly colored bike that weighs a ton, wink) so I don't think we're TOO
far off from making these awesome, but we may need to dial in r_scale_
(measurement noise) and q_scale_pos_ (process noise for position), but
we're going to need the direction and time of the GPX segments in question.
If you can snip them, that'd be great. If you want to just flip me the file
as a test, we can do that.

If you want to provide them to me privately, I can anonymize them, capture
the important parts, and shove them into the ocean as part of the test
suite. We have 42 kalman*out files at this point, each providing one or
more specific cases, and I'm sure we'll be adding more.

The presence of data-specific issues was one of the clouds that hung over
my head for not wanting to do this for so long. It's nightmarish.

I'm going to focus now on getting all the pieces landed. I'll get
everything into the tree overnight for you to do at least a coarse-level
review of them, and then we can do a more thorough one. I'm almost tempted
to just commit them and then do a review as they've landed, as this keeps
growing.

Back to it now.
RJL

PR #1468, opened 2025-10-06
update libusb to 1.0.29

tsteven4:


robertlipe (approved):
LGTM

Mumbling aloud...Twenty years on, I'm pretty sure that we'll never need libusb again.  We could almost drop this dependency and do our own IOKit implementation within our little OS USB abstraction, but that's mostly busy work with no realy upside. Thankfully, the turbulence in libusb settled a long time ago.

tsteven4:
Actually there is a [macOS 26 libusb issue](https://github.com/libusb/libusb/pull/1700) that may or may not bother us.  It has been fixed but is unreleased.

PR #1469, opened 2025-10-06
update shapelib to 1.6.2

tsteven4:


robertlipe:
I quit reading half through but this seems mostly as a good commercial for RAII

I don't miss this class of problems.

PR #1470, opened 2025-10-07
use Qt 6.10.0 for windows arm64 builds

tsteven4:
This is the first release to support QtWebEngine for WoA.

Unfortunately our Qt installer hasn't caught up yet, so a pre-release version of aqtinstall is used on windows.


tsteven4:
windows regression failed as it couldn't access secrets from a forked repo.

PR #1472, opened 2025-10-07
release candidate 1.11.0

tsteven4:
The following release versions are built:
macos: Qt 6.8.3. requires macOS 12 or higher.
windows x86_64: Qt 6.8.3. windows 10 or higher.
windows on arm: Qt ~~6.10.0~~6.10.2.  includes GUI with map preview.  ~~6.10.0 was released today, 10/7/2025.~~
ubuntu: snap for amd64 and arm64.  uses ubuntu distributed Qt from noble, Qt 6.4.2.  includes GUI with map preview.

We can merge other branches if desired before release.


robertlipe:
Wonderful. Thank you.

I'm planning to land my Kalman work before 1.11. It won't change anything
in base that I can think of and wouldn't invalidate any testing done. I
think the design and implementation is pretty well locked down with good
coverage. I just need to sort out branch wreckage and do some merge
squashing so all zillion versions don't peek through as there as a LOT of
back and forth, but there was also a git dustup that I think di a lot of
that squashing for us and deleted some other things. Whatever. I'm pretty
happy with what's in hand.

I may make multiple branches with smaller collections of things I did along
the way. I've had to take a few days off, but
https://github.com/GPSBabel/gpsbabel/commits/kalman-2/
got pretty out of hand. I'm going to break it into more manageable pieces.

tsteven4:
Issue: The version naming scheme in snapcraft.yaml is based on the tags that exist when it is built, and we usually don't tag a release before creating the release assets.  A workaround is to build the snaps for the release after the tag is generated.  This means when the new release appears initially the snaps won't be available, but they will be added later.  Note that many package systems monitor our tags, and the existence of a release tag will trigger package generation, so we don't want to create the tag until we are confident of the release.

tsteven4:
@robertlipe I think we are close
Qt 6.10.3 should release Thursday
Can we enter macos certificate data into github secrets for signing with a real cert instead of ad-hoc?
dg-388 might slip in under the wire.

tsteven4:
@robertlipe another preference of mine would be to remove deprecated/autotools which snuck back in.  But perhaps you find that useful.

tsteven4:
I'm closing this so I can reopen a native PR that has access to secrets.  SHA will be the same.

PR #1473, opened 2025-10-07
feat: Add testsuite validation for option types

robertlipe:
- **feat: Add runtime validation for option types**
- **Revert "feat: Add runtime validation for option types"**
- **feat: Add runtime validation for option types**


gui/coretool/core_strings.h
tsteven4, line 1: please restore this file.  I think our build system deletes in inappropriately on clean as cmake it is a (sometimes) generated file.

option.cc
```diff
@@ -140,3 +140,28 @@ bool OptionDouble::trailing_data_allowed() const
...
+
+const QString& OptionInt::get() const
+{
+  return value_;
+}
+
+OptionInt::operator const QString& () const
+{
+    return value_;
+}
```
tsteven4: why did these trivial functions move from the h file?  This is inconsistent with OptionString, OptionDouble, OptionBool.  It makes the changes harder to review.

option.h
```diff
@@ -46,6 +47,8 @@ class Option /* Abstract Class */
   virtual void init(const QString& id) {}
   virtual void reset() = 0;
   virtual void set(const QString& s) = 0;
+  virtual bool isValid(const QString& s) const = 0;
```
tsteven4: shouldn't isValid be [[nodiscard]]?  (everywhere).

defs.h
```diff
@@ -46,8 +46,9 @@
 #include "geocache.h"                // for Geocache
 #include "formspec.h"                // for FormatSpecificData
 #include "inifile.h"                 // for inifile_t
-#include "option.h"
 #include "session.h"                 // for session_t
+
+class Option;                      // for arglist_t
```
tsteven4: comment should be // forward declare
I would suggest moving it after all the #includes.

vecs.cc
```diff
@@ -1073,7 +1073,42 @@ bool Vecs::validate_args(const QString& name, const QVector<arglist_t>* args)
...
+
+          if (!is_ok) {
+              gbFatal("Mismatched option type for '%s' in format/filter '%s'. "
+                      "Declared: %u, Actual: %u",
+                      CSTR(arg.argstring), CSTR(name),
+                      declared_type, actual_type);
+          }
+        }
+      }
+
```
tsteven4, line 1111: Aren't these cases covered already by the original code below which uses a dynamic cast to deduce the type?  **I'm missing something or this whole PR seems redundant.**  I'm trying to find the original failure but with the forced pushes I'm having trouble seeing what the problem was.

tsteven4 (commented):
I can't recreate the original problem as history has been erased by forced pushes.  I think the original code should work, but evidently there was a hole?

tsteven4:
BTW the validation is not at startup, but when you run the validate_formats test.
./testo -p bld/gpsbabel validate_formats

tsteven4:
If I take kalman-2, restore options.h and options.cc from main, take out the new check code in vecs.cc so only the original checks exist, and then introduce type mismatches in kalman.h:
```
-  OptionDouble gap_factor_option_;
+  OptionInt gap_factor_option_;
```
and
```
     {"r_scale", &r_scale_option_, "Measurement noise covariance scaling factor",
-      "1.0", ARGTYPE_FLOAT, ARG_NOMINMAX, nullptr
+      "1.0", ARGTYPE_INT, ARG_NOMINMAX, nullptr
```
the the test shows:
```
Running validate_formats.test
main: "kalman" OptionInt option without trailing data "gap_factor" is not of ARGTYPE_INT.
main: "kalman" Int option "gap_factor" default value "10.0" is not an integer.
main: "kalman" OptionDouble without trailing data "r_scale" is not of ARGTYPE_FLOAT.
bld/gpsbabel returned error 1
```

tsteven4:
I see some activity in the kalman-2 branch that shows the kalman options were originally ARGTYPE_STRING instead of ARGTYPE_FLOAT.  If I 

1. take the kalman-2 branch and 
2. restore vecs.cc, option.h and option.cc from master, 
3. and change ARGTYPE_FLOAT -> ARGTYPE_STRING in kalman.h to create the mismatches
4. and run the test the mismatches are detected:

```
tsteven4@PEDALDAMNIT:~/work/kalman2$ ./testo -p bld/gpsbabel validate_formats
Running validate_formats.test
main: "kalman" OptionDouble without trailing data "gap_factor" is not of ARGTYPE_FLOAT.
main: "kalman" OptionDouble without trailing data "r_scale" is not of ARGTYPE_FLOAT.
main: "kalman" OptionDouble without trailing data "q_scale_pos" is not of ARGTYPE_FLOAT.
main: "kalman" OptionDouble without trailing data "q_scale_vel" is not of ARGTYPE_FLOAT.
main: "kalman" OptionDouble without trailing data "max_speed" is not of ARGTYPE_FLOAT.
main: "kalman" OptionDouble without trailing data "interp_max_dt" is not of ARGTYPE_FLOAT.
main: "kalman" OptionDouble without trailing data "interp_min_multiplier" is not of ARGTYPE_FLOAT.
bld/gpsbabel returned error 1
```

tsteven4:
> I see some activity in the kalman-2 branch that shows the kalman options were originally ARGTYPE_STRING instead of ARGTYPE_FLOAT. If I
> 
> 1. take the kalman-2 branch and
> 2. restore vecs.cc, option.h and option.cc from master,
> 3. and change ARGTYPE_FLOAT -> ARGTYPE_STRING in kalman.h to create the mismatches
> 4. and run the test the mismatches are detected:
> 
> ```
> tsteven4@PEDALDAMNIT:~/work/kalman2$ ./testo -p bld/gpsbabel validate_formats
> Running validate_formats.test
> main: "kalman" OptionDouble without trailing data "gap_factor" is not of ARGTYPE_FLOAT.
> main: "kalman" OptionDouble without trailing data "r_scale" is not of ARGTYPE_FLOAT.
> main: "kalman" OptionDouble without trailing data "q_scale_pos" is not of ARGTYPE_FLOAT.
> main: "kalman" OptionDouble without trailing data "q_scale_vel" is not of ARGTYPE_FLOAT.
> main: "kalman" OptionDouble without trailing data "max_speed" is not of ARGTYPE_FLOAT.
> main: "kalman" OptionDouble without trailing data "interp_max_dt" is not of ARGTYPE_FLOAT.
> main: "kalman" OptionDouble without trailing data "interp_min_multiplier" is not of ARGTYPE_FLOAT.
> bld/gpsbabel returned error 1
> ```

@robertlipe My conclusion is the original option checking was sufficient to detect the errors you had.  I suspect that you assumed the validation ran at startup instead of during the validate_formats test (which also validates the filter arguments).

gui/coretool/core_strings.h
robertlipe: It shows up as a zillion deleted files all the time, so I just made the deletion harder. It's back now and so it will stay.

tsteven4:
@robertlipe note that catching any mismatches at test in CI prevents any mismatches from being merged through a PR, but it doesn't prevent them from confusing a developer who hasn't run the test yet.

gui/coretool/core_strings.h
tsteven4: I think the problem stems from supporting builds without the linguist tools, but wanting the linguist tools to update the files other times.  This results in a generated file that is checked in.  This confuses the cmake clean target.  I have known of this issue but I haven't known how to fix it.

vecs.cc
robertlipe: Indeed. When we (Gemini and I) were adding new options in kalman.cc, we triggered a crash from a mismatched type. It was somethin glike the difference in the arguments between strings and floats (sometimes we "atof" and sometimes we ust want a number) and things would get ugly. We fixed the caller, but it occurred to. methat we should be able to catch it (when I started, I was thinking at startup and that probably stuck in my head) but we indeed learned it's in the test - which is cool! This is kind of how I got in trouble because i started two-timing the kalman work and this work at the same time and I think that having two PR-to-bes in one branch confuses it WAY more that it screws me up. I can USUALLY pull it off, but you can tell that I've spent the evening pullin the Kalman work apart into manageable PRs. I DOSes myself. :-/

It's also indeed horse-crap that it chose to delete everything checked in and rewrite git history, nuking a whole lot of history each at least once. SO maddening.

tsteven4: So this PR can be abandoned and closed, correct?

PR #1474, opened 2025-10-08
feat: Add core Matrix class for mathematical operations

robertlipe:
Introduces a `Matrix` class in `src/core` to support advanced
mathematical operations required by filters like the Kalman filter.

Features:
- Basic arithmetic (+, -, *)
- Transposition and inversion
- Modern C++ features including move semantics and in-place
  operators for performance.

The new class has been integrated into the CMake and qmake build
systems.


tsteven4:
Do you really want to reintroduce qmake.  I don't want to support it.

GPSBabelDeveloper:
No. I did a double take on it myself, but convinced it was used based on
ack -l qmake  . in usages in CI that seemed like they were in play. The
file was here, so I 'fixed' it. I would be happy to instead simply delete
it.

Say the word. I'll drop it in another PR.

src/core/matrix.cc
```diff
@@ -0,0 +1,209 @@
...
+    if (this != &other) {
+        rows_ = other.rows_;
+        cols_ = other.cols_;
+        data_ = std::move(other.data_);
+        other.rows_ = 0;
+        other.cols_ = 0;
+    }
+    return *this;
+}
+
```
tsteven4, line 52: I believe you can just use the implicitly defined move ctor and assignment operator.  Delete this block.

src/core/matrix.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <vector>
+
+class Matrix {
+public:
+    Matrix(int rows, int cols);
+    Matrix(const Matrix&) = default;
+    Matrix& operator=(const Matrix&) = default;
+    Matrix(Matrix&& other) noexcept;
+    Matrix& operator=(Matrix&& other) noexcept;
```
tsteven4, line 12: You can delete these declarations.  The copy and move ctors and assignment operators will be implicitly defined as you haven't define any other ...
What you don't get, and apparently don't need, is an implicit ctor, i.e. Matrix().

src/core/matrix.cc
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return result;
+}
+
+Matrix Matrix::operator+(const Matrix& other) const {
+    if (rows_ != other.rows_ || cols_ != other.cols_) {
+        gbFatal("Matrix dimensions must match for addition.");
+    }
+    Matrix result(rows_, cols_);
+    for (size_t i = 0; i < data_.size(); ++i) {
```
tsteven4, line 140: src/core/matrix.cc:123:10: warning: no header providing "size_t" is directly included [misc-include-cleaner]

The type of std::vector::size() is usually size_t but it isn't promised.  deduction with auto seems to use an int because of the initializer.  So we are left with this less than beautiful method, or knowing size_t is must be big enough just include a header for size_t ( which may or may not be signed).

for (std::vector<double>::size_type i = 0; ...

The same at L134, L157, L167, L177 (5 instances in total).

tsteven4 (commented):
I just got started, here is what I found so far.

src/core/matrix.cc
robertlipe: Fixed in #1478 which is fast follow to this one.

src/core/matrix.h
robertlipe: Yep, I think we have rule of 0 (?) now.

src/core/matrix.cc
robertlipe: Done in #1479

robertlipe (commented):
This one isn't likely to be contentious. I'm going to merge it as #1479 depends on this one.

src/core/matrix.h
tsteven4: yes.  rule of zero is the way to go if possible.

PR #1477, opened 2025-10-09
Add Kalman doc.

robertlipe:
Try to make Kalman understandable to our audience. Shows how it's applied, with visual (including animations for the web!) presentation when it helps.

The PDF looks great, IMO.
The HTML generation probably needs help and i don't know if a 900K GIF will fly, but IMO it _really_ helps.

I'm pretty proud of all this, too. :-)


xmldoc/filters/options/kalman-max_speed.xml
```diff
@@ -0,0 +1,35 @@
...
+  You could set:
+<screen>
+gpsbabel -i gpx -f my_ride.gpx -x kalman,max_speed=15 -o gpx -F my_ride_filtered.gpx
+</screen>
+</para>
+<para>
+  The units are always in meters per second.
+</para>
+<para>
+  The pre-defined profiles use the following values: 'walking' (2.0), 'running' (10.0), 'cycling' (15.0), 'driving' (50.0), and 'flying' (300.0). If you are using a profile, you do not need to set this option unless you want to override the profile's default.
```
tsteven4: cycling default in kalman-main is now 30.0.

xmldoc/filters/options/kalman-q_scale_pos.xml
```diff
@@ -0,0 +1,29 @@
...
+  This option controls how much the filter expects your position to
+  change unpredictably between GPS points. It models the uncertainty
+  in your movement dynamics.
+</para>
+<mediaobject>
+  <imageobject role="pdf">
+    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.png" format="PNG" width="100%" scalefit="1"/>
+  </imageobject>
+  <imageobject role="html">
+    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF"/>
```
tsteven4: To get this to adjust to the browser width I needed to add width and scalefit for html.  Without this change I had to adjust the browser width to see the whole gif.
```
-    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF"/>
+    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF" width="100%" scalefit="1"/>
```

robertlipe:
I know you're not happy about <trkseg>. Noted. I'll work on it.

What's remaining between here and an approval on these? The referernce files will need regenerated since we changed the thresholds, but I'm not going to do that until everything is all in one testable tree again.  I have tools to handle that regeneration that I'll toss in too, because these files are super fragile.

xmldoc/filters/options/kalman-q_scale_pos.xml
```diff
@@ -0,0 +1,28 @@
...
+  This option controls how much the filter expects your position to
+  change unpredictably between GPS points. It models the uncertainty
+  in your movement dynamics.
+</para>
+<mediaobject>
+  <imageobject role="pdf">
+    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.png" format="PNG" width="100%" scalefit="1"/>
+  </imageobject>
+  <imageobject role="html">
+    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF"/>  </imageobject>
```
tsteven4:
```suggestion
    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF" width="100%" scalefit="1"/>  </imageobject>
```

tsteven4 (changes requested):
I think you tried to do this but it didn't take.  Now I am trying to suggest the change which hopefully will be clickable to accept.

robertlipe:
I'm about ready to go back to subversion. Then commit and push is right there 

https://github.com/GPSBabel/gpsbabel/pull/1477/commits/79c7d30e4f9beffc321aab4beedad5807725c484

tsteven4:
> I'm about ready to go back to subversion. Then commit and push is right there
> 
> [79c7d30](https://github.com/GPSBabel/gpsbabel/commit/79c7d30e4f9beffc321aab4beedad5807725c484)

Yes, I saw that, it just isn't the change I was asking for which was to add the width and scalefit parameters.
```
-    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF"/>
+    <imagedata fileref="xmldoc/images/kalman_q_scale_pos.gif" format="GIF" width="100%" scalefit="1"/>
```

In any event you got it in https://github.com/GPSBabel/gpsbabel/pull/1477/commits/6b2dd3e1beeac13277ef04c6c40dcf5b2e9bdbaf, so the doc looks gootd to me now.

tsteven4:
> What's remaining between here and an approval on these?

There are a number of suggestions I have made on #1478 that are open, e.g. a patch today concerning median calculation, rule of zero for Matrix class, rule of zero for Kalman class, ~~nvector eculideandistance using norm~~.  These aren't earth shattering cleanups.  If you want to go with what is already there (#1478, #1479, #1480) the only required change is to fix the tests.  After that I could propose the outstanding suggestions as something you could review in a follow on PR.  That would leave the track segment issue.

robertlipe:
I'll suck it on that last one. That was the FOURTH time I was asked to re-commit that  for each of the two one-liners, so by the time I got there, I was pretty pissed and probably just copied the wrong line. So I'll take the blame for being sloppy, but still blame git for just being terrible at handling overlapping development tasks. 'You can't commit foo/bar. Base is out of date!' 'Yeah, foo/blah changed yesterday. So what?' 'Run in circles, scream and shout!!!!!" So I went to the web interface and just edited them.

Also, we've had a problem lately with text between angle brackets <likethis> being deleted. Not sure if it's the web or the mobile inferface (doing this from web) as I used both.

PR #1478, opened 2025-10-10
feat: Add Kalman filter for track smoothing

robertlipe:
Introduce a new track filter based on a Kalman filter to provide
advanced smoothing and outlier correction for GPS tracks.

This filter models the track dynamics (position and velocity) and uses a
Kalman filter to estimate the true path from noisy measurements. It is
highly effective at reducing GPS jitter and rejecting spurious points
(zingers) while preserving the underlying track shape.

This initial commit includes:
- The core Kalman filter implementation.
- Integration into the build system and filter registration.
- Updates to the NVector library for required vector mathematics.

Here's the one I think I want to land. There are more landing soon with doc, testsuite, etc. Trying to get to manageable pieces. This is the one with the meat.
Note that the thing about about testing testing against default options is something that I have a better solution up my sleeve. Sooo many moving pieces on this one.


kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+#include "src/core/vector3d.h"
+#include <QDebug>
+#include <deque>
+#include <numeric>
+
+// Constants
+static constexpr int STATE_SIZE = 6;   // [x,y,z,vx,vy,vz]
+static constexpr int MEAS_SIZE  = 3;   // [x,y,z] measurements
+
+// static constexpr int debugLevelInfo = 1;
```
robertlipe, line 38: Placeholder. These probably come back when gbDebug or something like that comes back.

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+// static constexpr int debugLevelDebug = 2;
+// static constexpr int debugLevelInterpolate = 7;
+
+
+extern RouteList* global_track_list;
```
robertlipe, line 42: Once stabiliized, I may move this to track_disp_all() style of iteration. IMO the code is easier to read that way.

robertlipe (commented):
I've re-read this. It's not perfect, but I'd like to land this.

tsteven4:
The windows failure is a secret problem.  It occurs on a PR from a fork.  It won't occur from a branch of this repo.  It will go away when this is merged.  The aqtinstall developer has gone quiet for 3 months so I haven't been able to get my fix merged and released and we are using a local copy of aqt for the time being.

> + curl -u : /aqtinstall-3.3.1.dev13-py3-none-any.whl -o /tmp/aqtinstall-3.3.1.dev13-py3-none-any.whl
curl: (3) URL rejected: Malformed input to a URL function

kalman.h
```diff
@@ -0,0 +1,112 @@
...
+#ifndef KALMAN_H_INCLUDED_
+#define KALMAN_H_INCLUDED_
+
+#include "filter.h"
+#include "src/core/matrix.h"
+#include "src/core/nvector.h"
+#include <QtCore/QVector>
+#include <deque>
+#include <vector>
+#include "option.h" // Added for OptionDouble
```
tsteven4: I ran iwyu.  For consistency with everything else I recommend replacing these includes with:

#include <QDateTime>           // for QDateTime
#include <QList>               // for QList
#include <QString>             // for QString
#include <QVector>             // for QVector

#include "defs.h"              // for arglist_t, ARG_NOMINMAX, ARGTYPE_FLOAT, ARGTYPE_STRING, Waypoint
#include "filter.h"            // for Filter
#include "option.h"            // for OptionDouble, OptionString
#include "src/core/matrix.h"   // for Matrix
#include "src/core/nvector.h"  // for NVector

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+#include "kalman.h"
+#include "defs.h"
+#include "option.h"
+#include "session.h"
+#include "src/core/logging.h"
+#include "src/core/nvector.h"
+#include "src/core/vector3d.h"
+#include <QDebug>
+#include <deque>
+#include <numeric>
```
tsteven4: I ran iwyu.  For consistency I recommend replacing these includes with:

#include <algorithm>            // for max, sort
#include <cstdlib>              // for abs
#include <cstddef>              // for size_t
#include <cmath>                // for round, floor
#include <iterator>             // for distance, prev
#include <vector>               // for vector

#include <QDebug>               // for QDebug
#include <QtGlobal>             // for qRound, qDebug

#include "defs.h"               // for Waypoint, WaypointList, route_head, RouteList, wp_flags, track_add_wpt, arglist_t
#include "option.h"             // for OptionDouble, OptionString
#include "src/core/datetime.h"  // for DateTime
#include "src/core/nvector.h"   // for NVector
#include "src/core/vector3d.h"  // for Vector3D

tsteven4 (commented):
just getting started

tsteven4:
While the codacy analysis issues inspired me to look at the includes, the codacy issues can be ignored.  cppcheck doesn't need the includes.

tsteven4:
the filter artifacts I reported for kalman-2 still exist here, but unlike kalman-2, auto select works and setting max_speed=30 fixes them.  I added qDebug to print out the points marked for deletion, and there were points near the start of those artifacts that were marked with the default cycling speed.  After setting speed only 2 points are marked for deletion, those are both at places were we stopped and snacked.  Those areas don't show much difference between the filtered and unfiltered tracks.

tsteven4:
htcut, no options, extra debug statements.  the points marked for deletion are 2 for gaps and the rest for speed.  the gaps are were we stopped.  the edge device was set to auto-pause at stops (as is normal).  I recommend a higher default for cycling speed, 25m/s would cover most things, 30m/s would be safer but won't cover record Tour de France speeds (81.2mph = 36.3m/s)
kalman: auto select 6.91097
kalman: auto select cycling
kalman: point marked 40.487172 -105.210634 gap: true speed: false
kalman: point marked 40.5231 -105.176401 gap: false speed: true
kalman: point marked 40.514174 -105.158132 gap: true speed: false
kalman: point marked 40.510057 -105.156621 gap: false speed: true
kalman: point marked 40.50496 -105.15275 gap: false speed: true
kalman: point marked 40.524393 -105.136294 gap: false speed: true
kalman: point marked 40.52428 -105.135803 gap: false speed: true
kalman: point marked 40.524619 -105.135284 gap: false speed: true
kalman: point marked 40.525152 -105.135182 gap: false speed: true

kalman.h
```diff
@@ -0,0 +1,112 @@
...
+
+#if FILTERS_ENABLED
+
+class Kalman : public Filter {
+ public:
+  Kalman();
+  ~Kalman() override = default;
```
tsteven4: kalman.h:34:7: warning: class 'Kalman' defines a default destructor but does not define a copy constructor, a copy assignment operator, a move constructor or a move assignment operator [cppcoreguidelines-special-member-functions]

delete the line declaring the dtor.

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+        }
+        // Clear the old list (now held by new_route_head) to prevent its destructor from deleting Waypoints
+        // that are now owned by the main wpt_list.
+        WaypointList empty_list_for_cleanup;
+        new_route_head->waypoint_list.swap(empty_list_for_cleanup);
+        delete new_route_head; // Clean up temporary route_head object
+
+        // Now apply Kalman filter to the cleaned and interpolated data
+        for (WaypointList::iterator wpt_it = wpt_list->begin(); wpt_it != wpt_list->end(); ++wpt_it) {
+            kalman_point_cb(*wpt_it);
```
tsteven4: warning: use range-based for loop instead [modernize-loop-convert]
in a bunch of places.

The results of
clang-tidy --checks=-*,modernize-loop-convert --fix kalman.cc
are
[rangefor.patch](https://github.com/user-attachments/files/22854566/rangefor.patch)

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+        if (!dt_samples.empty()) {
+            std::sort(dt_samples.begin(), dt_samples.end());
+            size_t mid = dt_samples.size() / 2;
+            median_dt = (dt_samples.size() % 2 == 0) ?
+                        (dt_samples[mid-1] + dt_samples[mid])/2.0 : dt_samples[mid];
+            if (median_dt < 1e-6) median_dt = 1.0;
+        }
+
+        // Interpolate moderate gaps and build new route
+        route_head* new_route_head = new route_head(); // Temporary route to build the new sequence
```
tsteven4, line 337: kalman.cc:398:9: warning: use auto when initializing with new to avoid duplicating the type name [modernize-use-auto]

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+        // Iterate through the rest of the original wpt_list
+        for (auto it = wpt_list->begin(); it != wpt_list->end(); ++it) {
+            Waypoint* current_original_wpt = *it;
+
+            if (current_original_wpt == last_kept_for_interp) {
+                continue; // Skip the first point, already handled
+            }
+
+            if (current_original_wpt->wpt_flags.marked_for_deletion) {
+                KalmanExtraData* extra_data = (KalmanExtraData*)current_original_wpt->extra_data;
```
tsteven4: kalman.cc:426:17: warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+                                double frac = double(k) / (n_insert + 1);
+
+                                // Linear interpolation in NVector space
+                                gpsbabel::NVector na(last_kept_for_interp->latitude, last_kept_for_interp->longitude);
+                                gpsbabel::NVector nb(current_original_wpt->latitude, current_original_wpt->longitude);
+                                gpsbabel::NVector interpolated_nvector = gpsbabel::NVector::linepart(na, nb, frac);
+
+                                QDateTime gen_time = last_kept_for_interp->GetCreationTime().addSecs(qRound(frac * gap));
+
+                                Waypoint* new_wpt = new Waypoint();
```
tsteven4: kalman.cc:454:33: warning: use auto when initializing with new to avoid duplicating the type name [modernize-use-auto]

kalman.cc
```diff
@@ -0,0 +1,594 @@
...
+
+    x_(6, 1),
+    P_(6, 6),
+    F_(6, 6),
+    H_(3, 6),
+    R_(3, 3),
+    Q_(STATE_SIZE, STATE_SIZE),
+    r_scale_option_(),
+    q_scale_pos_option_(),
+    q_scale_vel_option_()
```
tsteven4: use default member initialization.  requires moving some constants to h file as well.

[defaultmemberinit.patch](https://github.com/user-attachments/files/22857058/defaultmemberinit.patch)

tsteven4:
consider using msecsTo instead of secsTo in kalman.cc  Some devices are sampling at 5Hz.

note that when computing speed, the denominator is artificially limited to 1.0s.

kalman.h
robertlipe: Knowing where IWYU came from, it's odd that its style is so contrary to what is pounded into Google's own style.

robertlipe:
> qDebug to print out the points marked for deletion, 

You've seen some evidence that those have come and gone many times - enough that I leaked a different debugging scheme that may come back in some form. Anything per-point in track data is just overwhelming, but I may consider some -D thingy that prints jabber at the edges of when it starts interpolating or stops interpolating and so on.  Optionally attaching data inside <notes> (because almost nothing does anything with that field) didn't seem so bad, either.

> don't show much difference between the filtered and unfiltered tracks.

Indeed. If you have a GOOD gps in modern time, dropping a point per second Kalman does almost nothing for you. Honestly, at one point I just considered keeping our pre-filter (which is almost identical to the track filter, but a little more automatic) and tossing out the complexity of Kalman totally.

robertlipe:
> consider using msecsTo instead of secsTo in kalman.cc Some devices are sampling at 5Hz.

Loggers for activities and quadcopter-style devices are at 10 and more now.

> note that when computing speed, the denominator is artificially limited to 1.0s.

Only when trying to figure out if something is a zinger or not. Something that zings out and back in a sub-second multi-point bust might be a problem. Let's fret about that later...if we ever see it.

kalman.cc
```diff
@@ -0,0 +1,582 @@
...
+        enum class PreFilterState { NORMAL, RECOVERY, FIRST_GOOD_SEEN_IN_RECOVERY };
+        PreFilterState state = PreFilterState::NORMAL;
+        Waypoint* last_accepted_wpt = nullptr;
+
+        for (auto it = wpt_list->begin(); it != wpt_list->end(); ++it) {
+            auto* const current_wpt = *it;
+
+            // Ensure KalmanExtraData is attached
+            if (current_wpt->extra_data == nullptr) {
+                current_wpt->extra_data = new KalmanExtraData();
```
tsteven4, line 263: I adapted the kalman-2 tests and found:
```
==11617== 70 bytes in 70 blocks are definitely lost in loss record 14 of 20
==11617==    at 0x4846FA3: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==11617==    by 0x2447B4: Kalman::process() (kalman.cc:294)
==11617==    by 0x27CAE5: run(char const*) (main.cc:494)
==11617==    by 0x27EA23: main (main.cc:854)
```

kalman.cc
```diff
@@ -0,0 +1,582 @@
...
+            }
+            if (interp_max_dt_option_.get_result() == 300.0) {
+                interp_max_dt_ = 120.0; // 2 minutes
+            }
+            if (interp_min_multiplier_option_.get_result() == 1.5) {
+                interp_min_multiplier_ = 1.8;
+            }
+        } else if (current_profile == "cycling") {
+            if (max_speed_option_.get_result() == 15.0) {
+                max_speed_ = 15.0; // e.g., 54 km/h
```
tsteven4: IMO, as I have demonstrated and commented on previously, 15m/s is too slow for a cycling limit.

kalman.cc
```diff
@@ -0,0 +1,582 @@
...
+        last_timestamp_ = QDateTime();
+        last_nvector_ = gpsbabel::NVector();
+        x_ = Matrix(STATE_SIZE, 1);
+        P_ = Matrix::identity(STATE_SIZE) * INITIAL_UNCERTAINTY;
+        F_ = Matrix::identity(STATE_SIZE);
+        H_ = Matrix(MEAS_SIZE, STATE_SIZE);
+        H_(0, 0) = 1.0;
+        H_(1, 1) = 1.0;
+        H_(2, 2) = 1.0;
+        R_ = Matrix::identity(MEAS_SIZE) * MEASUREMENT_NOISE_SCALE * r_scale_;
```
tsteven4: This is a usage of r_scale_ before the Auto-profile inference logic has had a chance to run.

kalman.cc
```diff
@@ -0,0 +1,582 @@
...
+        // Iterate through the rest of the original wpt_list
+        for (const auto& current_original_wpt : *wpt_list) {
+            i++;
+
+            if (current_original_wpt == last_kept_for_interp) {
+                continue; // Skip the first point, already handled
+            }
+
+            if (current_original_wpt->wpt_flags.marked_for_deletion) {
+                auto extra_data = static_cast<KalmanExtraData*>(current_original_wpt->extra_data);
```
tsteven4, line 365: I think using auto* here makes it more obvious to the reader that extra_data is a pointer.

kalman.cc
```diff
@@ -0,0 +1,582 @@
...
+//    global_opts.debug_level = std::max(global_opts.debug_level, debugLevelInfo); // TODO: get from options
+}
+
+void Kalman::init() {
+  // All initialization is now done per-track in process().
+}
+
+void Kalman::deinit() {
+  // Nothing to do.
+}
```
tsteven4: These can be dropped here and in kalman.h.   Then init and deinit from the base class (Filter) will be used.

tsteven4 (commented):
Do you want to add tests and documentation to this PR?  Do you want to make the previously suggested changes to Matrix (rule of zero) in this PR?

tsteven4:
Currently track segments are ignored, and possibly corrupted by deletions.  RouteList::del_marked_wpts tries to forward any deleted track segment markers when doing deletions, but that isn't used and it's not clear it would work with your deletion insertion scheme.

I have examined more real track data.  When I convert from a fit file from my edge device any pauses cause a new track segment to start.  If I convert from a garmin connect supplied gpx file of the same route all the points are in one segment regardless of pauses.  With the former this triggers a bunch of interpolation activity that generates points that are spatially close, as during the pause one is usually not moving.  Something more useful may happen when connectivity is lost in a tunnel which also will generate a new segment when the gps signal is regained.

interpolation should estimate altitude if the bounding points have altitude.

in interpolation, I think we add msecs instead of secs.  This will be relevant for sub-second sampling devices.

kalman.cc
robertlipe: Nice find. Fixed.

kalman.cc
robertlipe: It makes them less resistant to zinger detection, but you're right. I also had to raise the auto threshold. I wish I'd done that before splitting off the test suite. :-/ Fortunately, they're pretty easy to regenerate.

kalman.cc
robertlipe: I'm going to fuss about that one. We can't have it both ways. Either it's 'use auto because we don't want to say what it is' or we say what it is.  IMO, if you have to look at the <> to see what type it is, noticing that '*' over there isn't much to ask.

kalman.cc
robertlipe: Leak fixed.

Doc went up for reveiew a couple days ago in https://github.com/GPSBabel/gpsbabel/pull/1477

Testsuite is ready, but it's going to fail until this is committed. It's #1479. Having it in a different branch means I can no longer run tests here, so this needs to land soon so we can continue with it.

robertlipe:
> Currently track segments are ignored, and possibly corrupted by deletions. RouteList::del_marked_wpts tries to forward any deleted track segment markers when doing deletions, but that isn't used and it's not clear it would work with your deletion insertion scheme.

I'm just about tired of trying to make track segments work in this code. If I knew then that I know now, either a track would be an array of segments (which would have wrecked everything and I have no idea how to make it work transparently for formats without concepts of this) . So that's a fair point - if you're losing signal near the end of a segment, we're probably going to wreck thm further. 


> I have examined more real track data. When I convert from a fit file from my edge device any pauses cause a new track segment to start. If I convert from a garmin connect supplied gpx file of the same route all the points are in one segment regardless of pauses. With the former this triggers a bunch of interpolation activity that generates points that are spatially close, as during the pause one is usually not moving. Something more useful may happen when connectivity is lost in a tunnel which also will generate a new segment when the gps signal is regained.

We just currently don't have segment support in this code. It's maybe something we should do. Your GPX cases would be interesting to study. I don't think that's a blocker to landing this...it'll probably take both of us to decide what IS right and then to make it handle right. For example, do you do average speed calculation per track or track segment? If it's the latter, we have big trouble. We probably also have this trouble in the KML writer, though, don't we?

> interpolation should estimate altitude if the bounding points have altitude.

Done.

> in interpolation, I think we add msecs instead of secs. This will be relevant for sub-second sampling devices.

I already changed it. THis just changes the scalar, but OK.

kalman.cc
robertlipe: They've each come and gone a few times. The 400-line process() is not lost on me, but it's where we are right now.

tsteven4:
I offer this patch to:
1) clean up earlier computations of stats used for profile selection.  code inspection tools were tripping over some of the loose ends (e.g. previously unused median_dt_for_stats, max_speed_observed)
2) use a more efficient algorithm for computing the median.

[median.patch](https://github.com/user-attachments/files/22889569/median.patch)

PR #1479, opened 2025-10-12
kalman tests

robertlipe:
This PR contains the kalman filter tests and reference data.


tsteven4 (changes requested):
please add kalman to the TESTS variable in CMakeLists.txt (after line 414).

There is history behind why we have a list of TESTS instead of figuring it out dynamically, let's leave that issue out of any kalman work.

robertlipe:
I like the way a change from 2023 just hitched a ride. Whatever. If it's been in my work tree this long, it can't be that bad...

robertlipe:
That file has been booted off the island because the same change was committed to the trunk long ago. No idea why it surfaced now.

Too many branches in too many trees.

tsteven4:
> That file has been booted off the island because the same change was committed to the trunk long ago. No idea why it surfaced now.
> 
> Too many branches in too many trees.

what file is that?

robertlipe:
unicsv.cc was in this patch (the gbFatal/fatal at the top of wr_init) but I caught it and now it's not.

Why doesn't that show up in the log of changes in the thing that's trusted to log changes? No idea.

I'll admit, having the muscle memory of creating it, that I recognized the "nn/nn Test #11" style thingy from test runs, but I always thought that was a CI thing. Never noticed the Makefile target because I run 'make -j 8 && ./testo FFF' so habitually. TIL. YOu handled all the cmake plumbing and I never really knew much about about it. (Still don't. This doesn't make me sad, either.)

I have some other work I've been scratching on in the background once the Kalman dust settles. I started with GTest as a framework and it was just so terrible to play nice with Qt. Unsurprisingly, QTEST works pretty well.  THAT is the reason I recently cleaved option.cc from defs. More on that later.

tsteven4:
If I merge robertlipe:kalman-tests (#1479) with robertlipe:kalman-main (#1478) many kalman tests fail with small changes to latitude and/or longitude values.

PR #1480, opened 2025-10-12
add Option methods to distinguish between defaulted and supplied option values

tsteven4:
This might be a better fit for Kalman::is_using_default_value().  If the use supplies an option on the command line we should respect it even if it happens to match the default value.   This PR makes a determination based on the source of the option, not a comparison of it's value.  If it comes from argdefault or the init file it is reported as defaulted, otherwise it must be on the command line and it is reproed as not defaulted.

### DIFF
--- a/filter_vecs.cc
+++ b/filter_vecs.cc
@@ -217,9 +217,9 @@ void FilterVecs::prepare_filter(const fltinfo_t& fltdata)
         qtemp = inifile_readstr(global_opts.inifile, "Common filter settings", arg.argstring);
       }
       if (qtemp.isNull()) {
-        Vecs::assign_option(fltdata.fltname, arg, arg.defaultvalue);
+        Vecs::assign_option(fltdata.fltname, arg, arg.defaultvalue, true);
       } else {
-        Vecs::assign_option(fltdata.fltname, arg, qtemp);
+        Vecs::assign_option(fltdata.fltname, arg, qtemp, true);
       }
     }
   }
@@ -231,7 +231,7 @@ void FilterVecs::prepare_filter(const fltinfo_t& fltdata)
       for (auto& arg : *args) {
         const QString opt = Vecs::get_option(fltdata.options, arg.argstring);
         if (!opt.isNull()) {
-          Vecs::assign_option(fltdata.fltname, arg, opt);
+          Vecs::assign_option(fltdata.fltname, arg, opt, false);
         }
       }
     }
--- a/option.h
+++ b/option.h
@@ -46,6 +46,8 @@ public:
   virtual void init(const QString& id) {}
   virtual void reset() = 0;
   virtual void set(const QString& s) = 0;
+  [[nodiscard]] virtual bool getDefaulted() const = 0;
+  virtual void setDefaulted(bool defaulted) = 0;
 
   /* Data Members */
   // I.25: Prefer empty abstract classes as interfaces to class hierarchies
@@ -98,6 +100,16 @@ public:
     value_ = s;
   }
 
+  [[nodiscard]] bool getDefaulted() const override
+  {
+    return isDefaulted;
+  }
+
+  void setDefaulted(bool defaulted) override
+  {
+    isDefaulted = defaulted;
+  } 
+
 // We use overloads instead of default parameters to enable tool visibility into different usages.
   int toInt() const;
   int toInt(bool* ok) const;
@@ -109,6 +121,7 @@ public:
 private:
   QString value_;
   QString id_;
+  bool isDefaulted{true};
 };
 
 class OptionInt : public Option
@@ -146,6 +159,16 @@ public:
     return value_;
   }
 
+  [[nodiscard]] bool getDefaulted() const override
+  {
+    return isDefaulted;
+  }
+
+  void setDefaulted(bool defaulted) override
+  {
+    isDefaulted = defaulted;
+  } 
+
   void init(const QString& id) override;
   void reset() override;
   void set(const QString& s) override;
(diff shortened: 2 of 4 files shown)


tsteven4:
@robertlipe I think this works better for kalman as any value supplied on the command line can be respected instead of being potentially changed based on the profile if it happens to match the default.

robertlipe:
It sounds ridiculous, but I think I spent about three hours crumpling
papers and tossing them at the wastebasket on this issue.This was something
I'd considered, and it was actually intentional. The behavior is the same -
if the option is the same here or there, it doesn't matter where it came
from, right?

I had - and tossed - a version that was almost like this. I wasn't sure
that I liked keeping state inide the option. Back when they were const
globals, we knew that an option was never written into and that options
appearing twice in a complicated pipeline was OK because they were
stateless. Your snlen could never clobber my snlen, even if they were in
the same filter in different places in the pipeline.

The successor to this one, also tossed, passed the default value back into
the arglist (back into the option? I've forgotten the direction right now)
and then did a comparison between value_ and default_value. That kept the
options stateless.

Then I decided that in our decades of history, we had never implemented
something so insane that we had to care before, so why pollute the option
code with this at all? Just keep kalman's dirty laundry in its own little
laundry basket and if I ever lost my mind and did something as stupid as
this "'auto'  for most of five things, but you can override any of them"
flag again that I could generalize this/

I actually had version zero of that where I was was trying to determine the
IsNull vs. ~IsEmpty state of the set which _almost_ worked, but I don't
even remember why that was the first thing I threw away. I think I
concluded that I'd implemented a mirror of / a compliment of some existing
behaviour.

THIS was the reason why I made a several hour detour into
implementing QTEST.

Our option handling in modern C++ would look very different than it does in
this code. It would look more like every other key/value store like a map
of Strings. Your Option work really tried, but the way it's woven into
arglist is just layering atop a questionable C-based design.

I put non-trivial thought into it. I'm not terribly proud of what I landed
on, but it wasn't just laziness. If you have a compelling reason to go with
this one after mine lands, go for it.

RJL


➜  gpsbabel-virgin git:(kalman-main) ✗ git stash show  -p
diff --git a/option.h b/option.h
index b2f69b13..05f7f990 100644
--- a/option.h
+++ b/option.h
@@ -41,6 +41,7 @@ public:

   /* Member Functions */
   [[nodiscard]] virtual bool has_value() const = 0;
+  [[nodiscard]] virtual bool isDefault() const = 0;
   [[nodiscard]] virtual bool isEmpty() const = 0;
   [[nodiscard]] virtual const QString& get() const = 0;
   virtual void init(const QString& id) {}
@@ -73,6 +74,12 @@ public:
     return !value_.isNull();
   }

+  // Has a value been set? False if so.
+  [[nodiscard]] bool isDefault() const override
+  {
+    return value_.isNull();
+  }
+
   [[nodiscard]] bool isEmpty() const override
   {
     return value_.isEmpty();
@@ -136,6 +143,12 @@ public:
     return !value_.isNull();
   }

+  // Has a value been set? False if so.
+  [[nodiscard]] bool isDefault() const override
+  {
+    return value_.isNull();
+  }
+
   [[nodiscard]] bool isEmpty() const override
   {
     return value_.isEmpty();
@@ -186,6 +199,12 @@ public:
     return !value_.isNull();
   }

+  // Has a value been set? False if so.
+  [[nodiscard]] bool isDefault() const override
+  {
+    return value_.isNull();
+  }
+
   [[nodiscard]] bool isEmpty() const override
   {
     return value_.isEmpty();
@@ -231,6 +250,12 @@ public:
     return !value_.isNull();
   }

+  // Has a value been set? False if so.
+  [[nodiscard]] bool isDefault() const override
+  {
+    return value_.isNull();
+  }
+
   [[nodiscard]] bool isEmpty() const override
   {
     return value_.isEmpty();

robertlipe (approved):
Read the prose. If you feel strongly about it, go for it.

tsteven4:
I agree this can go after kalman merges, but this is the current behavior I am trying to avoid in bold.  Here the code was modified to print out the command line and the profile and max_speed_ used after kalman goes through profile selection and profile based initialization.

main: QList("./bld/gpsbabel", "-i", "gpx", "-f", "htcut.gpx", "-x", "kalman,profile=cycling")
kalman: "cycling" max_speed 30
main: QList("./bld/gpsbabel", "-i", "gpx", "-f", "htcut.gpx", "-x", "kalman,profile=cycling,max_speed=14.9")
kalman: "cycling" max_speed 14.9
main: QList("./bld/gpsbabel", "-i", "gpx", "-f", "htcut.gpx", "-x", "kalman,profile=cycling,max_speed=15.0")
***kalman: "cycling" max_speed 30***
main: QList("./bld/gpsbabel", "-i", "gpx", "-f", "htcut.gpx", "-x", "kalman,profile=cycling,max_speed=15")
kalman: "cycling" max_speed 15
main: QList("./bld/gpsbabel", "-i", "gpx", "-f", "htcut.gpx", "-x", "kalman,profile=cycling,max_speed=15.1")
kalman: "cycling" max_speed 15.1

If the user specified max_speed=15.0 on the command line we should respect it.  max_speed=15 should result in the same max_speed being used as when max_speed=15.0.  incrementing max_speed values on the command line should result in incrementing values used in the filter.

The need to know if an option was passed on the command line isn't unique to kalman.  It has come up before.  I believe the lack of that ability influenced the way the options for extracting extra data from igc were defined (opt_enl, ...).  It also is used in other cases for ARGTYPE_BOOL which is the only ARGTYPE that already supports detection.

robertlipe:
Ah, yes. The current one probably does a string comparison when it should do a type specific comparison.  I suppose this can go in before or after you've finished reviewing the others.

There's a lot there, but I think we're already into nits more than design and implementation issues with the exception of track segmentation which just isn't coming in this round.

PR #1487, opened 2025-10-14
Fill in many missing options from doc. (from rjlmini)

robertlipe:


robertlipe:
This PR is on a branch that doesn't exist. Super weird. Picking up in a new one.

PR #1489, opened 2025-10-14
Try to hush clazy warnings about list detachment.

robertlipe:
I'm puzzled. This is admittedly a hit and miss at 0500 just to get the
build back green.

The loop iterators are all absolutely alreay const refs. (Because I went
through this once before...) My best guess is that it thinks that
the containers themselves (*global_track_list and *wpt_list) are not const.
Could a method be calld that would cause the whole container to detach?

I don't feel like spending an our on cppreference on this right now.
This is a shotgun attempt.

### DIFF
--- a/kalman.cc
+++ b/kalman.cc
@@ -64,7 +64,7 @@ void Kalman::process() {
     interp_max_dt_ = interp_max_dt_option_.get_result();
     interp_min_multiplier_ = interp_min_multiplier_option_.get_result();
 
-    for (const auto& route_it : *global_track_list) {
+    for (const auto& route_it : std::as_const(*global_track_list)) {
 
         // Per-track state initialization
         is_initialized_ = false;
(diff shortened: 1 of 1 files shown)


robertlipe:
Build aborted. 1488 came in just after this and is needed for success anyway, so we'll let the build farm focus on that one.

tsteven4:
> I'm puzzled. This is admittedly a hit and miss at 0500 just to get the build back green.
> 
> The loop iterators are all absolutely alreay const refs. (Because I went through this once before...) My best guess is that it thinks that the containers themselves (*global_track_list and *wpt_list) are not const. Could a method be calld that would cause the whole container to detach?
> 
> I don't feel like spending an our on cppreference on this right now. This is a shotgun attempt.

I believe the issue occurred because the list wasn't const, so non-const methods like QList::begin() and QList::end() were used implementing the for loop, which can cause detachment.

PR #1495, opened 2025-10-16
Generate data for kalman q_scale_pos documentation

### LINKED ISSUE #1494: kalman filter documentation: q_scale_pos simulation is incorrect
The data for the figure "Effect of q_scale_pos on Sharp Turns" doesn't reflect what our kalman filter actually does.  I noticed the sedan starts turning before the turn.   Instead of faking the data we could run our filter and use it's output.  The test data is attached, the track is sampled a 1Hz with a constant velocity of 10m/s.  The cycling profile is automatically selected.  We override q_scale_pos.

For example: (line_color AABBGGRR).
./bld/gpsbabel -t -i unicsv,utc -f k.csv -o kml,line_color=FF00FF00 -F kf.kml #green
./bld/gpsbabel -t -i unicsv,utc -f k.csv -x kalman,q_scale_pos=1 -o kml,line_color=FF0000FF -F kf1e+0.kml # red, sedan
./bld/gpsbabel -t -i unicsv,utc -f k.csv -x kalman,q_scale_pos=100 -o kml,line_color=FFFF0000 -F kf1e+2.kml # blue, sports car

Image

[k.csv](https://github.com/user-attachments/files/22955834/k.csv)

tsteven4:
This resolves #1494 

The other graphs changed as well even though there wasn't an intended change for them.  The change in them is not obvious, they appear the same to the eye (see the github diff display).


tsteven4:
You can see a difference in the r scale plot with swipe.  Perhaps that r_scale init bug was fixed after the plot was generated?

tsteven4:
> Perhaps that r_scale init bug was fixed after the plot was generated?

No, it is using fake data also.  But without a seed np.random is generating different noise.

robertlipe:
>  graphs changed as well even though there wasn't an intended change for them

It's likely that something like PNG compression depends upon some plugin based on the version of Python installed on the host that generates these as well. I'm not going to do perceptual hashing or any such thing on these. If they look the same, they're "the same enough."

> Perhaps that r_scale init bug was fixed after the plot was generated

Just remembering the timeline on this, even if that bug wasn't at issue, that was probably true. I had the "can we explain these via graphs" near the end of a day of describing points by the timestamp of the point and "can graphs be expressed sensibly in both postscript AND in an animated way via docbook, of all things" explorations somewhat early in the process that brought us this.

I like the basic idea so much I might go back and try it with some of our other filters.

I like this and will stamp it OK, but deeply question whether image source data belongs in tests. That seems weird. Perhaps it belongs inside tools/data or something? There can be some discussion on whether the images should be checked in or not. It feels OK to me, but it obviously blows up our repo somewhat.

robertlipe (approved):
Dear GitHub, why isn't there an approve button for this same form at the BOTTOM of the page?  And why does this take two minutes to load on a 2gbps downlink? Ugh.

Dear tsteven4 - nice. Thank you. But let's collecgtively think more about the location of those data files. 

Oversize Rubber Stamp: [ Approved ]

PR #1496, opened 2025-10-17
Add Garmin icons available in BaseCamp 4.8.13.

sebastic:
While mapping the QMapShack waypoint icons to Garmin ones I noticed that BaseCamp had a few more icons that weren't in the `garmin_icon_table` yet.

Not sure when those were introduced, but BaseCamp 4.8.13 has them now.

### DIFF
--- a/garmin_icon_tables.h
+++ b/garmin_icon_tables.h
@@ -342,6 +342,25 @@ static const QVector<icon_mapping_t> garmin_icon_table = {
   { 246, -1, "Upland Game" },
   { 247, -1, "Waterfowl" },
   { 248, -1, "Water Source" },
+
+  /* BaseCamp 4.8.13 */
+  { 249, -1, "Dog, running" },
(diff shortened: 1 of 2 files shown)


robertlipe (commented):
Man, I don't know what Garmin is thinking on some of these, but they didn't ask me.

Thank you for the submission. I'll approve this but will leave the decision to @tsteven4 whether we merge this before or after the upcoming 1.11 release. He has the power to merge either way, but I'm happy to do so when it doesn't cause undue grief. Since I think I've seen release candidates roll by (an active task for me this week), there's a good case these missed the train, but it's clearly low risk if the train has to back up anyway. It's not worth recutting candidates for this, though.

I think I just argued about five different angles here. I love being middle management!

tsteven4 (approved):
thanks.  I executed our process to update the doc with the new icons also.

PR #1497, opened 2025-10-21
kalman fixes and enhancements

tsteven4:
The kalman predict/update functionality seems to work well :smile:

- use a table to contain profile based parameters
- fatal on unrecognized profiles
- delete gap_factor fallback when option doesn't have a value.  since a default is provided a value will always exist.
- condition debug statements on global debug level
- add debug statements for profile selection and parameters
- question: enforce delta time > 0?
- when computing speed correct to allow for sub-second sampling.
- add debug for point deletions
- question: should a single isolated zinger cause 3 points to be deleted?
- question: because marked_for_deltion <==> zinger deletion why do we need extra data?
- enhance debug for interpolation
- simplify entry to interpolation loop.  I assert it isn't necessary to find the first kept point ahead of time.  Note loop can already recover after a gap that we don't interpolate across.
- in interpolation loop the gap is always >= 0 so we don't need abs. (assuming time moves forward)
- question: can we use track_del_marked_wpts so track segment markers are forwarded?
- question: is there a corresponding reference for the kalman implementation to cite?
- question: can we quit adding quantization noise?
- delete test for change before assinging process output.  was this related to the need for COORDINATE_PRECISION_FACTOR?

- update dt zero test output.  Because we are now calculating speed correctly for subsecond dt all the points are thrown out.
- update interpolation tests inputs.  add another sample so median dt reflects intended sample period.  don't use unix epoch as a time, we consider it invalid.
- update interpolation test.  The tests now generate interpolated and non-interpolated results exercising both related options.
- in prefilter_edge_in test input delete erroneous comment.

The questions have FIXME comments in the code that need to be deleted after resolution.  Depending on the resolution further changes may be needed, e.g. to use track_del_marked_wpts.


tsteven4:
https://github.com/GPSBabel/gpsbabel/pull/1497/commits/03e50545f6b19b8128ad8ac13a8300696de7f60c resolves #1501 and preserves track segments in the face of deletions.

tsteven4:
@robertlipe 
Regarding
// FIXME: marked_for_deletion <==> is_zinger deletion.

It seems to me we always set marked_for_deletion and is_zinger_deletion simultaneously.  So why do we need both?  Can we get rid of extra_data and just use marked_for_deletion?  

There are two ways we handle gaps and both seem to work although one may appear to do better than the other depending on the circumstances.

1) interpolate to fill in missing points, and run the filter as normal

2) let the kf predict across the gap.

Something like both these methods are in the literature.

Is there value in deleting a point but not calling it a zinger so we can interpolate across the resulting gap (even though I don't think we ever do this)?

tsteven4:
I just added made a change to estimate the speed from the kalman filter state vector.  When running this with real cycling data I notice a few odd things.  The garmin edge 840 reports "enhanced speed" for every data point.  This speed seems to be on average 6-7% low compared to either the speed computed by the kalman filter or the track filter.  From the edge location data the speed computed with the track filter shows areas where the speed is unrealistically noisy.  If we look at the track in these areas we see that despite lots of data points they are literally off in the forest!  There are ~62 points in the image the portion of the track shown.  The tracks are the data from the edge and that data filtered by our kalman filter.  While the edge location data and speed data are smooth, the spacing between the points seems to be irregular.  Perhaps the sampling instances were varying from the 1Hz ideal.  The fit file doesn't seem to have subsecond timestamps (record id 20, no gps_metata id 160).  Our kalman filter doesn't do much to change the track shape but it does come up with more consistent speeds, especially compared to those the track filter computes.  It also seems interesting that noisy speeds computed by the track filter may indicate areas where the track location is less trustworthy.

kawaba_glitch
kawaba_speed_glitch

PR #1498, opened 2025-10-22
Leaflet Map Integration and GUI Improvements

robertlipe:
introduces Leaflet maps into the GPSBabel GUI, laying the groundwork to replace the previous Google Maps implementation for map previews.

Changes and improvements include:
(A ton of stuff that I'll pull into this desription later) 
 - Refactored map preview settings to use a a `BabelData::MapPreviewType` enum
      for consistency and persistence, with migration logic for old settings.
 - Implemented clickable info balloons for waypoint and route point markers on
      the Leaflet map, displaying `name`, `description`, and `comment`.
 - Corrected the `GpxRoutePoint` class to include `description` and `comment`
      members, aligning with `GpxWaypoint` for richer data.
 - Updated browse buttons to use platform-native icons
      (`QStyle::StandardPixmap`).
 - Enhanced `leafletbase.html`, `markerclicker.h`, `leafletmap.h`, and
      `leafletmap.cc` to standardize map marker interactions and popup displays.
 - Overhauled `babeldata.h`, `advdlg.h`, `advdlg.cc`, `preferences.cc`, and
      `mainwindow.cc` to support the new map preview enum.


tsteven4:
Replacing google maps with leaflet sounds great.  Do you have a tile source in mind?

gui/advdlg.cc
```diff
@@ -28,21 +28,30 @@
...
-  previewGmap_(previewGmap),
+  mapPreviewSelection_(mapPreviewSelection),
   debugLevel_(debugLevel)
 {
   ui_.setupUi(this);
   ui_.synthShortNames->setChecked(synthShortNames);
-  ui_.previewGmap->setEnabled(mapPreviewEnabled);
-  if (!mapPreviewEnabled) {
-    previewGmap = false;
+
```
robertlipe, line 40: If we really care about DISABLE_LEAFLETMAPPREVIEW (I started with it but quickly didn't) maybe this is where we could just disable it to make it invisible to select.

If we care about either of these flags, we have to care about what we do in the Settings object. I learned the hard way that if your Settings gets wedged in a way that UI can't unwedge, it's a bad afternoon.

gui/gpx.h
robertlipe: Rainy day project: make Gpx{wpt,trk,rte}Point into a common base class that these three derive from to avoid all this busy-work repetition.

Good job for a robot minion helper...

gui/leafletdlg.cc
```diff
@@ -0,0 +1,559 @@
...
+  if (!wpt.getDescription().isEmpty()) {
+    it->appendRow(new StandardItem(tr("Desc: %1").arg(wpt.getDescription())));
+  }
+  if (!wpt.getComment().isEmpty() && wpt.getComment() != wpt.getDescription()) {
+    it->appendRow(new StandardItem(tr("Cmt: %1").arg(wpt.getComment())));
+  }
+  if (wpt.getElevation() > -50000) {
+    it->appendRow(new StandardItem(tr("Ele: %1").arg(wpt.getElevation())));
+  }
+
```
robertlipe, line 94: Stray blank

gui/leafletdlg.cc
```diff
@@ -0,0 +1,559 @@
...
+  gpx_(mapData)
+{
+  ui_.setupUi(this);
+  this->setWindowTitle(QString(appName) + " " + QString("Leaflet Maps"));
+
+  mapWidget_ = new LeafletMap(this, gpx_, geojsonData, te);
+  auto* lay = new QHBoxLayout(ui_.frame);
+  lay->setContentsMargins(0, 0, 0, 0);
+  lay->addWidget(mapWidget_);
+
```
robertlipe, line 159: Stray blanks throughout this method

gui/leafletdlg.cc
```diff
@@ -0,0 +1,559 @@
...
+}
+
+//------------------------------------------------------------------------
+void
+LeafletMapDialog::showOnlyThis(QStandardItem* top, int menuRow)
+{
+  trace("showOnlyThis", top->child(menuRow));
+
+  // Disconnect the signal to prevent re-entrancy during programmatic changes
+  disconnect(model_, &QStandardItemModel::itemChanged, this, &LeafletMapDialog::itemChangedX);
```
robertlipe: Fix indention, blank lines

gui/leafletmap.cc
```diff
@@ -0,0 +1,339 @@
...
+  }
+  return path;
+}
+
+//------------------------------------------------------------------------
+void
+LeafletMap::showGpxData()
+{
+  this->logTime("Start defining JS string");
+  QStringList scriptStr;
```
robertlipe: Reduce the number of individual writes into scritpStr

gui/leafletmap.cc
```diff
@@ -0,0 +1,339 @@
...
+      ;
+
+  this->logTime("Done defining JS string");
+  evaluateJS(scriptStr);
+  this->logTime("Done JS evaluation");
+  emit mapRendered();
+}
+//------------------------------------------------------------------------
+void
+LeafletMap::markerClicked(int t, int i)
```
robertlipe: pick better variable names

gui/leafletmap.cc
```diff
@@ -0,0 +1,339 @@
...
+{
+  qDebug() << "LeafletMap::setAllRoutesVisibility called with:" << show;
+  evaluateJS(QString("setAllRoutesVisibility(%1);").arg(show ? "true" : "false"));
+}
+
+//------------------------------------------------------------------------
+void
+LeafletMap::resetBounds()
+{
+  evaluateJS(QStringList{
```
robertlipe: Why a QStringList here?

gui/mainwindow.cc
```diff
@@ -229,6 +248,10 @@ MainWindow::MainWindow(QWidget* parent): QMainWindow(parent)
 
   //--- Restore from registry
   restoreSettings();
+// RJL
```
robertlipe: Remove these

gui/mainwindow.cc
```diff
@@ -1176,7 +1228,7 @@ void MainWindow::preferencesActionX()
   Preferences preferences(nullptr, formatList_, babelData_);
   preferences.exec();
 
-  // The user may have changed the list of displayed formats, and/or
+  // The user may have changed the list of of displayed formats, and/or
```
robertlipe: of of

gui/mainwindow.cc
```diff
@@ -1312,3 +1371,108 @@ QString MainWindow::getFormatNameForExtension(const QString& ext)
...
+    feature["properties"] = properties;
+    features.append(feature);
+  }
+
+  // Add Routes
+  for (int i = 0; i < gpxData.getRoutes().size(); ++i) {
+    const GpxRoute& rte = gpxData.getRoutes().at(i);
+    QJsonObject geometry;
+    geometry["type"] = "LineString"; // Assuming routes are LineString for simplicity, adjust if MultiLineString is needed
+        QJsonArray coordinatesArray;
```
robertlipe: fix indent

gui/markerclicker.h
```diff
@@ -0,0 +1,40 @@
+// -*- C++ -*-
```
robertlipe, line 16: ad copyright

gui/webenginepage.cc
```diff
@@ -0,0 +1,44 @@
+// -*- C++ -*-
```
robertlipe, line 16: copyright

gui/webenginepage.h
```diff
@@ -0,0 +1,22 @@
+// -*- C++ -*-
```
robertlipe, line 16: copyright

reference/track/segmented_tracks~geojson.json
```diff
@@ -99,6 +99,7 @@
                 "type": "LineString"
             },
             "properties": {
+		"gpsbabel_feature": "track",
```
robertlipe, line 102: whitespace

gui/leafletdlg.cc
```diff
@@ -0,0 +1,559 @@
+// -*- C++ -*
```
robertlipe, line 16: copyright

gui/leafletmap.cc
```diff
@@ -0,0 +1,339 @@
+// -*- C++ -*
```
robertlipe, line 15: copyright

gui/leafletdlg.h
```diff
@@ -0,0 +1,67 @@
+// -*- C++ -*-
```
robertlipe, line 16: copyright

gui/leafletmap.h
```diff
@@ -0,0 +1,76 @@
+// -*- C++ -*-
```
robertlipe, line 16: copyright

GPSBabelDeveloper:
Between Google constantly fussing about our API key and deprecating the
Marker API (seriously
<https://developers.google.com/maps/deprecations#googlemapsmarker_in_the_deprecated_as_of_february_2023>?)
I'd kind of fallen out of love with that relationship. Almost everything
with that API that I've ever created (outside of GPSBabel's map) is on that
deprecated list in one place or another. The cuddly, small
developer-friendly days are over. (Trivia: when I joined Google, I shared
an office with Paul Rademacher
<http://paulrademacher.com/#:~:text=HousingMaps%2C%20the%20first%2Dever%20map%20mashup>
when I was there...)

This development is something I've kicked around for months and only
recently picked it back up in earnest. I remember researching tile sources
in the beginning, thinking that I absolutely did not want to run one, and
being happy with the default OSM licensing for our case. Or maybe I decided
to think more about it later and this is later. :-) Is there some reason to
NOT just use the default? Sure, there's all kind of specialized stuff out
there, but I consciously reeled in many cases where I'd start to get
carried away with complexity and threw it away after reminding myself that
we're not a mapping app. The whole point of this feature, IMO, is to
visualize that your GPS adventure loaded, that the filters did something
reasonable, and give you some confidence that the data we're handing you as
"output" is actually what's promised, so if you load it into your thingy,
there's a problem with what we've written or in the of that thingy and not
in the actual content of what you're sending.

My read of the OSM license <https://www.openstreetmap.org/copyright> is
that our use is OK. Anyone looking to scrape or abuse the tile servers can
certainly find easier ways of doing it than hacking on Qt's JS/C++ bridge.
We're "using" the OSM data and not transforming it. obscuring the source,
or redistributing the data in any way. I *think* we're OK. Leaflet is
surely a well-known and presumably approved partner to them. If you know of
history here that I should be aware of or think there's a better default,
this would be a great time to have that discussion. It's *probably* as easy
as adding a 'url:' parameter or an extra param
<https://leafletjs.com/examples/layers-control/> in the constructor, IIRC.

At one point, I tried to get elevation profile plugins, but A) they seemed
to all be out of date and incompatible with current (as of the time)
Leaflet  and B) licensing was a deal-breaker.  That fell into my "remember
this is not a mapping app" bin.  Also in that category of "developed and
discarded", I once tried to control wpt/trk/rte in the layer control (the
thingy for toggling between vector layers) and to have the markers more
interactive with the C++ panel on the left but having multiple states to
synchronize in multiple languages was a nightmare. The other thing that was
a bit funny was the love/hate relationship of Leaflet with our JSON writer.
We currently end up sending GeoJSON for LineStrings because we otherwise
melted it (a common problem for maps in general) and then wound up plucking
apart the GPX also and then feeding it hybrid GeoJSON for markers as JSON
lacks the structure (it's like the bad old days of everything being a <div
type="footnote"> than <footnote>) to attach names and such so we'd have
wound up creating a custom GeoJSON writer for it anyway. It's like
shapefiles where there's a description of a shipping container, but not
what's inside that container; basically readers and writers have to almost
customize for each other and externally agree upon conventions.

I don't recall the usage numbers, but given the obscure location of the
button to even notice this feature, it's not like we're exactly staging a
DOS attack on them.


I know the timing of this is pretty weird. I'd forgotten how close it was
when I gave up on it months ago. Given how critical this feature isn't, I'm
wondering if we should add this as a 'default on' in 1.11, leaving Google
Maps as a contingency plan, and then just drop the "gMap" code when we're
far enough down the road that it's clear that we'd make a 1.12 without it
instead of needing a 1.11.1 for some reason.  It's also really low
priority, so maybe it waits and the next one just goes straight to google
maps.

tsteven4:
1. Is HTTP user agent sent as required by https://operations.osmfoundation.org/policies/tiles/? Is HTTP Referer sent?
2. gmaps is broken,  e.g. with bounds_test.gpx:
js: InvalidValueError: setPosition: not a LatLng or LatLngLiteral: in property lat: not a number
...
js: InvalidValueError: setPosition: not a LatLng or LatLngLiteral: in property lat: not a number
js: Uncaught InvalidValueError: not a LatLng or LatLngLiteral: not an Object

tsteven4:
in leafletmap.cc I think you can set the user agent with something like this->page()->profile()->setHttpUserAgent("my user agent id")

tsteven4:
I think this will get the user agent set.  However, I haven't captured packets to prove it.
[useragent.patch](https://github.com/user-attachments/files/23103384/useragent.patch)

robertlipe:
Probably.  I'm not opposed to it and will probably fold this in.  My read
was that Leaflet, not us, is the "customer" here. They control the caching,
fetching, etc.  We're no more responsible than, say, Chrome is for
loading leaflet. Is the correct User-agent Qt or Blink (open source Chrome
that's used in Qt) or Leaflet or us?

There's some ambiguity <https://github.com/Leaflet/Leaflet/pull/9883> here,
but it seems they hard block anything that doesn't meet their
requirements.
https://community.openstreetmap.org/t/access-blocked-leaflet-tiles-just-me/135657
(Hey, I know one of those posters... He's a keeper of GSAK, which uses
GPSBabel after the creator suffered a stroke several years ago.) Leaflet
doesn't mention it in their QSG
<https://leafletjs.com/examples/quick-start/>.

Re: choice of maps - There's an embarrassment of riches (Mapbox, Stadia,
Thunderforest, Lima, Bing, Fleaflet
<https://docs.fleaflet.dev/tile-servers/using-openstreetmap-direct>, OSM
and more OSM) of tile servers that we probably COULD use and I think we'd
be under the free limits of all of them. (I'm not saying that all of these
ARE compatible; it's just q quick eample list).

One of my earliest explorations into Digital carto was requesting the CD
set of the US Tiger network and experimenting with what would now be a tile
service. In 2000, the tools to do this were totally terrible, but that was
also the days when ordering a half dozen CDs would deliver the bits faster
than my ISDN connection would have. Slinging around a couple hundred GB is
a lot easier these days. I started to go down a rabbit hole of enabling a
bunch and then went back to the tagline that I had to keep using on myself:
"We're not a mapping app."

I don't plan to engage in behaviour I wouldn't sign my name to, but I'm
genuinely unsure if it's ours to sign. Anyone engaging in something sketchy
can find WAY easier ways of accomplishing it than driving the maps through
GPSBabel or commenting out that line above.  I'd rather have the discussion
of "it wasn't us" than even the appearance that we're trying to dodge
anything. We're all just trying to provide things for others for free and
NOT get ripped off in the process.

Sidebar: I tried the agentic UI in Copilot on GitHub for the first time
with a glorified version of "go do the slugwork of addressign the reviewer
comments in this PR".  It had high-impact things like replacing "to to" and
bogus indention, though I DID coerce it to refactor gui/gpx.h to derive
from a base class. Two hours ago it told me it was working on it.  I'm not
sure if that's any better than telling Cursor, VS Code, Windsurf, CLion, or
any of the other editors to deal with it. Like half the programming world,
I'm still trying to figure out when AI coding helpers are actually helpful.
Could I have changed "to to" faster than this? Certainly. But outsourcing
slugwork is certainly appealing at some level. I don't have interns any
more. :-)

Still, I'll roll something like this in. Thanx.

tsteven4:
> 2\. gmaps is broken,  e.g. with bounds_test.gpx:
>     js: InvalidValueError: setPosition: not a LatLng or LatLngLiteral: in property lat: not a number
>     ...
>     js: InvalidValueError: setPosition: not a LatLng or LatLngLiteral: in property lat: not a number
>     js: Uncaught InvalidValueError: not a LatLng or LatLngLiteral: not an Object

This, and a crash with google maps, is fixed with the following patch.  This works for both google maps and leaflet.
[gmap.patch](https://github.com/user-attachments/files/23109792/gmap.patch)

gui/advdlg.cc
robertlipe: Fixed in 10/25 push.

tsteven4:
I rediscovered some tracing I put into mapdlg.  By setting debug_ to true in mapdlg.h and leafletdlg.h you can trace the slots.  Some of the trace commands were removed from leafletdlg.cc, e.g. in itemChangedX but there are others.  These messages can be very helpful.  After the gmap.patch above I can flip between the two maps.  I noticed how clicking an item like "High Road" in Routes with bounds-test.gpx as input behaves differently between the two, the parent item becomes partially checked with leaflet but not gmap.  I think this led to the "Prevent re-entrancy" stuff in leafletdlg.cc.  This seems to be needed,  but I think it leads to ignored signals.  With some new trace messages in itemChangedX I see
itemChangedX: parent: "Routes" item: "High Road" (row: 1)
itemChangedX: parent: none item: "Routes" (row: 2)
redundant: parent: none item: "Routes" (row: 2)

Regarding  "Disconnect the signal to prevent re-entrancy during programmatic changes" stuff.  It seems these will result in lost signals.  I haven't found a case were these are needed.

robertlipe:
> This, and a crash with google maps, is fixed with the following patch. This works for both google maps and leaflet. [gmap.patch](https://github.com/user-attachments/files/23109792/gmap.patch)

Thank you. Committed in [leaflet ededce91]

robertlipe:
> I noticed how clicking an item like "High Road" in Routes with bounds-test.gpx as input behaves differently between the two, the parent item becomes partially checked with leaflet but not gmap. I think this led to 

This is intentional, though it is indeed different. IMO, this more correctly implements the treeview tristate "all checked, all unchecked, partially checked" model. When the total number of children clicked isn't the total number of children, we set Qt::PartiallyChecked.

This did indeed lead to a (nasty to debug) feedback loop where changing the parent would emit an itemChanged while we were changing things. Unchecking a single child would immediately force that child (and all other children) to a new state based on whatever the parent decided to do. This is why we have the kind of odd locking. (I'm about 90% sure this is, at least at a block diagram level, how we implemented the QT TreeView handling in Earth's left-hand panel which was WAY more complicated than we'd typically see here.)

You were right, though, that showHideAll and OnlyThis were totally blocking the map updates because they were lost. I think I've just fixed that.

tsteven4:
> > This, and a crash with google maps, is fixed with the following patch. This works for both google maps and leaflet. [gmap.patch](https://github.com/user-attachments/files/23109792/gmap.patch)
> 
> Thank you. Committed in [leaflet ededce91]

This patch hasn't been applied or was applied and not pushed.

robertlipe:
Go home, Codacy. You're drunk.

robertlipe:
I only realized this wasn't merged when I was spot-checking binaries. I just plain lost track of it.


I'm reluctant to kick the signing can down the road again, but should we just let 1.11 sail?

PR #1499, opened 2025-10-22
Fix symbol names for Dog icons.

sebastic:
The new icons in the hunting category for dogs had the wrong description, in #1496 I had used the ones from the tooltip instead from the <sym> element in GPX file.


tsteven4 (approved):
thanks

PR #1500, opened 2025-10-23
remove unused QNetworkAccessManger from map.

tsteven4:
Alternatively this could be done only in #1498, but it is unclear at this moment if that will be in the next release.  In any event it should be merged into #1498, and the same changes applied to leafletmap.cc and .h.


robertlipe:
Done in dd79fe7b.

As to whether it's included in the next release, that's up to us. I'm leaning somewhat toward "yes," but think it deserves some airtime, including me getting back to finish it. 

That'll either be in the next couple of days, or it may be far out; my availability is about to get really weird.

robertlipe:
LGTM.

(And I'm about tired of Codacy. It's been a long time since it's told us anything useful.)

PR #1509, opened 2025-11-07
Fix mac codesiging

tsteven4:
While #1507 allowed the bundled app to run on macOS,  errors were generated during codesigning and codesigning would fail verification.  These were due to two problems:
1) we traditionally have packaged non-executable code in Contents/MacOS (COPYING.txt, gmapbase.html and translations.)  The MacOS directory is intended for only executable code according to the Apple documentation.  Having non-executable code present caused macdeployqt to generate errors when codesigning.
2) An apparently unreported bug in macdeployqt, including release 6.10.0, exists when the -executable option is used.  The CFBundleExecutable, GPSBabelFE, needs to be signed last.  But it is signed before executables passed with the -executable option.  This results in gpsbabel and it's dependency QtCore5Compat being signed after GPSBabelFE, which results in a codesign verification error.  The actual error message is different on github and macstadium, but the underlying problem seems identical and this PR fixes the issue on both.

Issue 1) is fixed by moving the non-executable resources to the Resources directory in the bundle.  Issue 2) is worked around by resigning GPSBabelFE after macdeployqt finishes.  However, this means we can't use macdeployqt to create the disk image, we have to do that manual after the re-signing.

### DIFF
--- a/gui/mainwindow.cc
+++ b/gui/mainwindow.cc
@@ -253,13 +253,18 @@ void MainWindow::switchTranslator(QTranslator& translator, const QString& filena
 
   // Set a list of locations to search for the translation file.
   // 1. In the file system in the translations directory relative to the
-  //    location of the executable.
+  //    location of the executable, or on macOS in the Resources
+  //    directory relative to the location of the executable.
   // 2. In the Qt resource system under the translations path.  This is useful
   //    if the resource was compiled into the executable.
   // 3. In the translations path for Qt.  This is useful to find translations
   //    included with Qt.
   const QStringList directories = {
+#ifdef Q_OS_MACOS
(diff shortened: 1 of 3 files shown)


tsteven4:
I have tested the DMG produced by this PR on macstadium.  The app is installable with the usual System Settings/Security & Privacy hoop if the DMG is downloaded from the internet.  The installed app can be verified with codesign.  The map preview feature works, i.e. gmapbase.html is found.  Translations also work proving they are found.  It is convenient for testing that Systems Settings/General/Language & Region allows the language of a single app to be set.

tsteven4:
A better fix is to fix macdeplyqt: https://github.com/qt/qtbase/pull/125

PR #1510, opened 2025-11-13
support macos codesign with codesigning certificate

tsteven4:
This enhances our codesigning on macos to allow for using a codesigning certificate instead of ad hoc signing.
If the secretes BUILD_CERTIFICATE_BASE64, P12_PASSWORD, ~~KEYCHAIN_PASSWORD~~ exist then the certificate will be loaded into the keychain and used for signing.  If not then ad hoc signing will be used.  See https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/deploy/deploy-to-third-party-platforms/sign-xcode-applications


tsteven4:
> This enhances our codesigning on macos to allow for using a codesigning certificate instead of ad hoc signing. If the secretes BUILD_CERTIFICATE_BASE64, P12_PASSWORD, ~KEYCHAIN_PASSWORD~ exist then the certificate will be loaded into the keychain and used for signing. If not then ad hoc signing will be used. See https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/deploy/deploy-to-third-party-platforms/sign-xcode-applications

@robertlipe please see the first two bullets in the above reference.  If you have a developer account I would like to try this for the release.  Regarding the last two bullets: We don't use a provisioning profile as we don't release with xcode.  I make up a key chain password on the VM.

PR #1512, opened 2025-11-19
Lowrance usr4 utf 16 bugfix

kkoskelin:
# Fix UTF-16 character corruption in Lowrance USR v4+ formats

## Problem
Waypoint names containing certain Unicode characters were getting corrupted when reading Lowrance USR version 4+ format files. Specifically, characters with UTF-16LE encodings containing the byte `0x01` (such as "Ā" = U+0100 = `0x00 0x01`) were being corrupted.

### Example of corruption:
- **Input**: `Test_Ā_Point` (with "Ā" = U+0100)  
- **Output**: `Test_⨀_Point` (corrupted to "⨀" = U+2A00)

This could cause crashes when processing files with international characters or accented text in waypoint names.

## Root Cause
The `lowranceusr4_readstr()` function was applying a character replacement workaround intended for single-byte text to UTF-16 data:

```cpp
buf.replace((char)0x01, '*');  // IWay 350C workaround
```

## Solution
Made the character replacement conditional so it only applies to single-byte character data (`bytes_per_char == 1`), preserving UTF-16 data integrity while maintaining backward compatibility with the IWay 350C device workaround.

### Code changes:
- Moved `buf.replace((char)0x01, '*')` inside the `if (bytes_per_char == 1)` block
- UTF-16 data (USR v4+) now bypasses the character replacement
- Single-byte data (USR v2/v3) still receives the IWay 350C workaroundstr()` function was applying a character replacement workaround intended for single-byte text to UTF-16 data.

This replacement was corrupting valid UTF-16 sequences where 0x01 appears as part of multi-byte character encodings.

### DIFF
--- a/lowranceusr.cc
+++ b/lowranceusr.cc
@@ -210,8 +210,8 @@ LowranceusrFormat::lowranceusr4_readstr(gbfile* file, int bytes_per_char) const
     buf.resize(len);
     int bytesread = gbfread(buf.data(), 1, len, file);
     buf.truncate(bytesread);
-    buf.replace((char)0x01, '*');  // IWay 350C puts 0x01 for the accented o in the street name of the Montreal Holiday Inn.
     if (bytes_per_char == 1) {
+      buf.replace((char)0x01, '*');  // IWay 350C puts 0x01 for the accented o in the street name of the Montreal Holiday Inn.
       retval = QString(buf);
     } else {
       retval = utf16le_codec->toUnicode(buf);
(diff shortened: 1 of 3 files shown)


tsteven4:
Thanks.  That kludge went in here: https://github.com/GPSBabel/gpsbabel/commit/6aa3a36d4.  It doesn't appear any test data was captured.  I will investigate the history a bit more.

tsteven4 (approved):
Thanks for the nice explanation, fix, and test.

PR #1537, opened 2026-02-23
bug fix and modernization of gtrnctr format.

tsteven4:
- use XmlStreamWriter.
- remove MultiSport from the allowed values of the sport option. Per the schema it is not an allowed Sport_t.  Modify test as the sport option is used on write, not read.
- remove writing the split attribute for the Trackpoint element.  It is not allowed by the schema.
- make sure the Watts element refers to the activity extension.


tsteven4:
There are two end cases that can result in the writer producing a file that violates the schema.  These occur when a track has at least one point without a valid time, or when a track doesn't have any points with a valid time.  While these violations aren't new we now warn when they occur.

tsteven4:
As noted above if a track has a point that is missing a time, and you use that to create a tcx file, that file will violate the schema.

If you use that track to create a course then on import Garmin Connect will work correctly without complaint.

OTHO, if you use that track to create an activity then on import Garmin Connect will say "An error occurred with your upload. Please try again."

PR #1541, opened 2026-03-01
fix new bug, vcf should use text mode for line ending translation.

tsteven4:


robertlipe:
How does that NOT cause verification errors?

GPSBabelDeveloper:
...and has anyone used VCF in the last hundred years?

Is that the one that was added for iPod or something?

tsteven4:
> How does that NOT cause verification errors?

testo diff options:
-b ignore changes in the amount of white space
-w ignore all white space

```
compare()
{
  ${DIFF} -u -b -w $* ||  {
```

tsteven4:
> ...and has anyone used VCF in the last hundred years? Is that the one that was added for iPod or something?

I though I remembered you used it in the last few years, but all I could find is a comment you made Mon, 28 Feb 2022 

> vcard Vcard Output (for iPod)
    Oddly, this is still useful. Not so much for iPod. :-)

PR #1547, opened 2026-03-02
actions/attest-build-provenance@v3-> actions/attest@v4

tsteven4:


robertlipe:
Should this be aboard this release train?

tsteven4:
I parked this due to the rename "As of version 4, actions/attest-build-provenance is simply a wrapper on top of [actions/attest](https://github.com/actions/attest)."  The bot wanted to just use attest-build-provenance@v4, and I manually produced this to eliminate what seems to be a dead repo.  But the official docs still referred to the attest-build-provenance repo.  I see they have now been updated https://docs.github.com/en/actions/how-tos/secure-your-work/use-artifact-attestations/use-artifact-attestations, so I think it's ready.

PR #1553, opened 2026-03-12
automate building of all docker archives

tsteven4:


tsteven4:
This also adds gperf back for the releases that used it.  The build may or may not use it as the relative timestamps are indeterminant.

PR #1563, opened 2026-03-29
Get conditional compilation options working with map preview

tsteven4:
1. Get all four combinations of GPSBABEL_GOOGLE_MAPPREVIEW and GPSBABEL_LEAFLET_MAPPREVIEW working.
2. add documentation of cmake related options.  rename for clarity, to me it was initially unclear if GPSBABEL_MAPPREVIEW applied to the map preview function, or just it's implementation with the google maps API.
3. revert what appeared to me to be two stray and erroneous changes with the introduction of leaflet maps, #1498.  @robertlipe please see mainwindow.cc new/old lines 449/450 and 1298/1292.
4. remove MainWindow::leafletMapPreviewActionX() and associated elements.  It didn't do anything, and the menu item enabling wasn't working correctly.

@robertlipe regarding #1498:

- I did notice that the read/blue marker coloration we do with google maps waypoints to highlight the selected waypoint isn't done with leaflet.  It seems that leaflet uses one color for waypoints, and the other for routepoints, while google doesn't have markers for routepoints.  OTHO like google, leaflet doesn't show markers for trackpoints. 
- I also noticed MarkerCluster.Default.css, MarkerCluster.css  and  leaflet.css got checked in.  I think these are fetched from the web when used.  I don't see that they are distributed with gpsbabel.  I think we can delete these.
- I also see we now derefernce InputOptionsRef and OutputOptionsRef.  I left this alone, but I don't understand why this change was made.  
- I acknowledge the change of icons on the InputFile and OutputFile browse buttons, if you  like that better I'm fine with it.


tsteven4:
@robertlipe like the css files mentioned above leaflet.js, leaflet.markercluster.js also was added but not used.

robertlipe:
Thanx for the extra eyes on this. This branch sat out of the tree for too long and I threw away a LOT of revisions along the way, so things like the .css crept in for some time when I was thinking about having offline tile servers and/or had scripted a wget/curl of the things every leaflet dev needs. So on your list...

1) Good grief that's complicated, but since you're managing the complexity, I won't veto it.  We (I) should make it a priority to simply delete the Google Maps option from the code base pretty quickly after release just so these don't take root. (I'd like to disable that key shortly after that as our key seems to be abused as much as it's used.) I think I recall that whole option was based on an era years ago when the Linux guys were so opposed to Qt tracking Google's Clank over WebKit that they were shipping either ancient Qt versions or simply disabling Javascript and/or WebView completely.  I don't consider it a particularly useful option to disable the feature, but I'm not particularly tolerant of having our code jump through hoops so one guy can build for a Nokia 770 or something. You're much kinder to that "crowd" than I am.

2) That's fallout from (1). I'd delete it before I really thought it through. :-)

3) Excellent catches. Thank you.

4) Good call. There were lots of things where I started from the GMAP implementation and, by the end, left wires next to switches or switches in the drywall that weren't connected to anything even though they once were.

*) Contrary to (4), total visual fidelity really wasn't a goal. I tried to make it useful and tried to make it look like a Leaflet user would expect it to. The usage stats on this are so low I don't expect much pushback based on inertia.
* Re: *.css, *.js - At one time, I had it running offline with those served up locally and local tile servers, but concluded that really wasn't generally a useful configuration. Then I thought it might be nice to have them local so we could pin versions, but then I discovered that they serve versioned resources on Cloudflare and changed to that. I'd preapprove removing those + tools/update_leaflet_assets.sh if you agree.
* Derefeerencing things named Ref to not return a Ref, but to return the object. I agree. That's just weird. 
* Icons: times change. Should it have been a separate PR? Probably. Clearly I'm struggling with that lately.

Actually, I've just spun up and touch-tested a PR with fixes for some of these "obvious" issues. I'll let you approve/merge if you think it's worthwhile. That's now https://github.com/GPSBabel/gpsbabel/pull/1566

Thanx for the callout.

PR #1566, opened 2026-03-30
Cleanup: Remove unused Leaflet assets and refactor options getters

robertlipe:
- Delete unused local Leaflet/MarkerCluster JS and CSS files (preferring CDN).
- Remove tools/update_leaflet_assets.sh script that only fetched things we don't use.
- Remove unused layers.png images and update gui/leaflet.qrc.
- Refactor Format class to return QList<FormatOption>& instead of pointers.
- Clean up dereference syntax in MainWindow.


tsteven4:
LGTM.  Thanks for cleaning out the red herrings.

PR #1567, opened 2026-03-31
update format skeleton.

tsteven4:


format_skeleton.cc
```diff
@@ -36,37 +48,28 @@
```
robertlipe, line 69: I know this is beyond the repair you're making here, but let's consider stretching this. (Maybe this goes into the real docbook, replacing our README.contrib which is also out of date.)

7. Make a new docbook file in xmldoc/formats/your_format_name.xml that describes the use of your new format. Follow other examples in that directory. You can test building the doc with BLABLAH and see how the doc will look in BLAHBLAH.pdf and BLAHBLAH.html

8. Describe your options in docbook fragments named xmldoc/formats/options/your_format_name-option_name.xml The example below, would get four files thusly:

xmldoc/formats/options/yfn-{icon,utc,white,lat}.xml 

If these files are misnamed, they will not appear in the doc.

9. Make a tiny shell script in testo.d/yfn.test to exercise your new format. Add 'golden' files in reference/yfn that were made by the real application. Use GPSBabel to convert them to something that's easy to compare (we use GPX or unicsv a lot) and put those in reference/yfn~gpx. In your little test script, verify that calling gpsbabel will generate a GPX that matches the resulting GPX. (Note that 'now' timestamps are suppressed or locked to Jan 1, 1970 so they're reproducible.)

10. Patches to the reference/format* files (isn't there a script to regenerate these?) so that automation can test that your format is present and correctly formatted in our various CLI help formations.

format_skeleton.h
```diff
@@ -0,0 +1,100 @@
...
+  /* Types */
+
+  /* Constants */
+
+  /* Special Member Functions */
+
+  /* Member Functions */
+
+  /* Data Members */
+
```
robertlipe, line 82: Unlike the other symbols names, which aren't negotiable,  let's note these are examples. "skeleton_icon" or something.

robertlipe (approved):
Great idea. These have been neglected for a while.

Thank you.

format_skeleton.cc
tsteven4: Check format_skeleton.cc

format_skeleton.h
tsteven4: check format_skeleton.h

robertlipe:
Lgtm

PR #1569, opened 2026-04-01
fix autouic warnings

tsteven4:
```
AutoUic: advui.ui: Warning: Creating button group `advMapPreviewButtonGroup'
AutoUic: preferences.ui: Warning: Creating button group `mapPreviewButtonGroup'
```

These were fixed by opening them in Qt 6.4.2 designer on linux, assigning the selected buttons to a new button group, and renaming the new button group to match the existing group name.  It seems like the tool used to generate the buttons is out of sync with autouic and designer.

### DIFF
--- a/gui/advui.ui
+++ b/gui/advui.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>288</width>
-    <height>196</height>
+    <height>206</height>
    </rect>
   </property>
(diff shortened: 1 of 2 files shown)


tsteven4:
It's also disappointing that dimensions sometimes get adjusted when opening  a ui on different platforms.  The dimensions need to be big enough for all platforms even if the underlying fonts are a bit different.

PR #1570, opened 2026-04-01
warning fixes

tsteven4:
jeeps/gpsapp.cc:2037:24: warning: ‘%d’ directive output may be truncated writing between 1 and 11 bytes into a region of size 8 [-Wformat-truncation=]

Note that the Garmin Device Interface Spec for A101 Waypoint Category
Transfer Protocol says packets beyond the maximum for the device should
be ignored.


robertlipe:
I don't particularly trust sentences like that in Garmin's 20+ y/o doc, but I also suspect that the intersection of devices that use categories and the devices using garmin protocol and the pool of users of such devices that care is approaching the null set.

GPSBabelDeveloper:
I don't think we can a lot of category numbers outside 0-15 (1-16), to have
enough digits to blow that up, either...but I get why GCC is whining about
it. It can't know. (Well, it COULD know...we could make it easier for the
value range propagation pass to detect such things and maybe nudge it with
c++26 contracts, but expecting it to parse printf specifiers to know how
many MIGHT be written vs. how many COULD POSSIBLY be written is still
asking a lot.)

I just hate all of that Garmin fixed-length stuff so much. It's cost us so
much grief and time over the years. But there are so many hardened battle
scars learned and enshrined in gpsapp.c that I"m slightly reluctant to just
take it behind the deprecated/ barn and shoot it.

It's been over 20 years since Garmin released any new - not derivitave,
re-engineered for cost or market viability reasons (e.g. SirfStar being
sued into oblivion, forcing re-relase with Mediatek...) and almost that
even with that exception removed. I _really_ wonder if anyone would notice
if we deprecated Garmin protocol... Garmin quit caring a LONG time ago.
We've kept that hardware viable way longer than the maker has. How long do
we have to care about GPS-II, 12, or even 2006's venerable 60/70C[S]x?

tsteven4:
Well since we only allocate 16 entries, we should make sure we don't write outside the bounds of gps_categories anyway.

robertlipe:
I agree. It's good taste. But ISTR (and it's hinted at in the 'let's cheat'
comment block above - which sounds like my writing, as I was the one that
added 276) the hardware in question only has 16 categories. Sure, it's good
to enforce it in code in case a CRC-outsmarting lightning bolt hits the
cable of a 23 y/o marine unit while transferring packets that, honestly,
have probably never transferred by anyone but me and makes n and thus i go
out of bounds. (And I'm not a boater. My 276C never left the house.)

We've probably now spent more time contemplating this thing that can never
happen than anyone has ever spent actually using it.

Today is earmarked for a GPSBabel day...

tsteven4:
RIP Rick.  Still not to zero, sorry.

jeeps/gpsapp.cc
```diff
@@ -2030,8 +2031,17 @@ void GPS_D120_Get(int cat_num, char* s)
    * so mimic the behaviour of the 276/296.
    */
 
+  // Ignore packets beyond maximum number of categories.
+  static_assert(sizeof(gps_categories)/sizeof(gps_categories[0]) == 16);
+  // g++ 13 isn't smart enough to take advantage of the static assertion
+  // when evaluting for Wformat-truncation=, so we use a matching
+  // literal in the check below.
+  if ( cat_num < 0 || cat_num >= 16) {
```
robertlipe: I hate to bikeshed this since I've already expressed my skepticism that this code ever runs, byt maybe we can reduce our occurrences of '16' instead of adding to them.

-char gps_categories[16][17];
+std::array<char[17], 16> gps_categories;

+  // Do you still need this?
+  static_assert(gps_categories.size() == 16, "gps_categories.size() != 16");

/// but this then gets an ugly cast...unless we make cat_num a size_t
+  if (cat_num < 0 || static_cast<size_t>(cat_num) >= gps_categories.size()) {
+    return;
+  }

jeeps/gps.h needs a
-extern char gps_categories[16][17];
+extern std::array<char[17], 16> gps_categories;

jeeps/gpsapp.cc
```diff
@@ -2030,8 +2031,17 @@ void GPS_D120_Get(int cat_num, char* s)
...
+    return;
+  }
+
   if (*s) {
-    strncpy(gps_categories[cat_num], s, sizeof(gps_categories[0]));
+    qstrncpy(gps_categories[cat_num], s, sizeof(gps_categories[0]));
```
robertlipe, line 2044: Or we take the above to the next level and quit treating them as C arrays of 17 bytes. Make it an array of QStrings. Then read gets super-easy...and there is no write.

robertlipe (approved):
I'm fine with this. Thanx.

If you want to keep picking at gps_categories[], there's an easy way to do that but the only user of it that I can find is commented out.

Is this whole function effectively dead?

jeeps/gpsapp.cc
tsteven4: The problem isn't the strnop-truncation warning, qstrncpy fixes that.  It's the format-truncation= warning in the other branch of the if.  The only way I have found to not get that issued with the warning enabled is to use a literal in the range check.  For example 'if ( cat_num < 0 || cat_num >= std::ssize(gps_categories)) {' doesn't work with std::array<char[17], 16> gps_categories;.  Once cheap way is to allocate 20 bytes instead of 17, that would work until an int is bigger.

robertlipe: I poked at this harder. 

We can completely REMOVE the definition and extern rdecl for
// std::array<char[17], 16> gps_categories;
and #if 0 away th entire body of GPS_D120_Get ...

and the whole thing still works. If anything were looking at gps_categories[], it would fail to link.

The only reference to it that exists is at the end of print_categories() in garmin_fs.cc is commented out. :-/

Options.
1) Do as little or as much of the above as you feel like trying to repair this table that's never used.
2) Ignore the whole shebang 
3) Start pulling at the root and delete everything in *app.c touching gps_category_type, including GPS_D120_Get
4) Complete 3 and then accept that nobody uses ini files, these receivers are all dead realize the badly named print_categories() could very easily be a const static QStringList[16] of {"Category 0", "Category 1", ..." - or a dynamacally created version of the same.
5) Complete 4, then realize that you've gutted one of the two functions in garmin_fs.cc and once you apply the same logic (we're special casing numbers in GPX files that Garmin doesn't special case...) to the _other_ function, it dosn't make any sense and should just devolve into (return argv[1];) which means that function should probably be pulled up by the root, too. which really leaves an empty garmin_fs.cc
6) Reconsider our life choices, goto 2, enjoy a chilled beverage.


Nobody has noticed that none of this has worked sensibly in years. The hardware involved is all 20+ years old and had internal batteries that weren't intended to be user replaceable. (For the suitably determined, ALL batteries are replaceable, but we know this means they're in junk drawer or have been recycled.)

If you want to commit the above and leave some combination of the rest above after 1.11, just create a ticket, assign it to me, and I'll kill the whole chain.

robertlipe:
I'd respect you no less for doing that.

They're write-only variables. NOTHING reads these, and this is the only function that writes into them. They were a static size because that's the convenient tool we had in C.

If you don't want to do the full amputation (which I totally get), apply painkillers (pad the lengths of the strings we're writing into) and move on.

PR #1571, opened 2026-04-01
Don't just deprecate autotools, remove them.

robertlipe:
Don't just deprecate autotools; remove them.


robertlipe:
Actually, I've fallen out of love with the whole deprecated/ idea. The master plan was to have a way for people to easily roll back in deprecations they cared about but A) nobody has ever done this that I can recall B) git exist.

I might possibly be talked into keeping formats there for one release, but I know that once we've moved a format in the rental garage, it'll just collect dust and not get taken to the dump instead.

Is there anything in deprecated that you'd fight to keep? Can I just nuke the whole thing and we forget about this idea?

tsteven4 (approved):
I approve this with enthusiasm.

These were deleted at one time but snuck back in.  We have had at least one user try to resurrect configure, I want to make that as difficult as possible for current code.

robertlipe:
Meh: Total duration [50m 6s]

PR #1572, opened 2026-04-01
deprecate deprecate

robertlipe:
- **Don't carry deprecated junk. We have git.**


robertlipe:
> I approve this with enthusiasm.

I read |this| to mean the extended proposal, not just the PR. Here's the application of that idea. Let's leave the idea of deprecated/ behind.

Ugh. Git acted completely differently than it did three minutes earlier. How can removals have conflicts? Why does Git randomly drag up things from five+ years ago? I'll wrestle that...

tsteven4:
This branch is [7 commits ahead of](https://github.com/GPSBabel/gpsbabel/compare/master...deprecate-deprecate) and [434 commits behind](https://github.com/GPSBabel/gpsbabel/compare/deprecate-deprecate...master) master.

When you 'git checkout-b newbranch' newbranch will be based on whatever you branch you have checked out.

tsteven4:
> I read |this| to mean the extended proposal, not just the PR. Here's the application of that idea. Let's leave the idea of deprecated/ behind.

No, I just meant I really loath configure.  But I am not opposed to 'git rm -r deprecate'.

tsteven4:
Let's commit #1571 first.   That way if #1572 ever gets reverted autotools won't come back.

robertlipe:
Sometimes I'll let it gaslight me, but this was minutes ago and it's still in my scrollback.  I'm not ON a branch of a branch, am I? I was on a branch. Then I checked out master. Then I made a new branch. Same process that I did for d-a-h minutes earlier.

➜  gpsbabel git:(deprecate-autotools-harder) ✗ l deprecated

➜  gpsbabel git:(deprecate-autotools-harder) ✗ git checkout master
Switched to branch 'master'
➜  gpsbabel git:(master) ✗ pwd
/Users/robertlipe/src/gpsbabel
➜  gpsbabel git:(master) ✗ git checkout -b deprecate-deprecate
Switched to a new branch 'deprecate-deprecate'
➜  gpsbabel git:(deprecate-deprecate) ✗ git rm -r deprecated/
rm 'deprecated/README'

Still, I went through the resolution process, scrapped the stray unicsv entry, and am re-pushing.

Agreed, though. I'll wait for the autobuild of #1571 to finish, submit that, and then update this one to reflect that, and then commit this since we agree and are both at our keyboards and in this mode.

tsteven4:
When git misbehaves for me it's because an earlier git command failed and my fingers rushed ahead without realizing it.

tsteven4:
and since we are both live can we try a mac singing cert?

robertlipe:
I think I've learned that while I spent most of this century dealing with grown-up SCM tools that can handle overlapping merges and commits, git[hub] just doesn't handle it well. I'm tired of creating merge conflicts with myself in different branches/tabs. It just works badly. 

There's some irony that the more extensive the CI queue is, the more likely one is (I am...) tempted to actually treat it like a queue and throw multiple things at it, and the more badly that is to work. Then it'll trot out PRs from five years ago, like a cat proudly plopping a dead bird on my pillow.

This is probably why we're deprecating autotools again...

Indeed, Mac signing was on my list after reviewign patch/pr queues, but I'll move it to be next. I just discovered that my Apple dev account expired while I was displaced for those months, so I'm going to have to rekindle that flame. Hope there's not a waiting period.

tsteven4:
> Indeed, Mac signing

thanks!  I can help if needed, it should be github secret entry only, no repo changes.  We should download some CI artifacts afterwards to verify.

tsteven4:
git rm -r ./src/core/deprecated
git rm -r ./testo.d/deprecated
git rm -r ./style/deprecated

tsteven4:
196 files touched! that has to be close to the record!

GPSBabelDeveloper:
It's confirmation that our instinct/experience was/is spot-on: we're paying
for a storage unit that's holding junk we don't need. Tossing 197 when a
Q&D scan says we "only" have 264 *.cc and .h in the CLI and another 54 in
the GUI (I don't want to debate whether a .ui or a CMakefile is "code")
shows we just took to the curb about half as many files as we have that are
actually live. In broad strokes: 197 out vs 321 remaining.

Hopefully this will improve the signal-to-noise ratio in tools like `grep
-r` (I prefer `ack`) or Codacy-like tools (though I think we excluded that
one) when they flag issues for code no longer in use. Maybe it'll actually
help new contributors (hey, we got one!) more to point them at _current_
examples and not get distracted at how things were in the old days.

Onward!

tsteven4:
after deleting .git, with files in the top level (gpsbabel) listed as root:

$ dirs=$(find gpsbabel -maxdepth 1 -type d) && for d in ${dirs[@]}; do echo -n "$d: " && find $d -type f  | wc -l  ; done && echo -n "root: " && find gpsbabel -maxdepth 1 -type f | wc -l
gpsbabel: 1719
gpsbabel/tools: 109
gpsbabel/src: 20
gpsbabel/win32: 3
gpsbabel/mac: 20
gpsbabel/.idea: 1
gpsbabel/msvc: 1
gpsbabel/xmldoc: 430
gpsbabel/snap: 2
gpsbabel/strptime: 9
gpsbabel/jeeps: 42
gpsbabel/style: 16
gpsbabel/reference: 567
gpsbabel/gui: 136
gpsbabel/GPSBabel.xcodeproj: 5
gpsbabel/zlib: 56
gpsbabel/.github: 12
gpsbabel/intdoc: 1
gpsbabel/shapelib: 12
gpsbabel/testo.d: 76
root: 201

Perhaps .idea and GPSBabel.xcodeproject don't belong.

tsteven4:
post-release I am thinking of a directory reorg.  It makes it a bit harder to trace history, but we have so much in the root it makes it hard to see what is there.  I've played with it and moving the source in the root to a new directory cli is pretty easy.

robertlipe:
Agreed in principle.

I've wanted to do this for years. I started years ago with the promise that
all the NEW stuff would go into src/core and that I'd git mv ${SOMETHING}
src/filters and src/formats and src/third_party and ... I'm sure your list
is similar to, and likely a superset of, mine.

I don't even think we should inherently have an include/{formats,filters}.
The ONLY thing including height.h should be height.cc so it wouldn't make
me sad at all for them to be siblings. starting "${EDITOR} height* instead
of stacking the deck so they're at the same level of directory stacking so
shell wildcard would allod ${EDITOR} */*/formats/height* just feels icky.
I'll admit I'm close to the border on this one and could be swayed, but
that's a discussion for a different day.

 I SHOULD have bitten that bullet when moving from SVN to Git, but here we
are.

PR #1578, opened 2026-04-01
Qt 6.10.3

tsteven4:
move CI 6.10.2 jobs to 6.10.3.
The WoA release will be based on 6.10.3 as 6.10 is the first WoA with QtWebEngine support.

Unfortunately https://qt-project.atlassian.net/browse/QTBUG-142131 didn't make it into 6.10.3, so we will have to work around it and wait for it to make it to the next LTS.  The workaround is in place, #1548.


tsteven4:
I want to clear cache and wait for the official announcement to make sure we don't use any pre-release snapshot bits.  The release seems to be going out to the servers now.

tsteven4:
This can't run from a foreign repo.  will resubmit from native repo.

PR #1579, opened 2026-04-01
Delete generated files for XCode, JetBrains

robertlipe:


tsteven4 (approved):
LGTM

robertlipe:
Agreed. XCode files used to be magical, but can now generated by CMake, I think.  (I don't...) These are all generated IDE files that got sucked in by mistake.

tsteven4:
we do an xcode build every time in CI, generating the project from cmake.  However, we release with a Ninja build.

PR #1585, opened 2026-04-03
release candidate 1.11.0

tsteven4:
The following release versions are built:
macos: Qt 6.8.3. requires macOS 12 or higher.
windows x86_64: Qt 6.8.3. windows 10 or higher.
windows on arm: Qt 6.10.3. includes GUI with map preview.
ubuntu: snap for amd64 and arm64. uses ubuntu distributed Qt from noble, Qt 6.4.2. includes GUI with map preview.  These snaps will have to be rebuilt after we tag the release as the version numbers in them are based on tags.


tsteven4:
This is a continuation of #1472 with the source changed from the fork tsteven4 to a repo branch so secrets can be used by github actions.  Note that the final SHA of #1472 matches the SHA of this PR when it was opened.

tsteven4:
@robertlipe are we ready to merge this and collect the release artifacts?  If possible it would be good to check an existing [artifact](https://github.com/GPSBabel/gpsbabel/actions/runs/23988793370/artifacts/6273581147) from this PR to make sure a usb device can be accessed.  I can't do this with macstadium.  I think that we don't need any entitlements for this as we aren't sandboxed.  From what I can see on macstadium the gui and mappreview seem to work fine.

tsteven4:
I may be adding things like 
image

tsteven4:
@robertlipe can we merge and release now?

PR #1602, opened 2026-05-31
garmin_gpi: add "comment" option to write the multi-line POI comment record

naturalgeek:
## Summary

Adds an opt-in boolean option **`comment`** to the `garmin_gpi` writer.

The writer currently routes a waypoint's description/notes either to the address record (`0x8000b`, via `descr`/`notes`/`position`) or to the description record (`0x0a`). It has never been able to write the **comment/note record (tag `0x0e`, flag `0x01`)** — even though the *reader* already parses it (`read_tag()`).

That `0x0e` record is the field Garmin handheld units with a single comment box (e.g. **GPSMAP 276CX**) render as the **multi-line POI comment** on the detail page. The `0x0a` description record renders there as a single, truncated line — so a gpsbabel-built `.gpi` could only ever show one line of comment text on such devices.

## Change

When `comment` is given, the description/notes string is written to the `0x0e` comment record instead of the `0x0a` description record — a single, clean record. On-disk framing matches the reader (and Garmin POI Loader):

```
[u32 tag=0x0e][u32 size=len+9][u8 flag=0x01][ write_string(str, long) ]
```

The option is **off by default**, so existing output is byte-for-byte unchanged.

## Why an option (and how this differs from POI Loader)

Garmin's POI Loader emits `0x0e` only when a waypoint has **both** `<cmt>` and `<desc>` populated: it writes `<cmt>`→`0x0a` and `<desc>`→`0x0e` as two separate records. That suits devices which display "description" and "comment" as distinct fields — but on a single-comment-box handheld the two records stack and the text can appear duplicated. This option instead lets the user emit a **single clean `0x0e`**, which is what such devices need. Making it the default would change existing output (and auto-mimicking POI Loader's two-record behavior would reproduce the duplication), so it is opt-in.

## Testing

- `testo.d/garmin_gpi.test`: added a round-trip case (`gpx → garmin_gpi,comment → gpx`) that reproduces the reference GPX exactly. Full `garmin_gpi` suite passes.
- Verified on a GPSMAP 276CX: with `comment` a multi-line POI comment renders correctly (3 lines); the default writer shows a single line. Byte-level decode confirms the option emits exactly one `0x0e` record and that default output is unchanged.

## Notes

- `xmldoc/formats/garmin_gpi.xml` documents the option.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tsteven4:
The serialization test failures require updates to some reference files because you added an option.  You can run testo with the -d option to preserve the test output, and copy the test output to the reference directory.  Check to make sure the changes are expected.

naturalgeek:
> The serialization test failures require updates to some reference files because you added an option. You can run testo with the -d option to preserve the test output, and copy the test output to the reference directory. Check to make sure the changes are expected.

@tsteven4  This should be fixed now.

PR #1615, opened 2026-08-06
Set desktop file name for the application

City-busz:
This ensures that the XDG toplevel app ID matches with the desktop file name, so Wayland compositors could match the window with the application and show the appropriate icon for them.

Without this change, the app ID is `org.gpsbabel.gpsbabelfe`, but the desktop file is called as [`gpsbabelfe`](https://github.com/GPSBabel/gpsbabel/blob/278a635eb8b3cd0625b24b8386662d622419bebe/gui/gpsbabelfe.desktop).

References:
- https://github.com/qt/qtbase/blob/bf1ae5862807ec2b9a411506d59ccd566937a4c7/src/plugins/platforms/wayland/qwaylandwindow.cpp#L152C20-L177
- https://wayland.app/protocols/xdg-shell#xdg_toplevel:request:set_app_id

### DIFF
--- a/gui/main.cc
+++ b/gui/main.cc
@@ -49,6 +49,7 @@ int main(int argc, char** argv)
   QApplication::setOrganizationName("GPSBabel");
   QApplication::setOrganizationDomain("gpsbabel.org");
   QApplication::setApplicationName("GPSBabel");
+  QApplication::setDesktopFileName("gpsbabelfe");
 
   MainWindow mainWindow(nullptr);
   mainWindow.show();
--- a/gui/main.cc
+++ b/gui/main.cc
@@ -37,6 +37,9 @@ int main(int argc, char** argv)
 #error this version of Qt is not supported.
 #endif
 
+  // Set desktop filename before creating QApplication, see QTBUG-149180
+  QApplication::setDesktopFileName("gpsbabelfe");
+
   QApplication app(argc, argv);
   // Don't override the window icon on macos.
   // Overriding defeats style preferences which otherwise will modify the appearance
@@ -49,7 +52,6 @@ int main(int argc, char** argv)
   QApplication::setOrganizationName("GPSBabel");
   QApplication::setOrganizationDomain("gpsbabel.org");
   QApplication::setApplicationName("GPSBabel");
-  QApplication::setDesktopFileName("gpsbabelfe");
 
   MainWindow mainWindow(nullptr);
   mainWindow.show();


tsteven4:
On a windows 11 pro host, running ubuntu resolute under hyper-v, this results in:
1) the correct icon being displayed in the dock when the application (gpsbabelfe) is running instead of a generic icon.
2) a message sent to the invoking terminal when the application is run from the commandline:
`qt.qpa.services: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an application ID")`

City-busz:
This setting is unrelated to XDG Portals. It's applied only for the XDG toplevel ID on Wayland.

tsteven4:
The application dock with the correct icon shown along the left edge:
Screenshot 2026-08-08 120155


With a debug statement to show the platformName:
```
tsteven4@ubuntu-resolute:~/work/gpsbabel$ gpsbabelfe
platform name is "wayland"
qt.qpa.services: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an application ID")
```

tsteven4:
This may be related to https://qt-project.atlassian.net/browse/QTBUG-145847

City-busz:
What is the output of the following command on your system?
```
WAYLAND_DEBUG=1 gpsbabelfe |& grep 'xdg_toplevel#[0-9]\+\.set_app_id'
```
For me it shows `org.gpsbabel.gpsbabelfe` without this patch, and `gpsbabelfe` with this patch.

tsteven4:
without patch (a158d0ded68c1cda2c4196a1d3ca03578336be1b)
[ 685918.287] {Default Queue}  -> xdg_toplevel#40.set_app_id("org.gpsbabel.gpsbabelfe")
with patch
[ 938139.892] {Default Queue}  -> xdg_toplevel#40.set_app_id("gpsbabelfe")

Install for both was done with CMAKE_INSTALL_PREFIX=$HOME/.local

tsteven4:
Here is the entire log with the patch:
[dut.log](https://github.com/user-attachments/files/30918064/dut.log)

City-busz:
I'm not sure why the icon is shown for you when the app ID is `org.gpsbabel.gpsbabelfe`, but I couldn't see the icon on the dock in GNOME without this patch, and it also causes missing icon in KWin and in other Wayland compositors.

tsteven4:
> I'm not sure why the icon is shown for you when the app ID is `org.gpsbabel.gpsbabelfe`, but I couldn't see the icon on the dock in GNOME without this patch, and it also causes missing icon in KWin and in other Wayland compositors.

I'm sorry, I think you misinterpreted my reported observations.  To be clear, 
On a windows 11 pro host, running ubuntu resolute under hyper-v,



  | without PR1615 | with PR1615
-- | -- | --
icon in dock (on left in previous image) | generic | gpsbabel
console message | none | qt.qpa.services: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an application ID")

tsteven4:
The console warning is coming from QDesktopUnixServices.  This is relatively new Qt code (6.10 ) to "a new API in xdg-desktop-portal meant for toolkits to call before any other call to X-D-P that allows to explicitly register the app so it's correctly identified by X-D-P for sure, rather than based on heuristics."

https://qt-project.atlassian.net/browse/QTBUG-133401

While the intent was to allow the desktop file name to be set before or after the app is created (https://github.com/qt/qtbase/blob/ece423fe797d797ff9e28aa6f89a52c1c65c282d/src/gui/platform/unix/qdesktopunixservices.cpp#L450-L451) it does not appear to work in the later case.  This results in the warning
 `qt.qpa.services: Failed to register with host portal QDBusError("org.freedesktop.portal.Error.Failed", "Could not register app ID: Connection already associated with an application ID")`.
If xdg-desktop-portal is running verbose one can see
```
XDP: Adding XdpAppInfo: XdpAppInfoHost app '' for :1.95
```

I note that in the previous case where setDesktopFileName was not called, there was an informational message `$QT_LOGGING_RULES='qt.qpa.services=true' gpsbabelfe
qt.qpa.services: QGuiApplication::desktopFileName not set. Unable to register application with portal registry`
Registration also fails because unlike qxcbwindow and qwaylandwindow there is not default value for desktopfilename in qdesktopunixservices.
Again, with xdg-desktop-portal verbose:
```
XDP: Adding XdpAppInfo: XdpAppInfoHost app '' for :1.179
```

I also note that if the call to setDesktopFileName is done before the app is constructed X-D-P registration succeeds and an informationl message is logged (not a warning message):
` QT_LOGGING_RULES='qt.qpa.services=true' gpsbabelfe
qt.qpa.services: Successfully registered with host portal as "gpsbabelfe"`
If xdg-desktop-portal is running verbose one will see
```
XDP: Adding registered host app 'gpsbabelfe'
XDP: Adding XdpAppInfo: XdpAppInfoHost app 'gpsbabelfe' for :1.150
```

Therefore, I recommend the following change:
[register.patch](https://github.com/user-attachments/files/31036426/register.patch)

tsteven4:
While we can work around the released Qt behavior we should have been able to set the desktop file name after construction of the QApplication instance.  Therefore I filed a bug with Qt: https://qt-project.atlassian.net/browse/QTBUG-149180

gui/main.cc
```diff
@@ -49,6 +49,7 @@ int main(int argc, char** argv)
   QApplication::setOrganizationName("GPSBabel");
   QApplication::setOrganizationDomain("gpsbabel.org");
   QApplication::setApplicationName("GPSBabel");
+  QApplication::setDesktopFileName("gpsbabelfe");
```
tsteven4, line 52: Please move this before the QApplication instance is created.  See https://github.com/GPSBabel/gpsbabel/pull/1615#issuecomment-5284975696
which has a patch to do just that.

tsteven4 (changes requested):

tsteven4 (approved):

tsteven4:
Thanks for submitting this.

PR #1616, opened 2026-08-06
Update launchable desktop ID in metainfo file

City-busz:
It has been renamed in a19263055bc6c8bfb035dc06edd9de64f0cb9160.


tsteven4:
good catch, thanks.

PR #1617, opened 2026-08-06
Install metainfo files

City-busz:
These are expected to be installed into `/usr/share/metainfo/%{id}.metainfo.xml`

Reference: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location


tsteven4:
Thanks

PR #1628, opened 2026-09-17
fix infinite loop converting human readable lat/lng.

### LINKED ISSUE #1627: infnite loop with malformed unicsv/xcsv input
This can occur with unicsv flt_latitude or flt_longitude fields.
This can occur with xcsv LAT_HUMAN_READABLE, LON_HUMAN_READABLE, or LATLON_HUMAN_READABLE fields.

When human_to_dec detects a '.' it passes the input to strtod.  However, if the the '.' isn't followed by a non-empty sequence of decimal digits then
the subject sequence does not have the expected form and strtod sets end to cur.  As cur doesn't advance we are stuck in an infinite loop.
For example:
```
  double lat;
  double lon;
  char test[] ="N .° 09.12′ W 085° 09.36′";
  human_to_dec(test, &lat, &lon, 0 );
```

tsteven4:
This resolves #1627


tsteven4:
human_to_dec has other unresolved problems.  If we fail to convert degrees, minutes and seconds we can end up reading the unset pointed to value and either leave the result alone or do a nmea style conversion on it leading to mysterious returned values.

